PR #12, opened 2022-03-09
added output stats

otherview:
Adds a new struct that process data from both the stats and the network and prints it.

Ideally we can create a ticket to decouple the many features of the Stats struct.
Right now the Stats collects data from the mock node, the obscuro node and the tests, while also serving as test argument and end-run-summary display.


tudor-malene (approved):
LGTM

PR #32, opened 2022-03-16
upgrading to 1.18

otherview:
As per title.

Quick Note: Might delay this merge a few days until tooling has caught up. (mainly homebrew recipes and golangci-lint)


joeldudley (approved):
LGTM

PR #44, opened 2022-03-21
Adding fake enclave rpc client

otherview:
- Added a fake rpc client for the enclave - Raised this PR mainly because it's super hard to use debuggers with RPC connections.
- Enclave now creates it's own goroutine - Instead of allowing a caller to `go Start` the enclave
- Smaller changes to the simulation start.


integration/simulation/simulation.go
```diff
@@ -53,15 +55,22 @@ func NewSimulation(nrNodes int, l1NetworkCfg *L1NetworkCfg, l2NetworkCfg *L2Netw
 
 		// create an enclave server
 		nodeID := common.BigToAddress(big.NewInt(int64(i)))
-		port := uint64(ENCLAVE_CONN_START_PORT + i)
-		server, err := rpc.StartServer(port, nodeID, stats)
-		if err != nil {
-			panic(fmt.Sprintf("failed to create enclave server: %v", err))
+		var enclaveClient enclave.Enclave
+		if virtualEnclave {
```
joeldudley, line 59: @otherview Not in love with the term `virtualEnclave`. Perhaps `localEnclave`? It's still a real enclave, just not over rpc.

integration/simulation/simulation.go
```diff
@@ -53,15 +55,22 @@ func NewSimulation(nrNodes int, l1NetworkCfg *L1NetworkCfg, l2NetworkCfg *L2Netw
...
+			enclaveClient = rpc.NewEnclaveFakeRPCClient(nodeID, stats)
```
joeldudley, line 60: Similarly, if we're keeping the fake RPC client (see above), perhaps name it `NewEnclaveLocalClient` or similar instead?

go/obscuronode/enclave/enclave.go
```diff
@@ -130,44 +130,46 @@ func (e *enclaveImpl) Start(block types.Block) {
...
-		case <-e.speculativeWorkInCh:
-			b := make([]L2Tx, 0, len(currentProcessedTxs))
-			b = append(b, currentProcessedTxs...)
-			state := copyProcessedState(currentState)
-			e.speculativeWorkOutCh <- speculativeWork{
-				r:   currentHead,
-				s:   &state,
-				txs: b,
+	// start the speculative rollup execution loop on its own go routine
+	go func() {
```
joeldudley, line 134: @otherview I'd make this a non-anonymous function because the indentation is a bit much for my liking, but I don't feel too strongly about it.

joeldudley (changes requested):
See comments.

tudor-malene (changes requested):
Thanks, Pedro, I like the idea of bypassing the RPC layer during development.

I also agree with Joel, and think that the main acceptance test should go through the rpc.

We should think about a way to achieve this (have a pleasant dev experience, and also make sure the test runs properly when it runs on CI). I'm sure there must be a canonical way to do this in go.

A few questions on the implementation:
Q1: Why can't we inject an instance of the "enclaveImpl", like we had before? Is there a reason to create a wrapper on top of it?

Q2: I also don't quite understand why the main loop has to be started as a go routine.

integration/simulation/simulation.go
otherview: Yep changing it to `EnclaveTestRPCClient` and `NewEnclaveTestRPCClient` respectively.

integration/simulation/simulation_test.go
```diff
@@ -49,6 +49,7 @@ func TestSimulation(t *testing.T) {
 		l2NetworkCfg,
 		avgBlockDurationUSecs,
 		avgGossipPeriod,
+		true,
```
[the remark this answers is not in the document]
otherview: Good catch 👍

go/obscuronode/enclave/enclave.go
otherview: I agree, changed it to :
```
func (e *enclaveImpl) Start(block types.Block) {
	// start the speculative rollup execution loop on its own go routine
	go e.start(block)
}

func (e *enclaveImpl) start(block types.Block) {
	headerHash := block.Hash()
	s, f := e.db.FetchState(headerHash)
	if !f {
		panic("state should be calculated")
	}
....
```

As a side note @tudor-malene this is a place where we might want to start passing the error around because if the DB for some reason can't be started it should fail here and propagate upwards to the client.

joeldudley (approved):
LGTM. Tudor?

tudor-malene (changes requested):
Looks good, once we get rid of the proxy fake client that doesn't really do anything.

tudor-malene (approved):
lgtm

PR #82, opened 2022-04-11
Allows creation of Geth network from binary.

joeldudley:


integration/gethnetwork/geth_network.go
```diff
@@ -0,0 +1,140 @@
...
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"strings"
+	"time"
+)
+
+const (
+	genesisFileName = "genesis.json"
+	genesisConfig   = "{\n  \"config\": {\n    \"chainId\": 777,\n    \"homesteadBlock\": 0,\n    \"eip150Block\": 0,\n    \"eip155Block\": 0,\n    \"eip158Block\": 0,\n    \"byzantiumBlock\": 0,\n    \"constantinopleBlock\": 0,\n    \"petersburgBlock\": 0,\n    \"istanbulBlock\": 0,\n    \"berlinBlock\": 0,\n    \"londonBlock\": 0\n  },\n  \"alloc\": {\n    \"0x0000000000000000000000000000000000000001\": {\n      \"balance\": \"111111111\"\n    }\n  },\n  \"coinbase\": \"0x0000000000000000000000000000000000000000\",\n  \"difficulty\": \"0x20000\",\n  \"extraData\": \"\",\n  \"gasLimit\": \"0x2fefd8\",\n  \"nonce\": \"0x0000000000000042\",\n  \"mixhash\": \"0x0000000000000000000000000000000000000000000000000000000000000000\",\n  \"parentHash\": \"0x0000000000000000000000000000000000000000000000000000000000000000\",\n  \"timestamp\": \"0x00\"\n}"
```
otherview: Can we use a backtick ? Makes it more readable :)

integration/gethnetwork/geth_network.go
```diff
@@ -0,0 +1,140 @@
...
+	genesisFilePath := nodesDir + string(os.PathSeparator) + genesisFileName
+	err = os.WriteFile(genesisFilePath, []byte(genesisConfig), 0o600)
+	if err != nil {
+		panic(err)
+	}
+
+	// Each Geth node needs its own data directory.
+	dataDirs := make([]string, numNodes)
+	for i := 0; i < numNodes; i++ {
+		dataDirs[i] = fmt.Sprintf(nodesDir+nodeFolderName, i+1)
```
otherview, line 85: Handy to use os path join for filesystem path appends :)

integration/gethnetwork/geth_network.go
```diff
@@ -0,0 +1,140 @@
...
+
+	network := GethNetwork{
+		gethBinaryPath:  gethBinaryPath,
+		genesisFilePath: genesisFilePath,
+		dataDirs:        dataDirs,
+	}
+
+	for i, dataDir := range dataDirs {
+		_ = os.Remove(dataDir + ipcFileName) // We delete leftover IPC files from previous runs.
+		go network.createGethNode(dataDir, 30303+i)
```
otherview: Are we piping the logs somewhere ? I may have missed that.

Should we surface errors here ? Like if the node dies for some reason.

integration/gethnetwork/geth_network.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// IssueCommand sends the command via RPC to the nodeIdx'th node in the network.
+func (network *GethNetwork) IssueCommand(nodeIdx int, command string) string {
+	dataDir := network.dataDirs[nodeIdx]
+
+	args := []string{dataDirFlag, dataDir, attachCmd, dataDir + ipcFileName, execFlag, command}
+	cmd := exec.Command(network.gethBinaryPath, args...) // nolint
+
+	output, err := cmd.Output()
```
otherview, line 112: What if these two methods just called the IssueCommand ?

integration/gethnetwork/geth_network.go
```diff
@@ -0,0 +1,172 @@
...
+	// Each Geth node needs its own data directory.
+	dataDirs := make([]string, numNodes)
+	for i := 0; i < numNodes; i++ {
+		dataDirs[i] = fmt.Sprintf(nodesDir+nodeFolderName, i+1)
+	}
+
+	network := GethNetwork{
+		gethBinaryPath:  gethBinaryPath,
+		genesisFilePath: genesisFilePath,
+		dataDirs:        dataDirs,
```
otherview, line 91: Can we have a comment explaining why we have to init and then start please ?

integration/gethnetwork/geth_network_test.go
```diff
@@ -0,0 +1,25 @@
...
+	"testing"
+)
+
+const (
+	gethBinaryPath = "path/to/geth/binary"
+	numNodes       = 5
+	peerCountCmd   = "net.peerCount"
+)
+
+func TestCreateNetworkAndIssueCommands(t *testing.T) {
```
otherview, line 17: Sugestion: Test the genesis params.

otherview (commented):
Really good stuff, this will be super handy with running the simulation with real nodes. Nice one Joel!

integration/gethnetwork/geth_network.go
joeldudley: Fixed.

integration/gethnetwork/geth_network.go
joeldudley: I don't think it would work. `IssueCommand` sends an `attach` command to the node, which is followed by a string command. For initialising and starting a node, we need to send a series of flags, etc.

There is also the issue that for `IssueCommand`, we want to wait for the output. For starting the node, we want to run the command but not wait for it to finish (as it spins forever).

integration/gethnetwork/geth_network.go
joeldudley: Added.

integration/gethnetwork/geth_network.go
joeldudley: Thanks, couldn't find the right lib.

otherview (approved):

PR #100, opened 2022-04-27
Log out IP for the Azure VM in deploy script

BedrockSquirrel:
Tiny PR to log out the IP of the new VM in the azure_deployer script for convenience. (And so I can test out the github PR process!).

### DIFF
--- a/tools/azuredeployer/main/azure_deployer.go
+++ b/tools/azuredeployer/main/azure_deployer.go
@@ -149,7 +149,7 @@ func runSetupScript(ipAddress string) {
 		Timeout:         sshTimeout,
 	}
 
-	log.Print("SSH'ing into VM to complete set-up.")
+	log.Printf("SSH'ing into VM to complete set-up: %s:%s", ipAddress, sshPort)
 	var client *ssh.Client
 	var err error
 	for {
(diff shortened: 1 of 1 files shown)


BedrockSquirrel:
Just created a VM with it and can see the IP in the output:
Screenshot 2022-04-27 at 10 30 01

otherview (approved):

tools/azuredeployer/main/azure_deployer.go
```diff
@@ -149,7 +149,7 @@ func runSetupScript(ipAddress string) {
 		Timeout:         sshTimeout,
 	}
 
-	log.Print("SSH'ing into VM to complete set-up.")
+	log.Printf("SSH'ing into VM to complete set-up: %s:%s", ipAddress, sshPort)
```
joeldudley, line 152: I think we should put this info on line 176, where we log `VM set-up complete.`. If the VM takes a while to spin up, this will get buried under the `Waiting for VM to be ready...` messages.

BedrockSquirrel:
Updated to output at the end of the script and squashed it

joeldudley (approved):
LGTM

PR #101, opened 2022-04-28
bump

otherview:
Bumping the efficiency.

My guess is that the github action is running on some VM which the resources are scarce.
Because resources are scarce, running inmemory simulations is impacted by a running ganache.


joeldudley (approved):
Lgtm

PR #117, opened 2022-05-10
Ignores Geth binary.

joeldudley:


otherview (changes requested):
We shouldn't need this, the binary should be under the `.build` directory

PR #121, opened 2022-05-10
Shows how a Geth webserver can be spun up serving custom APIs.

joeldudley:


go/obscuronode/host/clientserver/client_server.go
```diff
@@ -0,0 +1,70 @@
...
+		},
+	}
+	clientServerNode.RegisterAPIs(rpcAPIs)
+
+	return clientServerImpl{
+		node: clientServerNode,
+	}
+}
+
+func (server clientServerImpl) Start() {
```
BedrockSquirrel, line 60: Subjective nitpick but I think I'd prefer `s` to `server` here. Don't really like the single letter variable names but I feel like your eye starts to get used to them  in that situation as an equivalent to `this` or `self` haha. Not a blocker to merge though, just wanted to start a vicious debate

PR #123, opened 2022-05-11
Joel/rpc apis

joeldudley:


go/obscuronode/host/client_server_apis.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/obscuronet/obscuro-playground/go/obscuronode/nodecommon"
+)
+
+// ObscuroAPI implements Obscuro-specific JSON RPC operations.
+type ObscuroAPI struct {
+	p2p P2P
+}
+
+func NewObscuroAPI(p2p P2P) *ObscuroAPI {
+	return &ObscuroAPI{p2p: p2p}
```
BedrockSquirrel, line 13: Feels a bit weird to me that the client-server directly needs the P2P service here. Do you see that as temporary? Because eventually I guess I expect it to just be a proxy through to some host internal service. For example, I'd expect a transaction received from a peer to follow the same codepath as a transaction received from a client eventually.

Might be worth making that separation now and just having this as a proxy before too much starts to accumulate in here.

integration/simulation/transaction_injector.go
```diff
@@ -179,7 +180,12 @@ func (m *TransactionInjector) issueRandomTransfers() {
 		signedTx := wallet_mock.SignTx(tx, fromWallet.Key.PrivateKey)
 		encryptedTx := core.EncryptTx(signedTx)
 		m.stats.Transfer()
-		m.rndL2Node().P2p.BroadcastTx(encryptedTx)
+
+		err := (*m.rndL2NodeClient()).Call(nil, obscuroclient.RPCSendTransactionEncrypted, encryptedTx)
+		if err != nil {
+			continue
```
BedrockSquirrel, line 186: Shouldn't we at least log transactions that failed to inject or is there some reason we'll expect to see regular failures?

BedrockSquirrel (commented):
Generally all looks good to me other than that. The simulation duplication painfully visible in this PR...

integration/simulation/transaction_injector.go
joeldudley: I reckon we'll see a few after the clients/nodes have been stopped. This will be picked up anyway in the tests at the end.

go/obscuronode/host/client_server_apis.go
joeldudley: I don't understand. That's exactly what happens. `SendTransactionEncrypted` proxies the transaction straight on to the P2P service:

    api.p2p.BroadcastTx(encryptedTx)

integration/simulation/transaction_injector.go
joeldudley: I'll add some logging locally, and see how noisy it is.

go/obscuronode/host/client_server_apis.go
BedrockSquirrel: In this code it goes

rpcCall -> rpcServer -> p2pBroadcast

I think it should go

rpcCall -> rpcServer -> someInternalTransactionHandler -> p2pBroadcast

Because I think otherwise this rpcServer has to know about everything. When a transaction arrives and it wants to check it against a blacklist, decide whether to send it the enclave, decide whether to broadcast it to to peers that logic would all go there.

Which wouldn't be too bad except the p2p incoming message also has to decide whether to send it to the enclave and whether to broadcast to its peers. So does that logic get duplicated and both p2pService and clientService know about everything the host does?

integration/simulation/transaction_injector.go
joeldudley: It's not noisy at all actually - logs added :)

BedrockSquirrel: Yeah fair, can always dial it back but at least this way if it breaks unexpectedly it'll be easier to debug than tests randomly failing

go/obscuronode/host/client_server_apis.go
BedrockSquirrel: As discussed, we can park this for now and revisit if it becomes a spaghetti/bloat issue when the host design is a bit more clear. Too hard to imagine its workflows clearly at the moment I think.

BedrockSquirrel (approved):
LGTM! This is going to be super helpful getting that separation for testing etc.

PR #127, opened 2022-05-12
Moves simulation to only interacting with Obscuro nodes via RPC interface

joeldudley:


BedrockSquirrel (commented):
The changes look good to me - I'm getting intermittent failures when I run the in-memory test locally though, have you seen that? Could be environmental maybe, but when it fails I'm seeing a bunch of duplicate failures like this:

Found Rollup duplicates: map[0xfbd3c54fdbd3f8f06186f528b79e5799252e638b94db89b9b4d1a9b6bb07f89e:2]

joeldudley:
> 

@BedrockSquirrel Nope, haven't been able to replicate that locally unfortunately. Ran it three times this morning and always fine.

BedrockSquirrel (approved):
LGTM - happy to approve now you've got the build more stable. We can investigate further build issues as we go but don't think this has made the build significantly flakier.

PR #128, opened 2022-05-14
Test to ensure node and enclave can be run as standalone processes

joeldudley:


go/obscuronode/obscuroclient/obscuro_client.go
```diff
@@ -36,14 +34,14 @@ type clientImpl struct {
 	rpcClient *rpc.Client
 }
 
-func NewClient(nodeID int64, address string) Client {
+func NewClient(nodeID common.Address, address string) Client {
```
otherview, line 37: Should this be a pointer ?

otherview (commented):
The runner architecture is sound.
But the noderunner should be a simulation so to go in line with the simulation we have.

PR #130, opened 2022-05-16
Provides more informative RPC error messages in the simulation.

joeldudley:


BedrockSquirrel (approved):
LGTM, bit repetitive but feel like that's Go's fault.

PR #131, opened 2022-05-16
Adds RPC method to stop a host remotely. Uses this to stop hosts in sims.

joeldudley:


go/obscuronode/host/host.go
```diff
@@ -142,6 +136,14 @@ func NewObscuroAggregator(
 
 		txHandler: txHandler,
 	}
+
+	var clientServer ClientServer
+	if cfg.HasRPC {
+		clientServer = NewClientServer(*cfg.RPCAddress, &host)
```
BedrockSquirrel, line 142: Minor thing but is there any need for the clientServer var here? I.e. could it just be

```
	if cfg.HasRPC {
		host.clientServer = NewClientServer(*cfg.RPCAddress, &host)
	}
```

BedrockSquirrel (approved):
LGTM. Yet more simulation wiring merge conflicts to fix on my branch but serves me right for letting it live for so long :)

ClientServer taking the whole host makes that ever growing list of params much nicer to work with.

go/obscuronode/host/host.go
joeldudley: Cool, good catch.

PR #132, opened 2022-05-16
Logs everything to stdout in CI.

joeldudley:


integration/simulation/utils.go
```diff
@@ -12,9 +12,17 @@ import (
...
 )
 
-const testLogs = "../.build/simulations/"
+const (
+	envLogToStdout = "LOG_TO_STDOUT"
+	testLogs       = "../.build/simulations/"
+)
+
+func setupTestLog() {
+	_, isPresent := os.LookupEnv(envLogToStdout)
```
otherview, line 21: Should probably check if the value is true/false rather than if the var it's set or not

otherview (changes requested):

integration/simulation/utils.go
joeldudley: I chose checking for presence to avoid the situation where there's a misunderstanding about what this should be set to (e.g. CI sets it to "1", but we check for "true"). But that creates its own problems. Changed :)

PR #135, opened 2022-05-16
removing build tags

otherview:
Removes build tags and uses env vars to run specific tests


integration/simulation/simulation_docker_test.go
```diff
@@ -25,19 +23,25 @@ import (
...
 	enclaveDockerPort = "11000/tcp"
+	dockerTestEnv     = "DOCKER_TEST_ENABLED"
 )
 
 // This test creates a network of L2 nodes, then injects transactions, and finally checks the resulting output blockchain
 // The L2 nodes communicate with each other via sockets, and with their enclave servers via RPC.
 // All nodes live in the same process, the enclaves run in individual Docker containers, and the Ethereum nodes are mocked out.
 // $> docker rm $(docker stop $(docker ps -a -q --filter ancestor=obscuro_enclave --format="{{.ID}}") will stop and remove all images
 func TestDockerNodesMonteCarloSimulation(t *testing.T) {
+	fooAddr := os.Getenv(dockerTestEnv)
```
joeldudley, line 40: "fooAddr"?

.github/workflows/build-push-docker.yml
```diff
@@ -27,4 +27,4 @@ jobs:
         run:  docker build -t obscuro_enclave -f dockerfiles/enclave_local_build.Dockerfile .
 
       - name: Test
-        run: go test -v -tags docker -run TestDocker ./...
+        run: DOCKER_TEST_ENABLED=true go test -v -run TestDocker ./...
```
joeldudley, line 30: Isn't it better practice to use an `env:` block here?

integration/simulation/simulation_docker_test.go
otherview: changing it to something more meaningful :)

.github/workflows/build-push-docker.yml
otherview: Think for one var I wouldn't bother, but I can change that 👍

tudor-malene (commented):
needs some readme instructions on how to run these skipped tests locally.
Otherwise looks good to me if it works

joeldudley (approved):
LGTM

PR #137, opened 2022-05-17
Creates main method for creating Geth network. Small improvements to host and enclave mains.

joeldudley:


go/obscuronode/enclave/enclaverunner/enclave_runner.go
```diff
@@ -26,6 +27,7 @@ func RunEnclave(config EnclaveConfig) {
 	if err := enclave.StartServer(config.Address, nodeAddress, txHandler, false, nil, nil); err != nil {
 		panic(err)
 	}
+	fmt.Println("Obscuro enclave service started.")
```
otherview, line 30: Should we log instead ?

go/obscuronode/host/hostrunner/cli.go
```diff
@@ -95,14 +95,20 @@ func ParseCLIArgs() HostConfig {
 
 	flag.Parse()
 
+	parsedP2PAddrs := strings.Split(*peerP2PAddrs, ",")
```
otherview, line 98: Much better

go/obscuronode/host/hostrunner/host_runner.go
```diff
@@ -13,6 +14,8 @@ import (
 	"github.com/obscuronet/obscuro-playground/go/obscuronode/host/p2p"
 )
 
+const logPath = "host_logs.txt"
```
otherview, line 17: Should this be in the config ? 
What if the defaults was the .build dir ?

go/obscuronode/host/hostrunner/host_runner.go
```diff
@@ -28,21 +31,24 @@ func RunHost(config HostConfig) {
 	nodeWallet := wallet.NewInMemoryWallet(config.PrivateKeyString)
 	contractAddr := common.HexToAddress(config.ContractAddress)
 	txHandler := mgmtcontractlib.NewEthMgmtContractTxHandler(contractAddr)
+
+	fmt.Println("Connecting to L1 network...")
```
otherview, line 35: Should we log instead ?

integration/gethnetwork/main/main.go
```diff
@@ -0,0 +1,21 @@
...
+// Spins up a new Geth network.
+func main() {
+	config := ParseCLIArgs()
+
+	gethBinaryPath, err := gethnetwork.EnsureBinariesExist(gethnetwork.LatestVersion)
+	if err != nil {
+		panic(err)
+	}
+	gethnetwork.NewGethNetwork(config.StartPort, gethBinaryPath, config.NumNodes, 1, config.PrefundedAddrs)
+	fmt.Println("Geth network started.")
```
otherview, line 18: Log maybe ?

integration/gethnetwork/main/main.go
```diff
@@ -0,0 +1,21 @@
...
+
+	select {}
```
otherview, line 20: Should we read for a CTRL+C / Interrupt and do a clean exit ?

Using this will leave zombie processes wont it ?

otherview (commented):
Looks good !

I think we should handle the termination properly so to not leave zombie processes.

go/obscuronode/enclave/enclaverunner/enclave_runner.go
joeldudley: I did this on purpose so there's some sign of life when things are pushed to the logs. Otherwise, you can think the process is blocked/dead, but you just haven't checked the logs.

integration/gethnetwork/main/main.go
joeldudley: I did this on purpose so there's some sign of life when things are pushed to the logs. Otherwise, you can think the process is blocked/dead, but you just haven't checked the logs.

go/obscuronode/host/hostrunner/host_runner.go
joeldudley: Yes to making this part of the config.

I don't want the default to be the .build dir though. The default should be what makes sense when it's a standalone binary being run by someone IMO.

integration/gethnetwork/main/main.go
joeldudley: Yeah. I've never done that - any good examples?

go/obscuronode/enclave/enclaverunner/enclave_runner.go
otherview: Fair enough 👍

go/obscuronode/host/hostrunner/host_runner.go
otherview: Do we need a logs folder ? But maybe that's something for the future.

integration/gethnetwork/main/main.go
otherview:
```go
c := make(chan os.Signal, 1)
signal.Notify(c, os.Interrupt)
go func(){
    for sig := range c {
        // sig is a ^C, handle it
    }
}()

```

something like https://stackoverflow.com/questions/11268943/is-it-possible-to-capture-a-ctrlc-signal-sigint-and-run-a-cleanup-function-i

I think that works, maybe just capture other interrupt signals like `syscall.SIGTERM`

otherview (approved):

PR #139, opened 2022-05-17
Runs sims in parallel.

joeldudley:


.github/workflows/build-pr.yml
```diff
@@ -24,4 +24,4 @@ jobs:
         run: bash ./integration/gethnetwork/build_geth_binary.sh --version=v1.10.17
 
       - name: Test
-        run: go test -v ./...
+        run: go test -v ./... -p 4
```
otherview, line 27: Is the `GOMAXPROCS` not set correctly should we override it ?

otherview (commented):
I'd remove the `-p`,  other than that looks good 👍

PR #143, opened 2022-05-18
Sim logs are now labeled based on the test type.

joeldudley:


integration/simulation/utils.go
```diff
@@ -14,13 +14,14 @@ import (
...
 
-func setupTestLog() *os.File {
+func setupTestLog(simType string) *os.File {
 	// create a folder specific for the test
 	err := os.MkdirAll(testLogs, 0o700)
 	if err != nil {
 		panic(err)
 	}
-	f, err := os.CreateTemp(testLogs, fmt.Sprintf("simulation-result-%d-*.txt", time.Now().Unix()))
+	timeFormatted := time.Now().Format("2006-01-02-15-04-05")
```
BedrockSquirrel: It's blowing my mind that this is how you format datetimes in Go haha.

Fwiw could we put a double hypen or an underscore before time for readability (i.e. `2006-01-02--15-04-05`? And maybe make it just "sim-result-" or "sim-log-" or something so they're a little less wide in the project explorer? (Sorry to nitpick but we're all gonna be reading these names 10 times a day haha)

BedrockSquirrel (approved):
LGTM, big QoL improvement :)

PR #151, opened 2022-05-19
Prevents test failure if wallet extension is connected too soon.

joeldudley:
Previously, there was a race condition whereby the wallet extension tests could fire off a request before the wallet extension had opened its port. We add a retry to prevent this causing a test failure.


BedrockSquirrel (approved):
LGTM, you could go pro as a whack-a-mole player

PR #155, opened 2022-05-20
Various changes to improve deployment of dev testnet

joeldudley:


BedrockSquirrel (approved):
LGTM!

Feature request that I think would be useful and not difficult: uptime (and then potentially block rate since you get it for free)

Could calc it front-end if you serve the genesis time in a one-off request, or could serve it as a network stats blob which might be a useful placeholder for more stuff in future.

PR #156, opened 2022-05-20
Allows use of hardcoded genesis to be switched on. Clears up duplication in Geth network.

joeldudley:


go/obscuronode/enclave/enclaverunner/cli.go
```diff
@@ -37,6 +41,7 @@ func DefaultEnclaveConfig() EnclaveConfig {
 		ContractAddress: "",
 		WriteToLogs:     false,
 		LogPath:         "enclave_logs.txt",
+		VerifyBlocks:    false,
```
BedrockSquirrel, line 44: Could we call the flag and variable something a bit less ambiguous, like VerifyL1Genesis?

BedrockSquirrel (changes requested):
Just minor rename request for clarity

go/obscuronode/enclave/enclaverunner/cli.go
joeldudley: It's not about verifying the L1 genesis. You need the genesis to verify L1 blocks (otherwise, you don't know the initial params to use to verify each block). I'll rename it to `VerifyL1Blocks`

BedrockSquirrel (approved):
LGTM

PR #157, opened 2022-05-20
Monitor goroutine leaks

BedrockSquirrel:
Some goroutines are leaking from simulations and are still running when the next simulation starts. This seems to be causing some issues and it would be good to reduce them over time/be able to rule it out as a contributor to flakey builds.

This change just adds logging at the beginning and end of sims to output number of running goroutines and does some refactors that might help (or at least shouldn't hurt). My very unscientific testing suggests it's knocked out about 50 leaked goroutines by the time the 3 sims have finished.

The changes mostly just add defers to try to guarantee cleanups will happen in the case of unexpected errors (reduce red herrings at least, even if the tests fail anyway). And I noticed from the log dump that about 15-20 goroutines were from the monitorBlocks() in host.go so I tweaked that to check for interrupts.

### DIFF
--- a/go/obscuronode/enclave/server.go
+++ b/go/obscuronode/enclave/server.go
@@ -157,8 +157,8 @@ func (s *server) RoundWinner(_ context.Context, request *generated.RoundWinnerRe
 }
 
 func (s *server) Stop(context.Context, *generated.StopRequest) (*generated.StopResponse, error) {
-	err := s.enclave.Stop()
 	defer s.rpcServer.GracefulStop()
+	err := s.enclave.Stop()
 	return &generated.StopResponse{}, err
 }
 
--- a/go/obscuronode/enclave/server.go
+++ b/go/obscuronode/enclave/server.go
@@ -158,7 +158,7 @@ func (s *server) RoundWinner(_ context.Context, request *generated.RoundWinnerRe
 
 func (s *server) Stop(context.Context, *generated.StopRequest) (*generated.StopResponse, error) {
 	err := s.enclave.Stop()
-	s.rpcServer.GracefulStop()
+	defer s.rpcServer.GracefulStop()
 	return &generated.StopResponse{}, err
 }
 
--- a/go/obscuronode/host/host.go
+++ b/go/obscuronode/host/host.go
@@ -520,21 +520,27 @@ func (a *Node) checkForSharedSecretRequests(block obscurocommon.EncodedBlock) {
 func (a *Node) monitorBlocks() {
 	listener := a.ethClient.BlockListener()
 	nodecommon.LogWithID(a.shortID, "Start monitoring Ethereum blocks..")
-	for {
-		latestBlkHeader := <-listener
-		block, err := a.ethClient.FetchBlock(latestBlkHeader.Hash())
-		if err != nil {
-			log.Panic("could not fetch block for hash %s. Cause: %s", latestBlkHeader.Hash().String(), err)
-		}
-		blockParent, err := a.ethClient.FetchBlock(block.ParentHash())
-		if err != nil {
-			log.Panic("could not fetch block's parent with hash %s. Cause: %s", block.ParentHash().String(), err)
-		}
 
-		nodecommon.LogWithID(a.shortID, "Received a new block b_%d(%d)",
-			obscurocommon.ShortHash(latestBlkHeader.Hash()),
-			latestBlkHeader.Number.Uint64())
-		a.RPCNewHead(obscurocommon.EncodeBlock(block), obscurocommon.EncodeBlock(blockParent))
+	for atomic.LoadInt32(a.interrupt) == 0 {
+		select {
+		case latestBlkHeader := <-listener:
+			block, err := a.ethClient.FetchBlock(latestBlkHeader.Hash())
+			if err != nil {
+				log.Panic("could not fetch block for hash %s. Cause: %s", latestBlkHeader.Hash().String(), err)
+			}
+			blockParent, err := a.ethClient.FetchBlock(block.ParentHash())
+			if err != nil {
+				log.Panic("could not fetch block's parent with hash %s. Cause: %s", block.ParentHash().String(), err)
+			}
+
+			nodecommon.LogWithID(a.shortID, "Received a new block b_%d(%d)",
+				obscurocommon.ShortHash(latestBlkHeader.Hash()),
+				latestBlkHeader.Number.Uint64())
+			a.RPCNewHead(obscurocommon.EncodeBlock(block), obscurocommon.EncodeBlock(blockParent))
+
+		case <-time.After(1 * time.Second):
+			// check for interrupt
+		}
 	}
 }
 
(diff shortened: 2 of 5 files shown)


integration/simulation/network/geth_network.go
```diff
@@ -115,11 +115,15 @@ func (n *networkInMemGeth) Create(params *params.SimParams, stats *stats.Stats)
...
-		go (*temp).Stop()
+		go func() {
+			_ = (*temp).Call(nil, obscuroclient.RPCStopHost)
+			(*temp).Stop()
+		}()
 	}
 
 	n.gethNetwork.StopNodes()
+	// double tap. todo: this is probably unnecessary
+	defer n.gethNetwork.StopNodes()
```
joeldudley, line 126: @BedrockSquirrel Is there any point `defer`ring the final call in a method?

joeldudley, line 126: @BedrockSquirrel I don't think it makes sense to defer  the final call in a method. This would make sense if you deferred `defer n.gethNetwork.StopNodes()` at the top of the method, in case stopping one of the clients blew up.

integration/simulation/network/geth_network.go
```diff
@@ -115,11 +115,15 @@ func (n *networkInMemGeth) Create(params *params.SimParams, stats *stats.Stats)
...
 		temp := client
-		go (*temp).Call(nil, obscuroclient.RPCStopHost) //nolint:errcheck
-		go (*temp).Stop()
+		go func() {
+			_ = (*temp).Call(nil, obscuroclient.RPCStopHost)
+			(*temp).Stop()
+		}()
 	}
 
 	n.gethNetwork.StopNodes()
```
joeldudley, line 124: @BedrockSquirrel Why twice?

integration/simulation/network/geth_network.go
```diff
@@ -115,11 +115,15 @@ func (n *networkInMemGeth) Create(params *params.SimParams, stats *stats.Stats)
 func (n *networkInMemGeth) TearDown() {
 	for _, client := range n.obscuroClients {
 		temp := client
-		go (*temp).Call(nil, obscuroclient.RPCStopHost) //nolint:errcheck
-		go (*temp).Stop()
+		go func() {
+			_ = (*temp).Call(nil, obscuroclient.RPCStopHost)
+			(*temp).Stop()
```
joeldudley, line 120: @BedrockSquirrel Would it be better to do:

```
defer (*temp).Stop()
_ = (*temp).Call(nil, obscuroclient.RPCStopHost)
```

That way, if RPCStopHost blows up, we'll still stop the client.

integration/simulation/network/inmemory.go
```diff
@@ -82,8 +82,10 @@ func (n *basicNetworkOfInMemoryNodes) Create(params *params.SimParams, stats *st
 func (n *basicNetworkOfInMemoryNodes) TearDown() {
 	for _, client := range n.obscuroClients {
 		temp := client
-		go (*temp).Call(nil, obscuroclient.RPCStopHost) //nolint:errcheck
-		go (*temp).Stop()
+		go func() {
+			_ = (*temp).Call(nil, obscuroclient.RPCStopHost)
```
joeldudley, line 86: @BedrockSquirrel Same comment as above.

integration/simulation/simulation_tester.go
```diff
@@ -45,5 +49,5 @@ func testSimulation(t *testing.T, netw network.Network, params *params.SimParams
 
 	// generate and print the final stats
 	t.Logf("Simulation results:%+v", NewOutputStats(&simulation))
-	netw.TearDown()
+	log.Info("goroutine leak monitor - simulation end - %d go routines currently running", runtime.NumGoroutine())
```
joeldudley, line 52: @BedrockSquirrel Isn't it misleading to log the remaining goroutines before stopping the network? What if we did the following?

```
defer go() {
    netw.TearDown()
	log.Info("goroutine leak monitor - simulation end - %d go routines currently running", runtime.NumGoroutine())
}()
```

go/obscuronode/enclave/server.go
```diff
@@ -159,7 +159,7 @@ func (s *server) RoundWinner(_ context.Context, request *generated.RoundWinnerRe
 
 func (s *server) Stop(context.Context, *generated.StopRequest) (*generated.StopResponse, error) {
 	err := s.enclave.Stop()
-	s.rpcServer.GracefulStop()
+	defer s.rpcServer.GracefulStop()
```
joeldudley, line 162: @BedrockSquirrel This should be further up:

```
defer s.rpcServer.GracefulStop()
err := s.enclave.Stop()
```

joeldudley (changes requested):
Some requested changes. I think we have a differing understanding of `defer` :D

go/obscuronode/host/host.go
```diff
@@ -520,21 +520,27 @@ func (a *Node) checkForSharedSecretRequests(block obscurocommon.EncodedBlock) {
...
+				log.Panic("could not fetch block's parent with hash %s. Cause: %s", block.ParentHash().String(), err)
+			}
+
+			nodecommon.LogWithID(a.shortID, "Received a new block b_%d(%d)",
+				obscurocommon.ShortHash(latestBlkHeader.Hash()),
+				latestBlkHeader.Number.Uint64())
+			a.RPCNewHead(obscurocommon.EncodeBlock(block), obscurocommon.EncodeBlock(blockParent))
+
+		case <-time.After(1 * time.Second):
+			// check for interrupt
```
joeldudley, line 542: @BedrockSquirrel Can we explain this better? I guess you're saying we only want to attempt to fetch new blocks every second?

integration/simulation/simulation_tester.go
BedrockSquirrel: Yeah good spot. This might help get the numbers down haha.

integration/simulation/simulation_tester.go
```diff
@@ -2,25 +2,33 @@ package simulation
...
+	"github.com/obscuronet/obscuro-playground/go/log"
 	stats2 "github.com/obscuronet/obscuro-playground/integration/simulation/stats"
 
 	"github.com/google/uuid"
 )
 
 // testSimulation encapsulates the shared logic for simulating and testing various types of nodes.
 func testSimulation(t *testing.T, netw network.Network, params *params.SimParams) {
+	defer func() {
+		// wait until clean up is complete before we log the lingering goroutine count
```
joeldudley, line 21: @BedrockSquirrel Are you sure `defer netw.TearDown()` is called before this `defer`? It's not clear to me it would be. I think you need:

```
defer func() {
        // wait until clean up is complete before we log the lingering goroutine count
	defer log.Info("goroutine leak monitor - simulation end - %d goroutines currently running", runtime.NumGoroutine())
        netw.TearDown()
}()
```

BedrockSquirrel: @joeldudley yeah defers are guaranteed reverse order (they're put on a LIFO stack https://go.dev/tour/flowcontrol/13 - I guess because the first resource open is often the last to be closed).

So thought it might be simpler to keep it with the other log message and to avoid ordering risk if someone ever added another cleanup defer to the simulation runner. But don't feel strongly, happy to change if you prefer.

joeldudley (approved):
lgtm

PR #160, opened 2022-05-22
Fail fast in geth nodes test when ports are not available

BedrockSquirrel:
This just adds a crude check of all the ports that the geth binaries will be running with are available before the simulation network is started and fails quickly if not.

I don't know if this is a big problem or if this is the best way to deal with it but it's bitten me multiple times locally and it would be nice to rule it out on the build servers.

I managed to get myself into the bad state somehow locally, can see the processes lurking here while no test is running:
Screenshot 2022-05-22 at 20 56 00

And on this branch the test fails immediately with:
Screenshot 2022-05-22 at 20 56 13

Then I resolve it with something like:
Screenshot 2022-05-22 at 21 03 24

And the test starts running fully again.

### DIFF
--- a/integration/gethnetwork/geth_network.go
+++ b/integration/gethnetwork/geth_network.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"os"
 	"os/exec"
 	"path"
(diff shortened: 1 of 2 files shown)


integration/gethnetwork/geth_network.go
```diff
@@ -353,3 +356,39 @@ func waitForIPC(dataDir string) {
 		counter++
 	}
 }
+
+func EnsurePortsAreAvailable(startPort int, numberNodes int) error {
```
joeldudley, line 360: @BedrockSquirrel I feel like this should be called as part of `NewGethNetwork`, instead of being a public function. In what cases would you not want the ports to be available?

integration/gethnetwork/geth_network.go
```diff
@@ -353,3 +356,39 @@ func waitForIPC(dataDir string) {
...
+	var unavailablePorts []int
+	mu := &sync.Mutex{}
+	wg := &sync.WaitGroup{}
+
+	for i := 0; i < numberNodes; i++ {
+		wg.Add(2)
+		go ensurePortAvailable(wg, mu, &unavailablePorts, startPort+i)              // commsPort
```
joeldudley, line 367: @BedrockSquirrel Do we gain much by parallelising here? Seems to add complexity for little benefit AFAICT.

BedrockSquirrel: Yeah agreed, I'll simplify. Felt like they were IO operations but actually how long does it take to listen and then immediately stop listening on a port. Was a useful exercise for me though, hadn't really dabbled in the black art of Go concurrency.

integration/gethnetwork/geth_network.go
BedrockSquirrel: @joeldudley Yeah that's fair, will change. I was following the pattern of EnsureBinariesExist() but hadn't clocked that that actually provides the path for the wiring too.

joeldudley (approved):
LGTM

PR #163, opened 2022-05-23
Removes double pointer

otherview:


joeldudley (approved):
LGTM

otherview:
Should we be using eth rpc package for this ? Not sure whats the upside @joeldudley

otherview:
Can't assign without pointer dereferencing, no point changing this

PR #165, opened 2022-05-23
Various changes to allow tx injection into dev testnet

joeldudley:


integration/gethnetwork/geth_network_test.go
```diff
@@ -35,7 +35,8 @@ func TestGethAllNodesJoinSameNetwork(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	network := NewGethNetwork(getStartPort(), gethBinaryPath, numNodes, 1, nil)
+	startPort := getStartPort()
+	network := NewGethNetwork(startPort, startPort+100, gethBinaryPath, numNodes, 1, nil)
```
BedrockSquirrel, line 39: It feels a bit of a shame that this 100 offset is magically known in so many places (harder to read/change). Worth keeping as a public constant somewhere perhaps (i.e. DefaultWsPortOffset)?

BedrockSquirrel (commented):
One minor thought about the 100 magic number

PR #166, opened 2022-05-23
The socket sim now uses a Geth network, to create a sim that is as realistic as possible.

joeldudley:


integration/simulation/network/socket.go
```diff
@@ -96,14 +89,11 @@ func (n *networkOfSocketNodes) Create(params *params.SimParams, stats *stats.Sta
 }
 
 func (n *networkOfSocketNodes) TearDown() {
+	defer n.gethNetwork.StopNodes()
+
 	for _, client := range n.obscuroClients {
 		temp := client
 		go (*temp).Call(nil, obscuroclient.RPCStopHost) //nolint:errcheck
```
BedrockSquirrel, line 96: I know you didn't change this but I think there's a race condition here which it's worth fixing while you're reworking the test. In the equivalent clean-up for geth-network we did it like this:

```
		go func() {
			defer (*temp).Stop()
			_ = (*temp).Call(nil, obscuroclient.RPCStopHost)
		}()
```

BedrockSquirrel (commented):
One comment that it'd be good to fix while you're there. Then happy to approve if you can convince the build to pass :)

BedrockSquirrel (approved):
LGTM

PR #174, opened 2022-05-24
Use a single config object across the host

joeldudley:


go/obscuronode/host/host.go
```diff
@@ -140,8 +118,8 @@ func NewObscuroAggregator(
 		ethWallet:       ethWallet,
 	}
 
-	if cfg.HasRPC {
-		host.clientServer = NewClientServer(*cfg.RPCAddress, &host)
+	if config.HasClientRPC {
+		host.clientServer = NewClientServer(config.ClientRPCAddress, &host)
 	}
```
BedrockSquirrel, line 124: At my last place there was loads of config and whenever a process started it dumped the config to the log, was super helpful when you were debugging an issue to be able to see the full config in one place (useful for making sure it had been run properly for example).

Worth dumping the config out to the log here or in `host.Start()` potentially?

integration/simulation/network/network_utils.go
```diff
@@ -4,20 +4,21 @@ import (
 	"math/big"
 	"time"
 
+	config2 "github.com/obscuronet/obscuro-playground/go/obscuronode/config"
+
+	"github.com/obscuronet/obscuro-playground/go/ethclient/erc20contractlib"
+	"github.com/obscuronet/obscuro-playground/go/obscurocommon"
+	"github.com/obscuronet/obscuro-playground/go/obscuronode/enclave"
 	"github.com/obscuronet/obscuro-playground/integration"
+	p2p2 "github.com/obscuronet/obscuro-playground/integration/simulation/p2p"
```
BedrockSquirrel, line 13: Again not a big fan of these "2"s. Like this one could be called sim_p2p or something perhaps so reader might have a sense of where it's coming from/why it's a different import

integration/simulation/transaction_injector.go
```diff
@@ -235,7 +235,7 @@ func (m *TransactionInjector) issueInvalidL2Txs() {
 			toWallet = m.rndObsWallet()
 		}
 		var tx types.TxData
-		switch rand.Intn(1) { //nolint:gosec
+		switch rand.Intn(2) { //nolint:gosec
```
BedrockSquirrel, line 238: Is this an unrelated bugfix? Don't want to be fussy but could maybe drop it in a separate PR. Feels like the cleaner thing to do, then if tests start behaving very differently it'll be more obvious whether it was this or the config changes.

BedrockSquirrel (changes requested):
A few comments (that I'm not necessarily right about haha, haven't had any caffeine yet)

go/obscuronode/host/l1_client.go
```diff
@@ -1,11 +1,15 @@
-package ethclient
+package host
```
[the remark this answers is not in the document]
joeldudley: No, that wasn't the reason, as the config is itself in its own package.

In a way, this felt to me like it should be part of `host`, because it's how the host connects to an L1 network. But I don't feel strongly, so I'll move it back.

go/obscuronode/host/host.go
joeldudley: Yeah, but I'll do that as a follow-up.

joeldudley: Added a todo.

integration/simulation/transaction_injector.go
joeldudley: I would do, but it's breaking the linter, so I have to fix it (the alternative is to suppress it for now, but that feels worse).

integration/simulation/network/network_utils.go
joeldudley: Fixed (didn't add an underscore as Goland doesn't like it).

go/obscuronode/config/host_config.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"math/big"
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// HostConfig contains the full configuration for an Obscuro host.
+type HostConfig struct {
```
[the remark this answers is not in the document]
BedrockSquirrel: As discussed you can disregard this. Something smells not quite right but we're not sure what it is and might be nothing - see how we go

go/obscuronode/host/l1_client.go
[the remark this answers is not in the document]
BedrockSquirrel: Yeah fair, I think the danger would be that host would be a gigantic ball of spaghetti if we put all of its code in the same host package.

If we have part in their own packages eventually then host itself will just be wiring together the `ethclient`, `p2p`, `enclaveclient` etc. and their logic can be self contained. Like by moving that in you added a dependency on geth to host but now it doesn't need to know how we connect to the L1.

PR #177, opened 2022-05-25
Deletes Geth repo after building binary.

joeldudley:
Its files kept showing up in my search results, which was annoying.

### DIFF
--- a/integration/gethnetwork/build_geth_binary.sh
+++ b/integration/gethnetwork/build_geth_binary.sh
@@ -48,6 +48,7 @@ fi
 # Make sure .build folder exists
 mkdir -p "${build_path}"
 
+# Only download geth code if binary does not exist
 if [ -f "${geth_path}/geth-${geth_version}" ]
 then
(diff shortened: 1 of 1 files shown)


integration/gethnetwork/build_geth_binary.sh
```diff
@@ -48,6 +48,12 @@ fi
 # Make sure .build folder exists
 mkdir -p "${build_path}"
 
+if [ -f "${geth_path}/geth-${geth_version}" ]
```
otherview, line 51:
```suggestion
# Only download geth code if binary does not exist
if [ -f "${geth_path}/geth-${geth_version}" ]
```

integration/gethnetwork/build_geth_binary.sh
```diff
@@ -48,6 +48,12 @@ fi
...
+then
+    echo "Skipping geth build - Found binary at ${geth_path}/geth-${geth_version}"
+    exit
```
otherview, line 54:
```suggestion
    exit 0
```

I think needs to be exit 0

PR #183, opened 2022-05-25
Better stopping of Geth nodes.

joeldudley:


integration/gethnetwork/geth_network.go
```diff
@@ -255,11 +255,25 @@ func (network *GethNetwork) IssueCommand(nodeIdx int, command string) string {
 
 // StopNodes kills the Geth node processes.
 func (network *GethNetwork) StopNodes() {
+	var wg sync.WaitGroup
 	for _, process := range network.nodesProcs {
 		if process != nil {
-			_ = process.Kill()
+			tempProcess := process
+			wg.Add(1)
+			go func() {
```
BedrockSquirrel, line 263: I think the idiomatic way of dealing with the closure here is to make process an argument of the anonymous function. I keep seeing it everywhere, i.e instead of creating `tempProcess`:

```
go func(process os.Process) {
...
}(process)

Or is there some reason that doesn't work here?

File: integration/gethnetwork/geth_network.go [hunk 2]
```diff
@@ -255,11 +255,25 @@ func (network *GethNetwork) IssueCommand(nodeIdx int, command string) string {
...
+				defer wg.Done()
+				err := tempProcess.Kill()
+				if err != nil {
+					log.Error("geth node could not be killed: %s", err)
+				}
+				_, err = tempProcess.Wait()
+				if err != nil {
+					log.Error("geth node was killed successfully but did not exit: %s", err)
```
BedrockSquirrel, line 271: Worth sending another kill signal in this case?

BedrockSquirrel (commented):
Minor thoughts

integration/gethnetwork/geth_network.go
joeldudley: Thanks, fixed.

integration/gethnetwork/geth_network.go
joeldudley: I tend to think if one didn't work, why would two?

PR #188, opened 2022-05-25
dont executed workflow on pull requests

otherview:
According to [this](https://github.community/t/self-hosted-runner-security-with-public-repositories/17860/15), this change should be enough to prevent forked repos to run actions


otherview:
No need as the settings have been changed 👍

PR #196, opened 2022-05-26
Shuts down node runner test.

joeldudley:


integration/noderunner/noderunner_test.go
```diff
@@ -60,9 +60,12 @@ func TestCanStartStandaloneObscuroHostAndEnclave(t *testing.T) {
 		t.Fatal(err)
 	}
 	network := gethnetwork.NewGethNetwork(int(gethPort), int(gethWebsocketPort), gethBinaryPath, 1, 1, []string{address.String()})
-	defer network.StopNodes()
 	go enclaverunner.RunEnclave(enclaveConfig)
 	go hostrunner.RunHost(hostConfig)
+	obscuroClient := obscuroclient.NewClient(clientServerAddr)
+
+	defer network.StopNodes()
```
BedrockSquirrel, line 67: I think best to leave this on the line below the network creation, seems to be the done thing to pair creation and tear-down for clarity and makes sure it gets cleaned up if NewClient() panics for example.

integration/noderunner/noderunner_test.go
```diff
@@ -60,9 +60,12 @@ func TestCanStartStandaloneObscuroHostAndEnclave(t *testing.T) {
...
+	defer obscuroClient.Call(nil, obscuroclient.RPCStopHost) //nolint:errcheck
```
BedrockSquirrel, line 68: I wonder if this is an example of where we could get into trouble between tests.

That call will return immediately now and then we plow onwards but if teardown took some time then shenanigans could ensue with the port not getting freed straight away. Worth waiting for the connection to die before exiting the defer or am I being paranoid?

BedrockSquirrel (commented):
Couple of minor points to consider

integration/noderunner/noderunner_test.go
joeldudley: The problem I found was that you can't wait for the stop call to finish, since the stop call shuts down the RPC server itself, preventing a response from ever coming back.

I could wait until the port is released in a loop, if you think that's a good idea?

integration/noderunner/noderunner_test.go
joeldudley: 👍

joeldudley: @BedrockSquirrel I remembered - the reason I did it this way is that it makes more sense to stop the Obscuro side first, and _then_ the L1 side, since L1 is the server and Obscuro the client in this relationship.

joeldudley: Ah ignore, it goes in the other direction, right? FIFO?

integration/noderunner/noderunner_test.go
BedrockSquirrel: Yeah it's not elegant but I think it's worth doing just to give us peace of mind that it's not a possibility.

integration/noderunner/noderunner_test.go
BedrockSquirrel: @joeldudley Yeah it's LIFO, so they're executed in reverse order. But you haven't changed the order of the deferrals by moving the network.StopNodes up, before my comment the L1 teardown deferral was before the L2 deferral and it still is (so both then and now it tears down L2 first).

integration/noderunner/noderunner_test.go
```diff
@@ -96,6 +96,20 @@ func TestCanStartStandaloneObscuroHostAndEnclave(t *testing.T) {
...
 }
 
+func teardown(obscuroClient obscuroclient.Client, clientServerAddr string) {
+	obscuroClient.Call(nil, obscuroclient.RPCStopHost) //nolint:errcheck
+
+	// We wait for the client server port to be closed.
+	wait := 20 // max wait in seconds
+	for tcpConnectionAvailable(clientServerAddr) {
+		if wait == 0 {
+			panic(fmt.Sprintf("RPC client server had not shut down after %d seconds", wait))
```
BedrockSquirrel, line 106: Panic message will show "after 0 seconds" slightly confusingly :)

go/obscuronode/host/host.go
```diff
@@ -198,16 +198,17 @@ func (a *Node) Stop() {
...
-	if a.clientServer != nil {
-		a.clientServer.Stop()
-	}
 
 	if err := a.EnclaveClient.Stop(); err != nil {
 		nodecommon.LogWithID(a.shortID, "Could not stop enclave server. Cause: %v", err.Error())
 	}
 	time.Sleep(time.Second)
 	a.exitNodeCh <- true
 	a.EnclaveClient.StopClient()
```
BedrockSquirrel, line 207: Maybe not as part of this PR but this Stop() method feels like it could be a bit more robust with careful defers. If P2p.StopListening() panicked, for example, then EnclaveClient and ClientServer might be left running.

integration/noderunner/noderunner_test.go
joeldudley: Haha good catch.

BedrockSquirrel (approved):
LGTM

PR #202, opened 2022-05-27
Add config to turn-on persistent enclave database

BedrockSquirrel:
As a pre-cursor to the edgeless DB work we need a way to tell nodes what DB setup to use. I think there are 3 situations worth using in testing:

- in-memory (quick and clean, default for test scenarios)
- persistent, SQL (integration tests to verify the persistence and the SQL implementation of ethdb key-value store work)
- persistent in-enclave, Edgeless DB (dev testnet, testnet, prod)


tudor-malene (approved):
looks reasonable Matt

PR #204, opened 2022-05-27
Bootstrap node updates

otherview:


go/obscuronode/host/host.go
```diff
@@ -580,7 +592,7 @@ func (a *Node) awaitSecret() {
 
 		case <-time.After(time.Second * 10):
 			// This will provide useful feedback if things are stuck (and in tests if any goroutines got stranded on this select
-			nodecommon.LogWithID(a.shortID, "Still waiting for secret from the L1...")
+			nodecommon.LogWithID(a.shortID, "Waiting for secret from the L1 after %s...", startTime)
```
BedrockSquirrel, line 595: time you mean to put `time.Since(startTime)` as the argument here?

go/obscuronode/host/host.go
```diff
@@ -34,8 +35,9 @@ type Node struct {
 	stats StatsCollector
 
 	// control the host lifecycle
-	exitNodeCh        chan bool
-	stopNodeInterrupt *int32
+	exitNodeCh          chan bool
+	stopNodeInterrupt   *int32
+	ignoreBootstrapping *int32
```
BedrockSquirrel, line 40: I'd probably call this `boostrapCompleted` or something, else stick a comment next to it to explain what the value of it means

go/obscuronode/host/host.go
```diff
@@ -538,30 +466,114 @@ func (a *Node) monitorBlocks() {
...
+		blkHeader := <-listener
+
 		// don't process blocks if the node is stopping
 		if atomic.LoadInt32(a.stopNodeInterrupt) == 1 {
 			return
 		}
-		block, err := a.ethClient.BlockByHash(latestBlkHeader.Hash())
+
+		// ignore blocks if bootstrapping is happening
+		if atomic.LoadInt32(a.ignoreBootstrapping) == 0 {
```
BedrockSquirrel, line 477: I'd be tempted to make a node method like `a.bootstrappingInProgress() bool` or something to hide the implementation details of this check away, then you wouldn't really need the comment because it's self-documenting

go/obscuronode/host/host.go
```diff
@@ -538,30 +466,114 @@ func (a *Node) monitorBlocks() {
...
+		nextBlk, err = a.ethClient.BlockByNumber(big.NewInt(currentBlock.Number().Int64() + 1))
+		if err != nil {
+			if errors.Is(err, ethereum.NotFound) {
+				break
+			}
+			panic(err)
+		}
+		currentBlock = nextBlk
+
+		if time.Since(startTime)%(10*time.Second) == 0 {
```
BedrockSquirrel, line 537: Are we sure this reliably works as intended? Looks a bit scary to me (like could it spam out multiple times during each 10th second, could it miss the 10th second and not reliably log, etc.)

For clarity I'd be tempted to do something like name the variable `lastProgressLog`, and then
``` 
if time.Since(lastProgressLog) > 10 * time.Second {
    // log the msg
    lastProgressLog = time.Now()
}
```

BedrockSquirrel (commented):
Few minor things but generally LGTM, some nice tidying up in here

go/obscuronode/host/host.go
otherview: Yes! Good catch ! 👍

go/obscuronode/host/host.go
otherview: Yeah that's a better name -  Updating comments

go/obscuronode/host/host.go
otherview: The updated var name makes it clearer I think, would avoid creating methods if they are not reused.

go/obscuronode/host/host.go
otherview: That's a good concern. This should work, but I triple checked https://go.dev/play/p/MZHrPvKCVPp :)

otherview:
Closing this PR to reduce the scope of changes.

PR #209, opened 2022-05-27
fixes block_by_number call

otherview:
Why is this needed
- Not sure how it was working before as it was comparing two pointers
- Returns the same error as ethereum


BedrockSquirrel (approved):
lgtm

PR #213, opened 2022-05-27
host file cleanups

otherview:

- In relation to https://github.com/obscuronet/obscuro-internal/issues/197
- To prepare for upcoming changes for the bootstrap as streaming

- removed single use methods
- added comments
- renamed some var
- Reduced Nesting https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting

- Hosts file

### DIFF
--- a/go/obscuronode/host/host.go
+++ b/go/obscuronode/host/host.go
@@ -6,16 +6,14 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/naoina/toml"
-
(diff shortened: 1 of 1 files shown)


BedrockSquirrel (approved):
lgtm

tudor-malene (changes requested):
I can't easily follow what's going on in this PR.
I thought I was tired yesterday.
I think the easiest way is to revert before more stuff is committed and we get confused

otherview:
Will revert 👍

PR #220, opened 2022-05-30
Create rollup genesis from the management contract block

otherview:

- From https://github.com/obscuronet/obscuro-internal/issues/492
- Adds a ContractMgmtBlkHash which signals the block that should be used to produce the genesis rollup

- Added ContractMgmtBlkHash as a config
- All tests and sims requires this field
- ethereum_mock tests will use the genesis block as the ContractMgmtBlkHash
- Allows the genesis node to create the genesis rollup from that block hash

- Host and tests


go/obscuronode/host/host.go
```diff
@@ -291,7 +291,18 @@ func (a *Node) startProcessing() {
 	a.EnclaveClient.Start(lastBlock)
 
 	if a.config.IsGenesis {
-		a.initialiseProtocol(&lastBlock)
+		// if the ContractMgmtBlkHash is not specified (like in the ethereum_mock)
+		// the genesis rollup will use the latest block
+		if a.config.ContractMgmtBlkHash.Hex() == (common.Hash{}).Hex() {
+			a.initialiseProtocol(lastBlock.Hash())
```
joeldudley, line 297: @otherview Isn't it better to make the contractmgmtblockhash mandatory, and handle this on the ethereum mock side? The Ethereum mock could check if the hash is some specific value, and return the latest block if so.

In general, IMO it's better for the "real" code to not know that mocks could be used, and take a single approach.

joeldudley:
@otherview You should have a CLI flag for optionally specifying the mgmtcontractblockhash in the host's main method.

go/obscuronode/host/host.go
```diff
@@ -291,7 +291,18 @@ func (a *Node) startProcessing() {
...
+		} else {
+			// the block should be available in the l1
+			blk, err := a.ethClient.BlockByHash(*a.config.ContractMgmtBlkHash)
+			if err != nil {
+				panic(err)
```
joeldudley, line 302: This error should indicate that it's the contract mgmt block in particular that's missing. Otherwise, it will be a generic "block missing" error.

joeldudley (changes requested):
A few changes.

go/obscuronode/host/host.go
otherview: Ya sure that makes sense 👍

go/obscuronode/host/hostrunner/cli_flags.go
```diff
@@ -50,6 +50,9 @@ const (
 	privateKeyName  = "privateKey"
 	privateKeyUsage = "The private key for the L1 node account"
 
+	contractMgmtBlkHashName  = "contractMgmtBlkHash"
+	contractMgmtBlkHashUsage = "The block has where the contract management was deployed"
```
joeldudley, line 54: Typo here - "block has"

go/obscuronode/host/hostrunner/cli.go
```diff
@@ -57,6 +58,7 @@ func ParseConfig() config.HostConfig {
 	logPath := flag.String(logPathName, defaultConfig.LogPath, logPathUsage)
 	chainID := flag.Int64(chainIDName, defaultConfig.ChainID.Int64(), chainIDUsage)
 	privateKeyStr := flag.String(privateKeyName, defaultConfig.PrivateKeyString, privateKeyUsage)
+	contractMgmtBlkHash := flag.String(contractMgmtBlkHashName, common.Hash{}.Hex(), contractMgmtBlkHashUsage)
```
joeldudley, line 61: For consistency, it would be good to use `defaultConfig.ContractMgmtBlkHash` here.

joeldudley (changes requested):
Minor changes.

go/obscuronode/host/hostrunner/cli_flags.go
otherview: 👍

go/obscuronode/host/hostrunner/cli.go
otherview: 👍

joeldudley (approved):
LGTM

otherview:

PR #222, opened 2022-05-30
Agg boot from the management contract block

otherview:

- Linked to https://github.com/obscuronet/obscuro-internal/issues/487
- Allows the node to avoid parsing AND storing ALL blocks
- Aggregator only parse/store blocks from the block where the management contract was deployed
- ~Enclave of the Genesis Node will not need to have the l1 block available to generate the genesis~
- In memory Host no longer validate blocks as that would prevent block ingestion
- This is a PR changeset for https://github.com/obscuronet/go-obscuro/tree/pedro/mgmt_contract_blk_hash

- Host now starts the bootstrap from the `a.config.ContractMgmtBlockHash` instead of genesis block
- functional PR

- Host


otherview:

PR #224, opened 2022-05-31
Creates a standalone client to deploy the management and ERC20 contracts.

joeldudley:

As part of https://github.com/obscuronet/obscuro-internal/issues/448, we need inject txs into the devnet. A precursor to this is the deployment of the mgmt contract and ERC20 contract to the L1.

This change introduces a tool that deploys these contracts

- Functional
- Introduces tool to deploy mgmt and ERC20 contracts
- Updates network deployer tooling to create these contracts

- I had to modify `DeployContract` to take the nonce as an argument (instead of generating it) and to return an error. This was necessary because the tool creates a new wallet on each run, so the nonce is out of sync, and it has to iterate through the nonces to find the correct one

### DIFF
--- a/run_logs.txt
+++ b/run_logs.txt
@@ -1,2 +1,0 @@
-Geth nodes created in: /var/folders/gm/ktn909c96ggf_yljbfg8787r0000gp/T/165399708118432270384
-Geth network started.


dockerfiles/enclave_local_build.Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM ghcr.io/edgelesssys/ego-dev:latest
 
 # build the enclave from the current branch
 RUN mkdir /home/obscuro-playground
-COPY ./ /home/obscuro-playground
+COPY . /home/obscuro-playground
```
joeldudley, line 5: This duplicates https://github.com/obscuronet/go-obscuro/pull/223, but I needed it for this PR.

integration/simulation/network/geth_network.go
```diff
@@ -158,30 +166,30 @@ func createEthClientConnection(id int64, port uint) ethclient.EthClient {
 	return ethnode
 }
 
-func deployContract(workerClient ethclient.EthClient, w wallet.Wallet, contractBytes []byte) *common.Address {
+func DeployContract(workerClient ethclient.EthClient, w wallet.Wallet, contractBytes []byte, nonce uint64) (*common.Address, error) {
```
joeldudley, line 169: Note that `DeployContract` now takes a nonce and returns an error.

This is required because the contract deployer doesn't know what nonce to use, so it has to start from zero and iterate in response to errors. This isn't an issue in the sims, because the wallet is created at the start and can keep track of the correct nonce to use.

tools/contractdeployer/README.md
```diff
@@ -0,0 +1,15 @@
+# Contract deployer
```
BedrockSquirrel, line 1: I feel like there might be a lib out there that does parts of this contract deployment stuff already (and handle things like the nonce stuff for us). But we can maybe look into that longer term if we find ourselves deploying L1 contracts more often.

joeldudley: Yeah. OOS for this PR imo.

PR #225, opened 2022-05-31
Catches nil pointer error.

joeldudley:

Sometimes the stream is nil and we error out, failing to close the Docker containers

- Functional


tudor-malene (approved):
lgtm

PR #226, opened 2022-05-31
Clean-up step in CI

joeldudley:

Lots of zombie CI processes across runs.

Functional.

Adds a clean-up step for building PRs and manual test CI.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -21,7 +21,7 @@ jobs:
         uses: docker/setup-buildx-action@v1
 
       - name: Clean-up
-        run: sudo pkill -f geth; sudo docker stop $(sudo docker ps -a -q); sudo docker rm $(sudo docker ps -a -q)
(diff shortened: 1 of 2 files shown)


.github/workflows/build-pr.yml
```diff
@@ -12,6 +12,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: Clean-up
+        run:  sudo pkill -f geth; sudo pkill -f obscuro; sudo docker stop $(sudo docker ps -a -q); sudo docker rm $(sudo docker ps -a -q)
```
BedrockSquirrel, line 16: can you make that `obscuronode` instead of `obscuro`, the github runner has obscuro in the path

BedrockSquirrel (changes requested):

.github/workflows/build-pr.yml
joeldudley: Not sure `obscuronode` will work. Let me just remove that for now, we can iterate.

BedrockSquirrel (approved):
LGTM

PR #227, opened 2022-05-31
Refactors contract deployer tool to be generic network manager tool

joeldudley:

We already have a contract deployer tool. Instead of spinning up a new tool for each management function, we should extend the existing tool, since it's easier to maintain that way.

- Refactoring
- Makes way for introduction of new commands in the future

N/A


BedrockSquirrel (approved):
LGTM

PR #234, opened 2022-06-02
Replace obscuro-playground with go-obscuro

otherview:

- No ticket, quick change that ensures all instances of `obscuro-playground` are now `go-obscuro`

- Find and Replace the name everywhere in the repo

- All Test pass
- Might be worth checking the AzureDeployer


BedrockSquirrel:
I asked about doing this the other day and @tudor-malene said he'd been waiting for a time when there were no big PRs in flight to avoid merge conflict hell. In general seems like a good change so long as the timing is right. 

FWIW the AzureDeployer changes will be fine, it's just the directory name that the repo gets checked out into on the VM but since it's always a fresh VM there's no backwards compatibility issues etc.

PR #236, opened 2022-06-03
gethnetwork tests now run in a single network

otherview:

- No ticket
- It's a small improvement on two fronts
- Reduces the duration of the tests by ~20secs
- Reduces the number of port allocations by the geth nodes

- gethnetwork package now run the tests in on network instead of spining a whole netwokr for each test
- functional PR

- gethnetwork package

### DIFF
--- a/integration/gethnetwork/geth_network_test.go
+++ b/integration/gethnetwork/geth_network_test.go
@@ -3,12 +3,13 @@ package gethnetwork
 import (
 	"errors"
 	"fmt"
-	"github.com/obscuronet/obscuro-playground/go/obscuronode/wallet"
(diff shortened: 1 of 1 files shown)


joeldudley (approved):
I thought best practice was to have a completely clean slate per test, to avoid interference between tests. And there's no risk of ports running out here, the tests are allocated 1k ports.

But happy to merge if you prefer it this way.

tudor-malene (changes requested):
Agree with @joeldudley . This can cause pain and bugs down the line, and take more time to fix

otherview:
My thinking was that none of these tests are overlapping, so they could run on the same network, and save us some time and process management. 
But I see the point you guys are making. It was a quick enough PR so no hassle.

PR #237, opened 2022-06-06
During cleanup in CI builds: kill processes on ports 30000-39999

BedrockSquirrel:

We still occasionally see failures where geth nodes fail to start because the test ports are still occupied.

- replace the kill commands from before with a port-range based command that covers the range 30000-39999 where we run our tests

- happy the command does what it's supposed to do?
- happy it can replace the commands that were there before or should we keep some/all of them?

### Evidence
I ran a build and ran the commands to convince myself they worked as intended. With no build running I saw no output for `lsof -ti:30000-39999`.

With a build running I saw:
Screenshot 2022-06-06 at 12 55 26

Then I ran `lsof -ti:30000-39999|kill` twice and the second time it output saying there was no input to the kill command (they were dead)

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -21,7 +21,7 @@ jobs:
         uses: docker/setup-buildx-action@v1
 
       - name: Clean-up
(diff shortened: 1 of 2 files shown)


tudor-malene (approved):
lgtm

PR #238, opened 2022-06-06
Allows injection of transactions into devnet Obscuro network

joeldudley:

As part of https://github.com/obscuronet/obscuro-internal/issues/448, we need the ability to injects txs into an Obscuro devnet.

Functional.

* Updates `run.sh` script that runs devnet to give nodes P2P addresses (required for injecting and broadcasting txs)
* Some clean-up in transaction_injector.go:
  * Timeout if hangs for too long trying to get nonce to inject transaction
  * Handles edge cases when trying to select random client/wallet when there is only one

Key changes are in the `run.sh` script.

### DIFF
--- a/integration/simulation/transaction_injector.go
+++ b/integration/simulation/transaction_injector.go
@@ -359,7 +359,7 @@ func NextNonce(cl *obscuroclient.Client, w wallet.Wallet) uint64 {
 		}
 		counter++
 
-		if counter > 10000 {
+		if counter > timeoutMillis {
 			panic("Transaction injector failed to retrieve nonce after ten seconds...")
 		}
 
--- a/integration/simulation/transaction_injector.go
+++ b/integration/simulation/transaction_injector.go
@@ -26,7 +26,7 @@ import (
 	stats2 "github.com/obscuronet/obscuro-playground/integration/simulation/stats"
 )
 
-const maxRetries = 13 // The number of times to retry waiting for an updated nonce for a wallet.
+const timeoutMillis = 10000 // The timeout in millis to wait for an updated nonce for a wallet.
 
 // TransactionInjector is a structure that generates, issues and tracks transactions
 type TransactionInjector struct {
@@ -349,7 +349,7 @@ func readNonce(cl *obscuroclient.Client, a common.Address) uint64 {
 }
 
 func NextNonce(cl *obscuroclient.Client, w wallet.Wallet) uint64 {
-	retries := 0
+	counter := 0
 
 	// only returns the nonce when the previous transaction was recorded
 	for {
@@ -357,11 +357,12 @@ func NextNonce(cl *obscuroclient.Client, w wallet.Wallet) uint64 {
 		if result == w.GetNonce() {
 			return w.GetNonceAndIncrement()
 		}
-		time.Sleep(time.Duration(2^retries) * time.Millisecond) // For 13 retries, this is around 17 seconds.
-		retries++
+		counter++
 
-		if retries > maxRetries {
+		if counter > 10000 {
 			panic("Transaction injector failed to retrieve nonce after ten seconds...")
 		}
+
+		time.Sleep(time.Millisecond)
 	}
 }
--- a/integration/simulation/transaction_injector.go
+++ b/integration/simulation/transaction_injector.go
@@ -26,6 +26,8 @@ import (
 	stats2 "github.com/obscuronet/obscuro-playground/integration/simulation/stats"
 )
 
+const timeoutMillis = 10000 // The timeout in millis to wait for an updated nonce for a wallet.
+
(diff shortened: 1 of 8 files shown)


integration/simulation/transaction_injector.go
```diff
@@ -296,18 +298,30 @@ func (m *TransactionInjector) createInvalidSignage(tx types.TxData, w wallet.Wal
 }
 
 func (m *TransactionInjector) rndObsWallet() wallet.Wallet {
+	if len(m.obsWallets) == 1 {
```
joeldudley, line 301: `rand.Intn` spews if the len is zero, here and below.

integration/simulation/transaction_injector.go
```diff
@@ -347,12 +361,19 @@ func readNonce(cl *obscuroclient.Client, a common.Address) uint64 {
...
 	for {
 		result := readNonce(cl, w.Address())
 		if result == w.GetNonce() {
 			return w.GetNonceAndIncrement()
 		}
 		time.Sleep(time.Millisecond)
+		counter++
+
+		if counter > timeoutMillis {
+			panic("Transaction injector failed to retrieve nonce after ten seconds...")
```
joeldudley, line 376: This used to spin forever if for some reason the nonce wasn't increasing (e.g. the enclave was down).

integration/simulation/transaction_injector.go
BedrockSquirrel: I think the `rand.Intn()` usage isn't quite right here, docs says it gives an int in the half-open interval `[0, n)` so it's safe to use it like `return m.obsWallets[rand.Intn(len(m.obsWallets))]` and with the current way we'd never get the last element in the slice I think. That would let you get rid of the length 1 handling.

integration/simulation/transaction_injector.go
BedrockSquirrel: Hitting it a thousand times a second feels a bit intense for 10 seconds. If the server is struggling under load then this is the sort of thing that causes a death spiral.

If it normally takes a couple of millis you could back off, like `time.sleep(2^retries * time.Millisecond)` and then `if (retries > 10) { panic(...) }`

tools/azuredeployer/networkdeployer/run.sh
```diff
@@ -7,17 +7,20 @@ rm -f ./run_logs.txt
...
-sudo docker run -e OE_SIMULATION=0 --privileged -v /dev/sgx:/dev/sgx -p 127.0.0.1:11001:11000/tcp obscuro_enclave --hostID 2 --address :11000 > ./run_logs.txt 2>&1 &
-go-obscuro/go/obscuronode/host/main/host --id=1 --isGenesis=true --enclaveRPCAddress=localhost:11000 --clientRPCAddress=0.0.0.0:13000 --l1NodePort=12100 --privateKey=$PRIV_KEY_ONE > ./run_logs.txt 2>&1 &
-go-obscuro/go/obscuronode/host/main/host --id=2 --isGenesis=false --enclaveRPCAddress=localhost:11001 --clientRPCAddress=localhost:13001 --l1NodePort=12101 --privateKey=$PRIV_KEY_TWO > ./run_logs.txt 2>&1 &
+sudo docker run -e OE_SIMULATION=1 --privileged -v /dev/sgx:/dev/sgx -p 127.0.0.1:11000:11000/tcp obscuro_enclave --hostID 1 --address :11000 > ./run_logs.txt 2>&1 &
+sudo docker run -e OE_SIMULATION=1 --privileged -v /dev/sgx:/dev/sgx -p 127.0.0.1:11001:11000/tcp obscuro_enclave --hostID 2 --address :11000 > ./run_logs.txt 2>&1 &
+go-obscuro/go/obscuronode/host/main/host --id=1 --isGenesis=true --p2pAddress=localhost:10000 --peerP2PAddresses=localhost:10001 --enclaveRPCAddress=localhost:11000 --clientRPCAddress=0.0.0.0:13000 --l1NodePort=12100 --privateKey=$PRIV_KEY_ONE > ./run_logs.txt 2>&1 &
```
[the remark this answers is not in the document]
BedrockSquirrel: Worth adding a todo to remind us to remove peer addresses when they discover the peers themselves or think we'll clean it up naturally?

BedrockSquirrel (commented):
Looks good, just a few minor thoughts

tools/azuredeployer/networkdeployer/run.sh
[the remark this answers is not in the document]
joeldudley: Think it'll be fine. This will change a lot.

BedrockSquirrel (approved):
LGTM once the lint/build stuff passing. We can rework the nonce stuff in future.

PR #239, opened 2022-06-06
Reinstate docker cleanup, ensure sudo commands

BedrockSquirrel:

- Concerns that docker containers might not respond well to this shutdown so we reinstate the docker clean-up
- Need sudo on the kill command and --verbose flag seems to be upsetting things

- update the clean-up command with above changes

- happy with the clean-up changes? I was able to reproduce the error on the CI box and this seemed to fix it


joeldudley (approved):
LGTM

PR #244, opened 2022-06-07
Bumps timeout for retrieving nonce.

joeldudley:

The timeout kept getting hit. It's not clear to me whether this is a genuine timeout or a failure. But increasing it should reduce flaky failures.

Also panics if remote nonce exceeds local nonce, which should never happen.

- Functional

N/A


BedrockSquirrel (approved):
LGTM - but I'm worried if that massive timeout helps...

PR #245, opened 2022-06-07
Increases threshold of txs to consider simulation failed

joeldudley:

Previously, the sim could deadlock after one or two transactions, and still be considered a success. We want to catch that edge-case.

- Functional

N/A


BedrockSquirrel (approved):
lgtm

PR #246, opened 2022-06-07
Error message from EVM facade that makes reading CI output easier.

joeldudley:

The EVM facade logs out harmless tx failed messages. However, the wording of these messages makes it harder to search the CI output for "true" failures.

Refactoring.

N/A


tudor-malene (approved):
lgtm, (after linter is happy)

PR #256, opened 2022-06-08
Allow Metamask to speak to Obscuro node via wallet extension

joeldudley:

Allows MetaMask to connect to an Obscuro node via the wallet extension

Functional.

- Exposes the existing RPC APIs via websockets, as required by the wallet extension
- Adds additional ethereum RPC APIs that are polled by Metamask
- Adds new CLI flags to control how the websockets layer is served


BedrockSquirrel (approved):
LGTM

PR #257, opened 2022-06-08
fix enclave deploy IP retrieval

BedrockSquirrel:

- when you run the enclave azure deployer at the moment it fails to get the IP and so fails to run the script that sets up docker and docker image

- tweak enclave deploy config to match the network deploy config so the script they both use works for enclave deploy
- also updated an out of date part of the readme

- the config change

### Evidence
Failure before the change:
Screenshot 2022-06-08 at 10 30 58

Success after the change:
Screenshot 2022-06-08 at 10 40 53


joeldudley (approved):
LGTM

PR #260, opened 2022-06-08
Enables websockets for simulation nodes.

joeldudley:

Simulation socket nodes don't currently expose an RPC websocket port. This will block the integration testing of the wallet extension.

Functional.

Enables websocket ports for socket simulation nodes.


BedrockSquirrel (approved):
LGTM

PR #261, opened 2022-06-08
Renables and fixes first wallet extension test.

joeldudley:

The wallet extension tests are disabled. This renables the first wallet extension test and gets it to pass.

As a trade-off, I've had to disable encryption of requests to the enclave for now. They will be reimplemented later.

Functional.

Reimplements first wallet extension test.


BedrockSquirrel (approved):
LGTM, keep us honest til the tests are back

PR #264, opened 2022-06-09
Allows viewing keys to be submitted to the enclave.

joeldudley:

Previously, the wallet extension stored the viewing keys locally. For the testnet, we need them stored inside the enclave.

- Added a host RPC method to submit viewing keys to the host
- Added an enclave RPC method to submit viewing keys to the enclave
- Added validation and storage of the viewing keys inside the enclave


go/obscuronode/enclave/enclave.go
```diff
@@ -70,6 +77,70 @@ type enclaveImpl struct {
 	blockProcessingMutex sync.Mutex
 }
 
+// NewEnclave creates a new enclave.
+// `genesisJSON` is the configuration for the corresponding L1's genesis block. This is used to validate the blocks
+// received from the L1 node if `validateBlocks` is set to true.
+func NewEnclave(
```
joeldudley, line 83: If you'll forgive some minor refactoring, I moved the `NewEnclave` method to the top of the file, for consistency with other files. Mea maxima culpa.

go/obscuronode/enclave/enclave.go
```diff
@@ -544,6 +615,29 @@ func (e *enclaveImpl) ShareSecret(att *obscurocommon.AttestationReport) (obscuro
 	return e.encryptSecret(att.PubKey, secret)
 }
 
+func (e *enclaveImpl) AddViewingKey(viewingKeyBytes []byte, signature []byte) error {
```
joeldudley, line 618: Most of this method is copied over from the old (now-deleted) Obscuro facade. See https://github.com/obscuronet/go-obscuro/blob/58751aa7bde8397238c01b52a94430a25114e6d1/tools/walletextension/obscuro_facade.go#L54.

go/obscuronode/enclave/enclave.go
```diff
@@ -656,71 +750,8 @@ type processingEnvironment struct {
...
-		privateKey:            privKey,
-		publicKeySerialized:   serializedPubKey,
-		transactionBlobCrypto: obscurocore.NewTransactionBlobCryptoImpl(),
-	}
-}
-
-// EncryptWithPublicKey encrypts data with public key
-func EncryptWithPublicKey(msg []byte, pub *rsa.PublicKey) ([]byte, error) {
+// encryptWithPublicKey encrypts data with public key
+func encryptWithPublicKey(msg []byte, pub *rsa.PublicKey) ([]byte, error) {
```
joeldudley, line 754: Some more minor refactoring. I made these two methods private to the package.

integration/walletextension/wallet_extension_test.go
```diff
@@ -36,10 +38,8 @@ const (
...
 	dummyAccountAddr = "0x8D97689C9818892B700e27F316cc3E41e17fBeb9"
 
-	startPort              = 3000
-	httpProtocol           = "http://"
-	pathGenerateViewingKey = "/generateviewingkey/"
```
joeldudley, line 41: Some more clean-up. I used the existing constants defined in `tools/walletextension`, rather than redefining them here.

tools/walletextension/wallet_extension.go
```diff
@@ -89,6 +91,9 @@ func (we *WalletExtension) Shutdown() {
 	}
 }
 
+// Used to check whether the server is ready.
+func (we *WalletExtension) handleReady(http.ResponseWriter, *http.Request) {}
```
joeldudley, line 95: I introduced this new method to check readiness. It used to hit the regular `/` endpoint to check for liveness, but that made the wallet extension actually try and process the body, and led to harmless but misleading errors.

tools/walletextension/wallet_extension.go
```diff
@@ -214,12 +223,6 @@ func logAndSendErr(resp http.ResponseWriter, msg string) {
 	http.Error(resp, msg, httpCodeErr)
 }
 
-// ViewingKey is the packet of data sent to the enclave when storing a new viewing key.
-type ViewingKey struct {
```
joeldudley, line 218: This structure was only relevant when shoving the viewing key down a channel to Obscuro facade.

go/obscuronode/enclave/enclave.go
BedrockSquirrel: Looking at this method and trying to figure out what if anything has changed I have a new appreciation for that rule haha

tools/walletextension/wallet_extension.go
```diff
@@ -118,9 +123,9 @@ func (we *WalletExtension) handleHTTPEthJSON(resp http.ResponseWriter, req *http
 	//}
 
 	// We forward the request on to the Obscuro node.
-	nodeResp, err := forwardMsgOverWebsocket(websocketProtocol+we.nodeAddr, body)
+	nodeResp, err := forwardMsgOverWebsocket(websocketProtocol+we.hostAddr, body)
```
BedrockSquirrel, line 126: Should this be handled by the obscuro client? Feels odd to have a client and then need to know the address and communicate with it separately.

BedrockSquirrel (approved):
LGTM, comments don't need to be addressed as part of this PR

tools/walletextension/wallet_extension.go
joeldudley: Yeah, you're right, maybe in the future.

PR #266, opened 2022-06-09
Simplify DB setup in enclave

BedrockSquirrel:
Small PR to make the edgeless DB connection PR simpler. 

- Edgeless DB will need various config (like the host address of the edgeless DB) and this abstraction I tried to guess at when I first wrote this is making that harder and not adding value.

- removes the db connector abstraction that was unhelpful and allows ethdb construction with the whole config object.

- Enclave DB connection setup

### DIFF
--- a/go/obscuronode/enclave/enclave.go
+++ b/go/obscuronode/enclave/enclave.go
@@ -90,8 +90,7 @@ func NewEnclave(
 ) nodecommon.Enclave {
 	nodeShortID := obscurocommon.ShortAddress(config.HostID)
 
-	connect := getDBConnector(config)
-	backingDB, err := connect(nodeShortID)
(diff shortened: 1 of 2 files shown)


go/obscuronode/enclave/enclave.go
```diff
@@ -773,20 +772,22 @@ func decryptWithPrivateKey(ciphertext []byte, priv *rsa.PrivateKey) ([]byte, err
 }
 
 // getDBConnector creates an appropriate EthDBConnector function based on your config
```
otherview, line 774: Might want to update the comment

otherview (approved):
Looks good - would update that comment.

tudor-malene (approved):
lgtm

PR #268, opened 2022-06-09
Encrypt eth_getBalance responses using viewing key

joeldudley:

This is a privacy feature we want on Testnet.

Functional.

- Introduces an API to get the balance from the enclave
- Implements this API to return the balance encrypted with the corresponding viewing key, if one exists
- Fixes up all the wallet tests


tools/walletextension/wallet_extension.go
```diff
@@ -129,37 +132,43 @@ func (we *WalletExtension) handleHTTPEthJSON(resp http.ResponseWriter, req *http
 		return
 	}
 
-	// This is just a temporary unmarshalling. We need to unmarshall once to check if we got an error response, then
```
joeldudley, line 132: In the new implementation, we encrypt the balance specifically, rather than the entire JSON object. This means we can improve on the old approach, where we had to tentatively unmarshal the JSON, and abort if the entire object was encrypted.

integration/walletextension/wallet_extension_test.go
```diff
@@ -32,99 +31,82 @@ import (
...
+	errInsecure          = "enclave could not respond securely to %s request because it does not have a viewing key for account"
+
+	networkStartPort = integration.StartPortWalletExtensionTest + 1
+	nodeRPCHTTPPort  = integration.StartPortWalletExtensionTest + 1 + network.DefaultHostRPCHTTPOffset
+	nodeRPCWSPort    = integration.StartPortWalletExtensionTest + 1 + network.DefaultHostRPCWSOffset
+)
+
+var (
+	walletExtensionAddr   = fmt.Sprintf("%s:%d", network.Localhost, integration.StartPortWalletExtensionTest)
+	walletExtensionConfig = walletextension.Config{
```
joeldudley, line 55: I cleaned up the tests by defining the various ports and config object just once. This makes the set up in each test much easier to follow.

tudor-malene (commented):
lgtm ( the enclave bits)

PR #270, opened 2022-06-10
Uses EncryptedResponse for GetBalance. Shifts ExecContract to Eth standard.

joeldudley:

Allows calls to be made against Obscuro nodes (via the wallet extension) without app-side changes.

- Refactoring?
- Replaces `ObscuroAPI.ExecContract` with `EthereumAPI.Call`, the standard Eth API
- Changes GetBalance to use `EncryptedResponse` type


tudor-malene (approved):
lgtm

PR #274, opened 2022-06-13
Splits viewing key management into dedicated type.

joeldudley:

This is a refactor. Previously, the logic for viewing keys was mixed in with the enclave. This logic will extend in the future, so best to split it out now.

Refactoring.


BedrockSquirrel (approved):
LGTM

PR #275, opened 2022-06-13
Use ECDSA for enclave private key. Use hardcoded key.

joeldudley:

Previously, enclave's private key was RSA. Moved to ECDSA, and hardcoded the key (so that the wallet extension know what key to encrypt requests with for now).

Functional.


BedrockSquirrel (commented):
LGTM if the network tests are passing (i.e. the secret communication over the L1 contract was unaffected)

BedrockSquirrel (approved):
Forgot to hit approve

PR #276, opened 2022-06-13
Renables encryption of viewing keys on wallet extension side.

joeldudley:

To prevent MitM attacks, viewing keys should be encrypted before they are sent to the enclave.

Functional.

- Encrypts the viewing key on the wallet side
- Decrypts the viewing key on the enclave side


BedrockSquirrel (approved):
LGTM

PR #280, opened 2022-06-17
Moves wallet extension tests to own package.

joeldudley:

The wallet extension integration tests were in the same folder as the sim tests due to a DB file overwriting issue. This is now fixed.

Refactoring.

- Moved file.


BedrockSquirrel (approved):
lgtm

PR #282, opened 2022-06-17
Extend viewing key manager to handle decryption of sensitive RPC requests

joeldudley:

Encapsulates the decryption of sensitive RPC requests to reduce the amount of logic in enclave.go.

Refactoring.


BedrockSquirrel (approved):
LGTM

PR #288, opened 2022-06-20
Avoids NPE in case contract fails to deploy.

joeldudley:

An NPE can occur if the contract fails to deploy, as there is never a receipt to check the status of.

Functional.

- Avoids NPE


BedrockSquirrel (approved):
LGTM

otherview (approved):

PR #289, opened 2022-06-20
Fork detection structure for rollups

otherview:

- [Ticket](https://github.com/obscuronet/obscuro-internal/issues/524)
- Detects if there has been a continued fork and marks the contract / locks withdrawals

- Management contract now has a tree structure that captures child/parent connection between rollups
- Each appended rollup will check for forks and disables forks if found
- This is a first working iteration, will need more work such as deploying the tree structure as a library and expanding the test suite.
- Again, this is prevention in the case the TEE is hacked. Such that a wrongful withdrawal is never executed.
- functional PR

- mgmt contract
- smart contract tests


CaisManai (approved):
LGTM

PR #294, opened 2022-06-21
Testnet: Manual Github Action to deploy a geth network

otherview:

- [Ticket](https://github.com/obscuronet/obscuro-internal/issues/543)
- Manually deploys an l1 network to the testnet machines in azure ( First part )
- This change will add a manual action that allows to trigger the functionality

- Adds a docker file that can run a geth network
- Adds a Github manual action that can start a network
- Pushes the image to dockerhub 
- Functional

- Github action files
- Dockerfile on the testnet folder ( the folder might be removed if it has few files)


testnet/gethnetwork.Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+RUN rm -rf /home/go-obscuro/integration/.build/geth_bin/ && rm -rf /home/go-obscuro/integration/gethnetwork/geth_bin/
+RUN ./build_geth_binary.sh --version=v1.10.17
+
+# build the gethnetwork exec
+WORKDIR /home/go-obscuro/integration/gethnetwork/main
+RUN go build
+
+# expose the http and the ws ports to the host
+EXPOSE 8025 8026 8027 9000 9001 9002
+ENTRYPOINT ["/home/go-obscuro/integration/gethnetwork/main/main", "--numNodes=3", "--startPort=8000","--websocketStartPort=9000"]
```
joeldudley, line 30: @otherview And a comment here explaining why we start on port 8k, but expose the seemingly random set of ports above.

testnet/gethnetwork.Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM golang:1.17-alpine
```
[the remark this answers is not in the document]
otherview: Good point, I'll add the reference to the github action, where the documentation "lives".

There's more stuff like the ports that is hardcoded and withouth that pointer it seems like random yeah.

testnet/gethnetwork.Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+WORKDIR /home/go-obscuro
+RUN go get -d -v ./...
+
+# make sure the geth network code is available
+COPY . /home/go-obscuro
+
+# reset any previous geth build
+WORKDIR /home/go-obscuro/integration/gethnetwork/
+RUN rm -rf /home/go-obscuro/integration/.build/geth_bin/ && rm -rf /home/go-obscuro/integration/gethnetwork/geth_bin/
+RUN ./build_geth_binary.sh --version=v1.10.17
```
joeldudley, line 22: @otherview Can't we reuse the prebuilt binary in the repo? Would be much quicker.

joeldudley (changes requested):
Just some documentation comments.

testnet/gethnetwork.Dockerfile
otherview: No, I don't seem to be able to reuse it. I think it was built with ubuntu maybe ?

Also, there's something funky going on because we have the `geth_bin/geth-v1.10.17` and the `.build../geth-v1.10.17` 
I didn't look much into it, but something is off with the scripts.

joeldudley (approved):
Ok lgtm

PR #296, opened 2022-06-21
Testnet: Manual Github Action to deploy a geth network ( Second Part )

otherview:

- [Ticket](https://github.com/obscuronet/obscuro-internal/issues/543)
- Manually deploys an l1 network to the testnet machines in azure ( First part )
- This change will add a manual action that allows to trigger the functionality

- Adds a docker file that can run a geth network
- Adds a Github manual action that can start a network
- Pushes the image to dockerhub 
- Functional

- Github action files
- Dockerfile on the testnet folder ( the folder might be removed if it has few files)


joeldudley (approved):
LG

PR #298, opened 2022-06-22
Enclave SQL: Fix binary length bug and column naming

BedrockSquirrel:

- there is a bug with the 32-byte fixed length binary column that we weren't hitting. Some of the keys that geth uses are 41-bytes long and sqlite just truncates the key on insert and select, the truncated hashes always happened to match in testing

- MariaDB (underlying Edgeless DB (edb)) doesn't allow `key` to be used as a column name because it's a keyword

- renamed SQL columns/table to avoid the protected keyword
- Changed the key column to be varbinary of up to 64-bytes to fix the bug and avoid having to pad the 32-byte keys 

- happy with the col length fix?
- any suggestions to improve the table/col names?

### DIFF
--- a/go/obscuronode/enclave/sql/sql.go
+++ b/go/obscuronode/enclave/sql/sql.go
@@ -9,10 +9,10 @@ import (
 )
 
 const (
-	getQry    = `select kv.value from kv where kv.key = ?;`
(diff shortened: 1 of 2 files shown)


tudor-malene (approved):
lgtm

PR #299, opened 2022-06-22
Store aggregator host addresses in management contract

joeldudley:

We need to store host addresses in the management contract to avoid the need to hardcode addresses and allow the network to grow organically.

Functional.

- Store host addresses in management contract and allow them to be retrieved
- Extend mgmt contract lib to pack and unpack eth_call requests

### FYI

If you need to test this, you can stick the following code somewhere in the sim:

```
        println("jjjj making call")
	msg, err := simulation.Params.MgmtContractLib.GetHostAddresses()
	if err != nil {
		panic(err)
	}
	response, err := simulation.EthClients[0].EthClient().CallContract(context.Background(), msg, nil)
	if err != nil {
		panic(err)
	}
	decodedResponse, err := simulation.Params.MgmtContractLib.DecodeCallResponse(response)
	if err != nil {
		panic(err)
	}
	for _, entry := range decodedResponse {
		println(fmt.Sprintf("jjj call result: %s", entry.([]string)))
	}
```


contracts/management_contract.sol
```diff
@@ -1,15 +1,14 @@
...
 pragma solidity >=0.7.0 <0.9.0;
 
 contract ManagementContract {
-
-    // TODO these should all be private
+    // TODO these should all be private, but are public for testing in integration/smartcontract
     mapping(uint256 => Rollup[]) public rollups;
     mapping(address => string) public attestationRequests;
     mapping(address => bool) public attested;
+    string[] public hostAddresses; // The addresses of all the Obscuro hosts on the network.
```
otherview, line 11: It can be fetched using GetHostAddresses - it's fine to be private right ?

integration/simulation/network/obscuro_node_utils.go
```diff
@@ -130,8 +130,10 @@ func startRemoteEnclaveServers(startAt int, params *params.SimParams, stats *sta
 	for i := startAt; i < params.NumberOfNodes; i++ {
 		// create a remote enclave server
 		enclaveAddr := fmt.Sprintf("%s:%d", Localhost, params.StartPort+DefaultEnclaveOffset+i)
+		hostAddr := fmt.Sprintf("%s:%d", Localhost, params.StartPort+DefaultHostP2pOffset+i)
```
otherview, line 133: Why are these changes needed ? Sorry I didn't get it.

otherview (changes requested):
Overall looks good!

The bug description is weird, can we write a unit test leveraging the `smartcontracts_test.go` to make sure the data is being writen as expected ?

integration/simulation/network/obscuro_node_utils.go
otherview: Wait, is it so the enclave generates the attestment ?

go/ethclient/mgmtcontractlib/mgmt_contract_lib.go
```diff
@@ -183,6 +190,22 @@ func (c *contractLibImpl) CreateInitializeSecret(tx *obscurocommon.L1InitializeS
...
 
+func (c *contractLibImpl) GetHostAddresses() (ethereum.CallMsg, error) {
+	data, err := c.contractABI.Pack(GetHostAddressesMethod)
+	if err != nil {
+		return ethereum.CallMsg{}, fmt.Errorf("could not pack the call data. Cause: %w", err)
+	}
+	return ethereum.CallMsg{To: c.addr, Data: data}, nil
+}
+
+func (c *contractLibImpl) DecodeCallResponse(callResponse []byte) ([]interface{}, error) {
```
otherview, line 201: Can this return a []string ?

joeldudley: Not sure yet, haven't used it in a wide enough set of scenarios. I'm tempted to leave as interface for now, and we can narrow it down later. Unless you're sure it's always []string?

joeldudley:
@otherview Re: the test to catch the bug, let's discuss that on the bug PR: https://github.com/obscuronet/go-obscuro/pull/300/files. Don't want to put the fix in this PR, it's a bit unrelated.

integration/simulation/simulation_tester.go
```diff
@@ -62,6 +63,23 @@ func testSimulation(t *testing.T, netw network.Network, params *params.SimParams
 	// execute the simulation
 	simulation.Start()
 
+	println("jjjj making call")
```
BedrockSquirrel, line 66: Did you mean to leave these 'jjj's in? (Other one in the loop below)

joeldudley: Nope, removed, sorry

go/ethclient/mgmtcontractlib/mgmt_contract_lib.go
joeldudley: Fixed

otherview (approved):

PR #301, opened 2022-06-22
Setup /data folder for enclave secure persistence

BedrockSquirrel:

- The enclave needs to be able to persist certificates for connection to Edgeless DB
- Anything the enclave persists should be sealed using its built-in private keys
- We may wish to persist other things here going forwards to make enclave restart smoother (e.g. the sealed and encrypted secret)

- change dockerfiles to create directories `/home/obscuro/go-obscuro` for the code and `/home/obscuro/data` for the data
- change enclave.json to add a file mount, make `/home/obscuro/data` available inside the enclave as `/data` (as per https://docs.edgeless.systems/ego/#/reference/config?id=mounts)

- The config change
- The dockerfile changes:
-- I have tested the enclave_local_build on the SGX server as part of my edb branch testing
-- I'll watch the build for this PR closely, it will build and test the enclave.Dockerfile
-- I haven't tested the plain_go dockerfile, I think it's probably just used ad-hoc for specific issue testing?

### DIFF
--- a/dockerfiles/enclave.Dockerfile
+++ b/dockerfiles/enclave.Dockerfile
@@ -1,8 +1,14 @@
 FROM ghcr.io/edgelesssys/ego-dev:latest
 
-RUN git clone https://github.com/obscuronet/obscuro-playground
-RUN cd obscuro-playground/go/obscuronode/enclave/main && ego-go build && ego sign main
+# on the container:
+#   /home/obscuro/data       contains working files for the enclave
+#   /home/obscuro/go-obscuro contains the src
+RUN mkdir /home/obscuro
+RUN mkdir /home/obscuro/data
+WORKDIR /home/obscuro
+
+RUN git clone https://github.com/obscuronet/go-obscuro
+RUN cd go-obscuro/go/obscuronode/enclave/main && ego-go build && ego sign main
 
 ENV OE_SIMULATION=1
-ENTRYPOINT ["ego", "run", "obscuro-playground/go/obscuronode/enclave/main/main"]
-EXPOSE 11000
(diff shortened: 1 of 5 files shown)


dockerfiles/plain_go_enclave_local.Dockerfile
```diff
@@ -1,9 +1,18 @@
 FROM golang:1.17-alpine
 
+# on the container:
+#   /data                    contains working files for the enclave
+#   /home/obscuro/go-obscuro contains the src
+#
+# Note: on ego it maps /home/obscuro/data to /data inside the enclave,
```
otherview, line 7: This is the mapping of the virtual fs yeah ?
Could that be (more) explicit  ?

dockerfiles/enclave.Dockerfile
```diff
@@ -1,8 +1,14 @@
 FROM ghcr.io/edgelesssys/ego-dev:latest
```
otherview, line 1: Do we still need a docker file that builds from main regardless of the branch that it runs on ?

otherview (approved):
Just a couple of q's, nothing that blocks merging

dockerfiles/enclave.Dockerfile
BedrockSquirrel: Yeah I don't think we do, only usage at the moment is in the azure enclave deployers I think. We should migrate that to do its own checkout and use local build I think. But think we should wait until we see what your orchestration work comes out looking like since that will probably impact.

dockerfiles/plain_go_enclave_local.Dockerfile
BedrockSquirrel: Yeah exactly, I've updated to try and make the wording more clear, it's a bit of a fudge tbh, maybe we make the enclave data dir configurable eventually.

PR #303, opened 2022-06-23
Retrieves network peers from L1

joeldudley:

The list of network peers is now stored in the L1 contract, but our P2P layer was still using the hardcoded list.

Functional.

- Monitors for incoming respond-secret-request L1 transactions
- Updates the P2P layer's peer addresses based on the contents of the L1 contract
- Deletes the peer-addresses config option
- Deletes the peer-addresses variables that are passed around in the sims and are no longer used

### DIFF
--- a/go/obscuronode/host/host.go
+++ b/go/obscuronode/host/host.go
@@ -543,27 +543,30 @@ func (a *Node) processSharedSecretResponse(_ *obscurocommon.L1RespondSecretTx) e
 		return err
 	}
 	hostAddresses := decodedResponse[0]
-	for _, entry := range decodedResponse {
-		a.config.AllP2PAddresses = entry
-	}
 
-	// We filter out any duplicate host addresses.
-	var noDupsHostAddresses []string
-	for _, newHostAddress := range hostAddresses {
-		found := false
-		for _, existingHostAddress := range noDupsHostAddresses {
-			if newHostAddress == existingHostAddress {
-				found = true
+	// We filter down the list of retrieved addresses.
+	var filteredHostAddresses []string //nolint:prealloc
+	for _, hostAddress := range hostAddresses {
+		// We exclude our own address.
+		if hostAddress == a.config.P2PAddress {
+			continue
+		}
+
+		// We exclude any duplicate host addresses.
+		isDup := false
+		for _, existingHostAddress := range filteredHostAddresses {
+			if hostAddress == existingHostAddress {
+				isDup = true
 			}
 		}
-		if !found {
-			noDupsHostAddresses = append(noDupsHostAddresses, newHostAddress)
+		if isDup {
+			continue
 		}
+
+		filteredHostAddresses = append(filteredHostAddresses, hostAddress)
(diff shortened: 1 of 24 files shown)


go/obscuronode/host/host.go
```diff
@@ -470,43 +488,86 @@ func (a *Node) handleStoreSecretTx(t *obscurocommon.L1RespondSecretTx) bool {
 	return true
 }
 
-func (a *Node) checkForSharedSecretRequests(block obscurocommon.EncodedBlock) {
```
joeldudley, line 473: GitHub made a real pig's ear of this, but the logic in `checkForSharedSecretRequests` was moved more or less intact to `processSharedSecretRequest`.

go/obscuronode/host/host.go
```diff
@@ -470,43 +488,86 @@ func (a *Node) handleStoreSecretTx(t *obscurocommon.L1RespondSecretTx) bool {
...
+		// We exclude our own address.
+		if hostAddress == a.config.P2PAddress {
+			continue
+		}
+
+		// We exclude any duplicate host addresses.
+		isDup := false
+		for _, existingHostAddress := range filteredHostAddresses {
+			if hostAddress == existingHostAddress {
+				isDup = true
```
BedrockSquirrel, line 559: Can't believe some of the stuff Go doesn't give you out of the box, super gross haha. You can `break` here I think

go/obscuronode/host/host.go
```diff
@@ -470,43 +488,86 @@ func (a *Node) handleStoreSecretTx(t *obscurocommon.L1RespondSecretTx) bool {
...
+	if err != nil {
+		return err
+	}
+	decodedResponse, err := a.mgmtContractLib.DecodeCallResponse(response)
+	if err != nil {
+		return err
+	}
+	hostAddresses := decodedResponse[0]
+
+	// We filter down the list of retrieved addresses.
```
BedrockSquirrel, line 547: Actually, could we just hide this away in a convenience method. One day to be replaced with some utility from a library or something I guess.

Something like `filteredHostAddresses := getUniqueStrings(hostAddresses)` just so this method isn't cluttered up with implementation details.

BedrockSquirrel (commented):
Code changes generally look fine, just need to convince myself the bootstrap phase is all fine (no chicken and egg race conditions).

go/obscuronode/host/host.go
joeldudley: Nice.

go/obscuronode/host/host.go
joeldudley: The convenience method would be ugly, since we also check that the host address isn't our own.

If we had a `getUniqueStrings` helper, we'd still need to loop again to remove our own address.

So I'm tempted to leave it for now, if ok?

BedrockSquirrel (approved):
Seems fine to me - hopefully we can come up with a cleaner mechanism when we upgrade all the peer-to-peer stuff

PR #320, opened 2022-06-28
Tx injector tool - Set nonces correctly, catches exit, performs quick check of tx injection success

joeldudley:

Rerunning the tx injector tool against an existing network causes a nonce issue.

We also need the tx injector tool to perform smoke testing of its success

Functional.

- Sets wallet nonces correctly
- Catches exit from tool loop
- Basic smoke test of tx injector activity


tools/networkmanager/inject_txs.go
```diff
@@ -19,19 +21,49 @@ import (
...
+		stats.NewStats(1),
+		[]ethclient.EthClient{l1Client},
+		createWallets(cfg, l1Client, l2Client),
+		&cfg.mgmtContractAddress,
+		[]obscuroclient.Client{l2Client},
+		mgmtcontractlib.NewMgmtContractLib(&cfg.mgmtContractAddress),
+		erc20contractlib.NewERC20ContractLib(&cfg.mgmtContractAddress, &cfg.erc20ContractAddress),
+	)
+
+	// We listen for interrupts, to log statistics before exiting.
```
BedrockSquirrel, line 47: Maybe this method could do with a high level comment somewhere that says what it does. I'm confused, it looks like it's stopping injecting transactions before it starts injecting them...

BedrockSquirrel (commented):
Some minor confusion but generally seems fine

tools/networkmanager/inject_txs.go
BedrockSquirrel: Oh I'm being dense, this just listens for the interrupt, doesn't immediately stop.

BedrockSquirrel (approved):
LGTM

PR #322, opened 2022-06-28
Separates out L1 and L2 chain IDs in host config. Fixes bug where wrong chain ID was returned.

joeldudley:

Host was returning L1 chain ID instead of L2 chain ID via RPC.

Functional.

- Host config now stores L2 chain ID as well
- Clarified config entry names


go/obscuronode/config/host_config.go
```diff
@@ -51,7 +50,9 @@ type HostConfig struct {
 	// The stringified private key for the host's L1 wallet
 	PrivateKeyString string
 	// The ID of the L1 chain
-	ChainID big.Int
+	L1ChainID int64
```
joeldudley, line 53: Using int64, for consistency with enclave config.

tools/networkmanager/cli.go
```diff
@@ -60,8 +59,8 @@ type Config struct {
 	l1NodeWebsocketPort  uint
 	l1ConnectionTimeout  time.Duration
 	privateKeys          []string
-	ethereumChainID      big.Int
-	obscuroChainID       big.Int
+	l1ChainID            int64
```
joeldudley, line 62: Harmonised config over in network manager tool as part of same PR.

PR #334, opened 2022-06-29
Smoketest checks L2 transactions

joeldudley:

Currently, the smoketest only checks L1 transactions. It should check L2 transactions as well.

Functional.

- Checks success of L2 transactions

### DIFF
--- a/integration/simulation/validate_chain.go
+++ b/integration/simulation/validate_chain.go
@@ -17,10 +17,14 @@ import (
 	"github.com/obscuronet/obscuro-playground/go/common"
 )
 
-// The threshold number of transactions below which we consider the simulation to have failed. We generally expect far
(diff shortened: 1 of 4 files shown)


integration/simulation/validate_chain.go
```diff
@@ -224,28 +225,18 @@ func checkBlockchainOfObscuroNode(
 		t.Errorf("Node %d: L2 to L1 Efficiency is %f. Expected:%f", nodeAddr, efficiency, s.Params.L2ToL1EfficiencyThreshold)
 	}
 
-	// check that all expected transactions were included.
```
joeldudley, line 227: Moved this logic into a function, below.

BedrockSquirrel (approved):
LGTM!

PR #337, opened 2022-06-29
Move test log setup to util for reuse

BedrockSquirrel:

- minor refactor because test log setup is duplicated and I was about to duplicate it again (for wallet extension)

- refactor test logging to a logutil

- happy with naming, generic-ness etc? I went for logutil rather than log because it'll be annoying to have SetupTestLog start showing up as an option when you're looking for log.Info

### DIFF
--- a/go/common/log/logutil/README.md
+++ b/go/common/log/logutil/README.md
@@ -1,0 +1,1 @@
+This package contains utils related to logging.
--- a/integration/noderunner/noderunner_test.go
+++ b/integration/noderunner/noderunner_test.go
@@ -3,12 +3,11 @@ package noderunner
 import (
 	"encoding/hex"
 	"fmt"
+	"github.com/obscuronet/obscuro-playground/go/common/log/logutil"
 	"net"
 	"os"
 	"testing"
 	"time"
-
-	"github.com/obscuronet/obscuro-playground/go/common/log"
 
 	"github.com/obscuronet/obscuro-playground/go/config"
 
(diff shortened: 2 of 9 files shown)


go/common/log/logutil/README.md
```diff
@@ -0,0 +1 @@
+This package contains utils related to logging.
```
joeldudley, line 1: Do these utils really need another package, vs just putting them in the `log` package?

integration/noderunner/noderunner_test.go
```diff
@@ -126,16 +126,9 @@ func tcpConnectionAvailable(addr string) bool {
...
-		panic(err)
-	}
-	timeFormatted := time.Now().Format("2006-01-02_15-04-05")
-	f, err := os.CreateTemp(testLogs, fmt.Sprintf("noderunner-%s-*.txt", timeFormatted))
-	if err != nil {
-		panic(err)
-	}
-	log.OutputToFile(f)
-	return f
+	return logutil.SetupTestLog(&logutil.TestLogCfg{
```
joeldudley, line 129: I'd kill off the use of a struct - it's only three params - and avoid the wrapping of `logutil.SetupTestLog` inside `setupTestLog`, which seems unnecessary.

joeldudley: I'm going to approve because it's a matter of opinion, but I disagree.

Someone will change the code - most likely not you - by adding a param that should default to `true`, and will forget to change all the usages, and a bug will appear. Declaring all the params forces you to input them all. The cost of going around putting `true` or `false` in ~10 places through the codebase is very minor IMO.

BedrockSquirrel: @joeldudley yeah I hadn't considered how those default values can silently bite you that is a v good point.

But I think if it was at 4 parameters already then we wouldn't be having the debate and the person that came along to add that next param may well think "this is getting a bit hard to read, maybe it's time to wrap these up in a struct".

Anyway, thanks for approving, and we can always change it, be interesting to hear what the others think for similar examples in future (probably a resounding "I don't care" haha). Not being able to provide defaults does feel like a pretty major pain point in the language.

BedrockSquirrel: Would add as well, I don't think updating things is the only cost to not using a config. The readability is potentially an even bigger factor (I know Goland can give you nice helper labels for the fields but not every reader will have those available, especially looking at git diffs and stuff).

PR #338, opened 2022-06-30
Fixes serving of static files.

joeldudley:

Serving of static files was previously broken. The Go binary would simply point to the location of the static files on disk, outside the binary. This worked fine when compiling the binary yourself, since it pointed to your files. But if the binary was moved to another machine, the files couldn't be found.

Functional.

- Embeds static files
- Adds `.exe` to Windows binary in readme instructions

### DIFF
--- a/tools/walletextension/wallet_extension.go
+++ b/tools/walletextension/wallet_extension.go
@@ -3,14 +3,13 @@ package walletextension
 import (
 	"context"
 	"crypto/rand"
+	"embed"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"io/fs"
 	"io/ioutil"
 	"net/http"
-	"path"
-	"path/filepath"
-	"runtime"
 	"strings"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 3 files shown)


tools/walletextension/wallet_extension.go
```diff
@@ -93,17 +90,24 @@ func NewWalletExtension(config Config) *WalletExtension {
...
+
 	// Handles Ethereum JSON-RPC requests received over HTTP.
 	serveMux.HandleFunc(pathRoot, we.handleHTTPEthJSON)
 	serveMux.HandleFunc(PathReady, we.handleReady)
-	// Handles the management of viewing keys.
-	serveMux.Handle(pathViewingKeys, http.StripPrefix(pathViewingKeys, http.FileServer(http.Dir(path.Join(basepath, staticDir)))))
 	serveMux.HandleFunc(PathGenerateViewingKey, we.handleGenerateViewingKey)
 	serveMux.HandleFunc(PathSubmitViewingKey, we.handleSubmitViewingKey)
+
+	// Handles the management of viewing keys.
```
BedrockSquirrel, line 100: Maybe could mention again that it is web assets getting served here, confused me a bit at first. Like `// Static dir contains the web interface for management of viewing keys` or something.

BedrockSquirrel (approved):
Good spot, hadn't come across the go:embed stuff before. LGTM (with optional comment tweak)

PR #340, opened 2022-06-30
Add logging for wallet ext simulations

BedrockSquirrel:

- hard to debug wallet ext tests when the nodes in the network are a black box

- add log file per test
- add the config for github to include those files in the downloadable bundle

- my main concern is that this might be a bit heavy handed, it spams out a lot of these files. But we can always rethink it once things have settled down I suppose...

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -48,4 +48,5 @@ jobs:
           path: |
             integration/.build/simulations/sim-log-*.txt
             integration/.build/noderunner/noderunner-*.txt
+            integration/.build/wallet_extension/wal-ext-*.txt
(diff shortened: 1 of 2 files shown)


tudor-malene (approved):
lgtm

joeldudley (approved):
IMO a log per test is overkill but fine by me if that's what people prefer.

BedrockSquirrel:
> IMO a log per test is overkill but fine by me if that's what people prefer.

Yeah I'm inclined to agree but I don't know what other option there is really, given there's a network per test. Could find a way to set the logger before all the tests and then have each test just append but think that's harder to debug and same amount of space.

joeldudley:
If we're going to revisit, let's add a todo.

PR #341, opened 2022-06-30
Checks whether HTTP request to add viewing key failed.

joeldudley:

In the wallet extension tests, if the request to add a viewing key returns successfully but with a non-200 error code, this is treated as a success. We should fail the test instead. Otherwise, we get confusing errors later on.

Functional.

- Catch broader set of failures to add viewing key


tudor-malene (approved):
lgtm

PR #348, opened 2022-07-02
Preallocs balance.

joeldudley:

Currently, txs fail because no balances are preallocated.

Functional.

- Preallocates balance to sender of tx (short-term solution)


tudor-malene (approved):
lgtm

PR #349, opened 2022-07-02
Fixes signer for tx receipts.

joeldudley:

Previously, the incorrect signer type was used to get the sender for tx receipts, preventing tx receipts from being retrieved.

Functional.

- Switches to correct signer type


tudor-malene (approved):
lgtm

PR #351, opened 2022-07-04
Add viewing key encryption/decryption to rpc client lib

BedrockSquirrel:

- We want to re-use the viewing key encryption/decryption for tests and scripts
- Communicating to the obscuro node is done a bit differently in many places, adding the enc/dec into the rpc client should give us a single codepath to worry about

- move the `isSensitive(method)` function and the encrypt/decrypt if necessary out from wallet extensions to be part of rpc client lib
- change wallet extension to use rpc client lib
- fix simulation tests (transaction injector) to use rpc client in a way that is consistent

- `obscuro_client.go` in the rpcclientlib, the wallet_extension changes and the transaction injector changes
- there will be follow-ups to:
    * make viewing key enc/dec mandatory (means updating lots of test code and removing the viewing key disabled flag/codepaths from enclave code)
    * refactor to make sure encryption and transport encoding details don't leak into the logic implementations in the enclave/rollup_chain
    * make more methods sensitive/encrypted (i.e. the getNonce)

### DIFF
--- a/go/enclave/rpcencryptionmanager/rpc_encryption_manager.go
+++ b/go/enclave/rpcencryptionmanager/rpc_encryption_manager.go
@@ -44,10 +44,12 @@ func NewRPCEncryptionManager(viewingKeysEnabled bool, enclavePrivateKeyECIES *ec
 
 // DecryptBytes decrypts the bytes with the enclave's private key if viewing keys are enabled.
 func (rpc *RPCEncryptionManager) DecryptBytes(encryptedBytes []byte) ([]byte, error) {
-	if !rpc.viewingKeysEnabled {
-		return encryptedBytes, nil
+	bytes, err := rpc.enclavePrivateKeyECIES.Decrypt(encryptedBytes, nil, nil)
+	if err != nil {
+		return nil, fmt.Errorf("could not decrypt bytes with enclave private key. Cause: %w", err)
 	}
-	return rpc.decryptWithEnclavePrivateKey(encryptedBytes)
+
+	return bytes, nil
 }
 
 // AddViewingKey - see the description of Enclave.AddViewingKey.
@@ -129,12 +131,9 @@ func (rpc *RPCEncryptionManager) EncryptTxReceiptWithViewingKey(address gethcomm
 }
 
 func (rpc *RPCEncryptionManager) ExtractTxFromBinary(encodedTx []byte) (*common.L2Tx, error) {
-	var err error
-	if rpc.viewingKeysEnabled {
-		encodedTx, err = rpc.decryptWithEnclavePrivateKey(encodedTx)
-		if err != nil {
-			return nil, fmt.Errorf("could not decrypt transaction with enclave private key. Cause: %w", err)
-		}
+	encodedTx, err := rpc.DecryptBytes(encodedTx)
+	if err != nil {
+		return nil, fmt.Errorf("could not decrypt transaction with enclave private key. Cause: %w", err)
 	}
 
 	// We need to extract the transaction hex from the JSON list encoding. We remove the leading `"[0x`, and the trailing `]"`.
@@ -148,14 +147,4 @@ func (rpc *RPCEncryptionManager) ExtractTxFromBinary(encodedTx []byte) (*common.
 	}
 
 	return tx, nil
-}
-
-// Decrypts the bytes with the enclave's private key.
-func (rpc *RPCEncryptionManager) decryptWithEnclavePrivateKey(encryptedBytes []byte) ([]byte, error) {
-	bytes, err := rpc.enclavePrivateKeyECIES.Decrypt(encryptedBytes, nil, nil)
-	if err != nil {
-		return nil, fmt.Errorf("could not decrypt bytes with enclave private key. Cause: %w", err)
-	}
-
-	return bytes, nil
 }
--- a/go/enclave/rpcencryptionmanager/rpc_encryption_manager.go
+++ b/go/enclave/rpcencryptionmanager/rpc_encryption_manager.go
@@ -128,15 +128,17 @@ func (rpc *RPCEncryptionManager) EncryptTxReceiptWithViewingKey(address gethcomm
 	return rpc.EncryptWithViewingKey(address, txReceiptBytes)
 }
 
-// DecryptTx decrypts an L2 transaction encrypted with the enclave's public key.
-func (rpc *RPCEncryptionManager) DecryptTx(encryptedTx common.EncryptedTx) (*common.L2Tx, error) {
-	txBinaryListJSON, err := rpc.decryptWithEnclavePrivateKey(encryptedTx)
-	if err != nil {
-		return nil, fmt.Errorf("could not decrypt transaction with enclave private key. Cause: %w", err)
+func (rpc *RPCEncryptionManager) ExtractTxFromBinary(encodedTx []byte) (*common.L2Tx, error) {
+	var err error
+	if rpc.viewingKeysEnabled {
+		encodedTx, err = rpc.decryptWithEnclavePrivateKey(encodedTx)
+		if err != nil {
+			return nil, fmt.Errorf("could not decrypt transaction with enclave private key. Cause: %w", err)
+		}
 	}
 
 	// We need to extract the transaction hex from the JSON list encoding. We remove the leading `"[0x`, and the trailing `]"`.
-	txBinary := txBinaryListJSON[4 : len(txBinaryListJSON)-2]
+	txBinary := encodedTx[4 : len(encodedTx)-2]
 	txBytes := gethcommon.Hex2Bytes(string(txBinary))
 
 	tx := &common.L2Tx{}
--- a/go/host/in_mem_obscuro_client.go
+++ b/go/host/in_mem_obscuro_client.go
@@ -2,8 +2,10 @@ package host
 
 import (
 	"context"
+	"crypto/rand"
 	"encoding/json"
 	"fmt"
+	"github.com/ethereum/go-ethereum/crypto"
 
 	"github.com/ethereum/go-ethereum/crypto/ecies"
 
@@ -15,16 +17,30 @@ import (
 	"github.com/obscuronet/obscuro-playground/go/rpcclientlib"
 )
 
+const (
+	// todo: this is a convenience for testnet testing and will eventually be retrieved from the L1
+	enclavePublicKeyHex = "034d3b7e63a8bcd532ee3d1d6ecad9d67fca7821981a044551f0f0cbec74d0bc5e"
+)
+
 // An in-memory implementation of `rpcclientlib.Client` that speaks directly to the node.
 type inMemObscuroClient struct {
-	obscuroAPI ObscuroAPI
-	ethAPI     EthereumAPI
+	obscuroAPI       ObscuroAPI
+	ethAPI           EthereumAPI
+	enclavePublicKey *ecies.PublicKey
 }
 
 func NewInMemObscuroClient(host *Node) rpcclientlib.Client {
+	// todo: this is a convenience for testnet but needs to replaced by a parameter and/or retrieved from the target host
+	enclPubECDSA, err := crypto.DecompressPubkey(gethcommon.Hex2Bytes(enclavePublicKeyHex))
+	if err != nil {
+		panic(err)
+	}
+	enclPubKey := ecies.ImportECDSAPublic(enclPubECDSA)
+
 	return &inMemObscuroClient{
-		obscuroAPI: *NewObscuroAPI(host),
-		ethAPI:     *NewEthereumAPI(host),
+		obscuroAPI:       *NewObscuroAPI(host),
+		ethAPI:           *NewEthereumAPI(host),
+		enclavePublicKey: enclPubKey,
 	}
 }
 
@@ -42,7 +58,11 @@ func (c *inMemObscuroClient) Call(result interface{}, method string, args ...int
 		if err != nil {
 			return fmt.Errorf("failed to marshal the rpc args for %s - %w", rpcclientlib.RPCSendRawTransaction, err)
 		}
-		_, err = c.ethAPI.SendRawTransaction(context.Background(), bytes)
+		enc, err := c.encryptParamBytes(bytes)
+		if err != nil {
+			return err
+		}
+		_, err = c.ethAPI.SendRawTransaction(context.Background(), enc)
 		return err
 
 	case rpcclientlib.RPCGetCurrentBlockHead:
@@ -87,8 +107,11 @@ func (c *inMemObscuroClient) Call(result interface{}, method string, args ...int
 		if err != nil {
 			return fmt.Errorf("failed to marshal the rpc args for %s - %w", rpcclientlib.RPCGetTransactionByHash, err)
 		}
-
-		encryptedTx, err := c.ethAPI.GetTransactionByHash(context.Background(), bytes)
+		enc, err := c.encryptParamBytes(bytes)
+		if err != nil {
+			return err
+		}
+		encryptedTx, err := c.ethAPI.GetTransactionByHash(context.Background(), enc)
 		if err != nil {
 			return fmt.Errorf("`eth_getTransactionByHash` call failed. Cause: %w", err)
 		}
@@ -102,8 +125,11 @@ func (c *inMemObscuroClient) Call(result interface{}, method string, args ...int
 		if err != nil {
 			return fmt.Errorf("failed to marshal the rpc args for %s - %w", rpcclientlib.RPCCall, err)
 		}
-
-		encryptedResponse, err := c.ethAPI.Call(context.Background(), bytes)
+		enc, err := c.encryptParamBytes(bytes)
+		if err != nil {
+			return err
+		}
+		encryptedResponse, err := c.ethAPI.Call(context.Background(), enc)
 		if err != nil {
 			return fmt.Errorf("`eth_call` call failed. Cause: %w", err)
 		}
@@ -128,8 +154,11 @@ func (c *inMemObscuroClient) Call(result interface{}, method string, args ...int
 		if err != nil {
 			return fmt.Errorf("failed to marshal the rpc args for %s - %w", rpcclientlib.RPCGetTxReceipt, err)
 		}
-
-		encryptedResponse, err := c.ethAPI.GetTransactionReceipt(context.Background(), bytes)
+		enc, err := c.encryptParamBytes(bytes)
+		if err != nil {
+			return err
+		}
+		encryptedResponse, err := c.ethAPI.GetTransactionReceipt(context.Background(), enc)
 		if err != nil {
 			return fmt.Errorf("`obscuro_getTransactionReceipt` call failed. Cause: %w", err)
 		}
@@ -155,4 +184,12 @@ func (c *inMemObscuroClient) SetViewingKey(_ *ecies.PrivateKey, _ []byte) {
 
 func (c *inMemObscuroClient) RegisterViewingKey(_ gethcommon.Address, _ []byte) error {
 	panic("viewing key encryption/decryption is not currently supported by in-memory obscuro-client")
+}
+
+func (c *inMemObscuroClient) encryptParamBytes(params []byte) ([]byte, error) {
+	encryptedParams, err := ecies.Encrypt(rand.Reader, c.enclavePublicKey, params, nil, nil)
+	if err != nil {
+		return nil, fmt.Errorf("could not encrypt request params with enclave public key: %w", err)
+	}
+	return encryptedParams, nil
 }
--- a/go/host/in_mem_obscuro_client.go
+++ b/go/host/in_mem_obscuro_client.go
@@ -5,6 +5,7 @@ import (
 	"crypto/rand"
 	"encoding/json"
 	"fmt"
+
 	"github.com/ethereum/go-ethereum/crypto"
 
 	"github.com/ethereum/go-ethereum/crypto/ecies"
@@ -45,25 +46,13 @@ func NewInMemObscuroClient(host *Node) rpcclientlib.Client {
 }
 
 // Call bypasses RPC, and invokes methods on the node directly.
-func (c *inMemObscuroClient) Call(result interface{}, method string, args ...interface{}) error { //nolint:gocognit
+func (c *inMemObscuroClient) Call(result interface{}, method string, args ...interface{}) error {
 	switch method {
 	case rpcclientlib.RPCGetID:
 		*result.(*gethcommon.Address) = c.obscuroAPI.GetID()
 
 	case rpcclientlib.RPCSendRawTransaction:
-		if len(args) != 1 {
-			return fmt.Errorf("expected 1 arg to %s, got %d", rpcclientlib.RPCSendRawTransaction, len(args))
-		}
-		bytes, err := json.Marshal(args)
-		if err != nil {
-			return fmt.Errorf("failed to marshal the rpc args for %s - %w", rpcclientlib.RPCSendRawTransaction, err)
-		}
-		enc, err := c.encryptParamBytes(bytes)
-		if err != nil {
-			return err
-		}
-		_, err = c.ethAPI.SendRawTransaction(context.Background(), enc)
-		return err
+		return c.sendRawTransaction(args)
 
 	case rpcclientlib.RPCGetCurrentBlockHead:
 		*result.(**types.Header) = c.obscuroAPI.GetCurrentBlockHead()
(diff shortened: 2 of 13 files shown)


go/enclave/rpcencryptionmanager/rpc_encryption_manager.go
```diff
@@ -140,9 +139,12 @@ func (rpc *RPCEncryptionManager) DecryptTx(encryptedTx common.EncryptedTx) (*com
 		return nil, fmt.Errorf("could not decrypt transaction with enclave private key. Cause: %w", err)
 	}
 
-	// We need to extract the transaction hex from the JSON list encoding. We remove the leading `"[0x`, and the trailing `]"`.
-	txBinary := txBinaryListJSON[4 : len(txBinaryListJSON)-2]
-	txBytes := gethcommon.Hex2Bytes(string(txBinary))
+	var txList []string
+	err = json.Unmarshal(txBinaryListJSON, &txList)
+	if err != nil || txList == nil || len(txList) == 0 {
+		return nil, fmt.Errorf("failed to extract from tx bytes from json binary: %w", err)
```
joeldudley, line 145: Typo here: "from tx bytes from json"

go/enclave/rpcencryptionmanager/rpc_encryption_manager.go
```diff
@@ -140,9 +139,12 @@ func (rpc *RPCEncryptionManager) DecryptTx(encryptedTx common.EncryptedTx) (*com
 		return nil, fmt.Errorf("could not decrypt transaction with enclave private key. Cause: %w", err)
 	}
 
-	// We need to extract the transaction hex from the JSON list encoding. We remove the leading `"[0x`, and the trailing `]"`.
-	txBinary := txBinaryListJSON[4 : len(txBinaryListJSON)-2]
-	txBytes := gethcommon.Hex2Bytes(string(txBinary))
+	var txList []string
+	err = json.Unmarshal(txBinaryListJSON, &txList)
+	if err != nil || txList == nil || len(txList) == 0 {
```
joeldudley, line 144: I think we should give a different error for the final condition - `len(txList) == 0`. That implies that you submitted nothing, not that there was some extraction failure.

go/rpcclientlib/README.md
```diff
@@ -1 +1,16 @@
...
+
+Viewing keys generated inside the wallet extension (or other users of the obscuro rpc client), and then signed by the wallet (e.g. MetaMask)
+to which the keys relate.
+The keys are then are sent to the enclave via RPC and processed by:
+- checking the validity of the signature over the viewing key
+- finding the account which has empowered this viewing key.
+We can do that by retrieving the signing public key from the signature.
+By hashing the public key, we can then determine the address of the account.
+- finally the enclave will save the viewing key (which is a public key) against the account, and use it to encrypt any
+sensitive requests (e.g. "eth_call" and "eth_getBalance") that have that address as a "from" field.
```
joeldudley, line 13: It's not always the `from` field. The logic varies. e.g. for a `eth_getBalance`, we encrypt it with the viewing key associated with the address being queried.

go/rpcclientlib/README.md
```diff
@@ -1 +1,16 @@
...
\ No newline at end of file
+This package contains library code to allow client applications to connect to Obscuro nodes via RPC.
+
+### Viewing keys
+
+Viewing keys generated inside the wallet extension (or other users of the obscuro rpc client), and then signed by the wallet (e.g. MetaMask)
+to which the keys relate.
+The keys are then are sent to the enclave via RPC and processed by:
+- checking the validity of the signature over the viewing key
+- finding the account which has empowered this viewing key.
```
joeldudley, line 9: More like "finding the account to which this viewing key corresponds".

go/rpcclientlib/obscuro_client.go
```diff
@@ -30,28 +47,132 @@ type Client interface {
 	Call(result interface{}, method string, args ...interface{}) error
 	// Stop closes the client.
 	Stop()
+
+	//RegisterViewingKey()
```
joeldudley, line 51: What's this?

go/rpcclientlib/obscuro_client.go
```diff
@@ -30,28 +47,132 @@ type Client interface {
...
 func (c *clientImpl) Call(result interface{}, method string, args ...interface{}) error {
-	return c.rpcClient.Call(&result, method, args...)
+	if !isSensitive(method) {
+		// for non-sensitive methods we just delegate directly to the geth RPC client
+		return c.rpcClient.Call(&result, method, args...)
+	}
+
+	var encryptedParams interface{}
+	var err error
+	encryptedParams, err = c.encryptArgs(args)
```
joeldudley, line 92: I think maybe we should check the len of args before passing it in here?

go/rpcclientlib/obscuro_client.go
```diff
@@ -30,28 +47,132 @@ type Client interface {
...
+// The result must be a pointer so that package json can unmarshal into it. You
+// 	can also pass nil, in which case the result is ignored.
 func (c *clientImpl) Call(result interface{}, method string, args ...interface{}) error {
-	return c.rpcClient.Call(&result, method, args...)
+	if !isSensitive(method) {
+		// for non-sensitive methods we just delegate directly to the geth RPC client
+		return c.rpcClient.Call(&result, method, args...)
+	}
+
+	var encryptedParams interface{}
```
joeldudley, line 90: Clearer to make the type of this `[]byte`.

go/rpcclientlib/obscuro_client.go
```diff
@@ -30,28 +47,132 @@ type Client interface {
...
+		return nil, fmt.Errorf("could not json encode request params: %w", err)
+	}
+	encryptedParams, err := ecies.Encrypt(rand.Reader, c.enclavePublicKey, paramsJSON, nil, nil)
+	if err != nil {
+		return nil, fmt.Errorf("could not encrypt request params with enclave public key: %w", err)
+	}
+	return encryptedParams, nil
+}
+
+func (c *clientImpl) decryptResponse(resultBlob interface{}) ([]byte, error) {
```
joeldudley, line 152: Check out the changes I had to make to the decryption on `main`. Some responses are just strings, but some are JSON lists. We ran into issues before because the transaction receipts were being returned as double-quoted json strings.

tools/walletextension/wallet_extension.go
```diff
@@ -141,54 +136,36 @@ func (we *WalletExtension) handleHTTPEthJSON(resp http.ResponseWriter, req *http
...
-	maybeEncryptedBody, err := we.encryptParamsIfNeeded(body, method, reqJSONMap)
-	if err != nil {
-		logAndSendErr(resp, fmt.Sprintf("could not encrypt request parameters: %s", err))
-		return
-	}
 
-	// We forward the request on to the Obscuro node.
-	nodeResp, err := forwardMsgOverWebsocket(websocketProtocol+we.hostAddr, maybeEncryptedBody)
+	var method string
+	err = json.Unmarshal(reqJSONMap[reqJSONKeyMethod], &method)
```
joeldudley, line 149: Do we need all this wacky unmarshalling? Can't we just do, `method := reqJSONMap[reqJSONKeyMethod]`?

tools/walletextension/wallet_extension.go
```diff
@@ -141,54 +136,36 @@ func (we *WalletExtension) handleHTTPEthJSON(resp http.ResponseWriter, req *http
...
+	fmt.Printf("Received %s request from wallet: %s\n", method, body)
 
-	// We unmarshall the JSON response.
-	var respJSONMap map[string]interface{}
-	err = json.Unmarshal(nodeResp, &respJSONMap)
+	// we know params is always in a list format, so we do this unpleasantness to be able to spread it into the rpc call
+	var params []interface{}
+	err = json.Unmarshal(reqJSONMap[reqJSONKeyParams], &params)
 	if err != nil {
-		logAndSendErr(resp, fmt.Sprintf("could not unmarshall enclave response to JSON: %s", err))
```
joeldudley, line 173: We should `logAndSendErr` here.

joeldudley:
I think this logic (currently in the wallet extension) should be kicked over to the RPC client. The RPC client should manage encryption + submission of viewing keys IMO, in the same way it handles decryption.

```
        // We encrypt the viewing key bytes.
	encryptedViewingKeyBytes, err := ecies.Encrypt(rand.Reader, we.enclavePublicKey, we.viewingPublicKeyBytes, nil, nil)
	if err != nil {
		logAndSendErr(resp, fmt.Sprintf("could not encrypt viewing key with enclave public key: %s", err))
		return
	}
```

joeldudley (changes requested):
`WalletExtension` still has a `viewingPrivateKeyEcies` field, but it's not used.

go/enclave/rpcencryptionmanager/rpc_encryption_manager.go
```diff
@@ -125,17 +125,23 @@ func (rpc *RPCEncryptionManager) EncryptTxReceiptWithViewingKey(address gethcomm
 
 // DecryptTx decrypts an L2 transaction encrypted with the enclave's public key.
 func (rpc *RPCEncryptionManager) DecryptTx(encryptedTx common.EncryptedTx) (*common.L2Tx, error) {
+	if !rpc.viewingKeysEnabled {
```
joeldudley, line 128: @BedrockSquirrel I think it's cleaner to move this check down into `extractTxFromBinary`, and just return the bytes from there if there's no decryption needed.

go/rpcclientlib/README.md
```diff
@@ -1 +1,16 @@
...
+This package contains library code to allow client applications to connect to Obscuro nodes via RPC.
+
+### Viewing keys
+
+Viewing keys are generated inside the wallet extension (or other users of the obscuro rpc client), and then signed by the wallet (e.g. MetaMask)
+to which the keys relate.
+The keys are then are sent to the enclave via RPC and processed by:
+- checking the validity of the signature over the viewing key
+- finding the account to which this viewing key corresponds
+We can do that by retrieving the signing public key from the signature.
```
joeldudley, line 10: Don't think this formatting will work. This needs to be a sub-bullet I think.

go/rpcclientlib/README.md
```diff
@@ -1 +1,16 @@
...
+By hashing the public key, we can then determine the address of the account.
+- finally the enclave will save the viewing key (which is a public key) against the account, and use it to encrypt any
+sensitive requests (e.g. "eth_call" and "eth_getBalance") permitted to be viewed by that account
```
joeldudley, line 13: Think you need to indent this a bit as well, or it won't be considered part of the bullet

go/rpcclientlib/obscuro_client.go
```diff
@@ -19,39 +29,228 @@ const (
...
+	// RegisterViewingKey takes a signature for the public key, it verifies the signed public key matches the currently set private viewing key
+	//	and then submits it to the enclave
+	RegisterViewingKey(signature []byte) error
 }
 
-// A Client implementation that wraps rpc.Client to make calls.
-type clientImpl struct {
-	rpcClient *rpc.Client
+// RPCClient is a Client implementation that wraps rpc.Client to make calls.
+type RPCClient struct {
```
joeldudley, line 69: @BedrockSquirrel I think this struct should be private, and `NewClient` should return a `Client`. Exposing `ViewingKeyAddr` isn't a good enough reason to expose the specific implementation.

I think the `from` field for eth_call requests should be set here, rather than in the wallet extension. Then you won't need access to expose `ViewingKeyAddr`.

go/rpcclientlib/obscuro_client.go
```diff
@@ -19,39 +29,228 @@ const (
...
+
+func (c *RPCClient) SetViewingKey(viewingKey *ecies.PrivateKey, viewingPubKeyBytes []byte) {
+	c.viewingKey = viewingKey
+	c.viewingPubKey = viewingPubKeyBytes
+}
+
+func (c *RPCClient) RegisterViewingKey(signature []byte) error {
+	// recover the signer address from the signature and store it
+	// verify signature contains expected pub key
+	msgToSign := ViewingKeySignedMsgPrefix + hex.EncodeToString(c.viewingPubKey)
```
joeldudley, line 227: @BedrockSquirrel IMO it doesn't make sense to recover the viewing key address here. Just pass it in as a parameter.

We need to recover it on the enclave side to avoid attacks where you lie about which address the signature was for, but that attack isn't relevant here as we are not crossing a trust boundary.

go/rpcclientlib/obscuro_client.go
```diff
@@ -19,39 +29,228 @@ const (
...
+	}
 }
 
-func (c *clientImpl) Stop() {
+func (c *RPCClient) Stop() {
 	c.rpcClient.Close()
 }
+
+func (c *RPCClient) SetViewingKey(viewingKey *ecies.PrivateKey, viewingPubKeyBytes []byte) {
+	c.viewingKey = viewingKey
```
joeldudley, line 220: @BedrockSquirrel Can we rename this to `c.viewingPrivKey` or similar for clarity?

go/rpcclientlib/obscuro_client.go
```diff
@@ -19,39 +29,228 @@ const (
...
+	// SetViewingKey sets the current viewing key on the client to be used for all sensitive request decryption
+	SetViewingKey(viewingKey *ecies.PrivateKey, pubKeyBytes []byte)
+
+	// RegisterViewingKey takes a signature for the public key, it verifies the signed public key matches the currently set private viewing key
+	//	and then submits it to the enclave
+	RegisterViewingKey(signature []byte) error
 }
 
-// A Client implementation that wraps rpc.Client to make calls.
-type clientImpl struct {
```
joeldudley, line 36: @BedrockSquirrel Not sure why we renamed this?

go/rpcclientlib/obscuro_client.go
```diff
@@ -19,39 +29,228 @@ const (
...
+	if c.viewingKey != nil {
+		// method is sensitive, so we decrypt it before unmarshalling the result
+		decrypted, err = c.decryptResponse(rawResult)
+		if err != nil {
+			return fmt.Errorf("failed to decrypt args for rpc call - %w", err)
+		}
+	}
+
+	// process the decrypted result to get the desired type and set it on the result pointer
+	err = c.prepareResult(decrypted, result)
```
joeldudley, line 154: @BedrockSquirrel Can this get a better name? `setResult` is better than `prepareResult`, I think.

go/rpcclientlib/obscuro_client.go
```diff
@@ -19,39 +29,228 @@ const (
...
+	}
+
+	// if caller not interested in response, we're done
+	if result == nil {
+		return nil
+	}
+
+	// todo: remove this non-decryption part when we make viewing key encryption mandatory across all tests
+	// extract result from the data as-is, in case we can't decrypt/process it below
+	decrypted, ok := rawResult.([]byte) // if viewing key was nil we try and extract result from the data as-is
```
joeldudley, line 138: @BedrockSquirrel This should be in an `if`/`else`. Otherwise, we do all the processing, only to find out the viewing key wasn't nil, and so we should overwrite that processing with a completely different result.

go/rpcclientlib/obscuro_client.go
```diff
@@ -19,39 +29,228 @@ const (
...
+
+	// we setup a generic rawResult to receive the response (then we can decrypt it as necessary into the requested result type)
+	var err error
+	var rawResult interface{}
+	if result == nil {
+		// we set result receiver to nil if caller set nil since no return needs to be set
+		rawResult = nil
+	}
+
+	if len(args) == 0 {
```
joeldudley, line 115: @BedrockSquirrel Can you just encrypt the args no matter what, even if it's an empty list? Think it would make the code much cleaner here.

go/rpcclientlib/obscuro_client.go
BedrockSquirrel: I think impl is a legacy naming convention that comes from languages that required interfaces for everything and it's generally better practice to give implementations a meaningful name where poss.

Client has two implementations, InMem and RPC seems better than InMem and Impl

(But also it was because I stopped using the interface in the wallet extension which you quite rightly called me out on haha)

go/rpcclientlib/obscuro_client.go
BedrockSquirrel: Yeah I did that originally and it fell foul of the complexity linter (and also required jumping through some var declaring hoops). Was lazy with it because I don't think this snippet will live very long, but you're right, I'll sort it out.

go/rpcclientlib/obscuro_client.go
joeldudley: Maybe make it `InMem` and `Network`?

tools/walletextension/wallet_extension.go
```diff
@@ -54,20 +46,11 @@ const (
 	corsAllowHeaders = "Access-Control-Allow-Headers"
 	corsHeaders      = "Accept, Content-Type, Content-Length, Accept-Encoding, X-CSRF-Token, Authorization"
 
-	Localhost         = "127.0.0.1"
-	websocketProtocol = "ws://"
+	Localhost = "127.0.0.1"
```
joeldudley, line 49: @BedrockSquirrel This can move over to the `main` package, since that's it's only usage.

tools/walletextension/wallet_extension.go
```diff
@@ -153,105 +131,111 @@ func (we *WalletExtension) handleHTTPEthJSON(resp http.ResponseWriter, req *http
...
 		return
 	}
 
-	// We report any errors from the request.
-	if respJSONMap[respJSONKeyErr] != nil {
-		logAndSendErr(resp, respJSONMap[respJSONKeyErr].(map[string]interface{})[respJSONKeyMsg].(string))
-		return
-	}
+	respMap := make(map[string]interface{})
+	respMap[resJSONKeyID] = reqID
```
joeldudley, line 171: @BedrockSquirrel What's this addition about? The response should be coming from the host properly formatted already (e.g. with the ID and the JSON version.

joeldudley: Oh, I see. We're now making an explicit JSON call, rather than passing something back and forth.

tools/walletextension/wallet_extension.go
```diff
@@ -153,105 +131,111 @@ func (we *WalletExtension) handleHTTPEthJSON(resp http.ResponseWriter, req *http
...
 
-// If an `eth_call` request doesn't have a `from` field, we won't be able to encrypt the response. In that case, we use
-// the viewing key address as the `from` field to allow encryption and decryption.
-func (we *WalletExtension) ensureCallsHaveFromField(method interface{}, reqJSONMap map[string]interface{}) (map[string]interface{}, error) {
-	if method != ReqJSONMethodCall {
-		// We only modify `eth_call` requests.
-		return reqJSONMap, nil
+// extract.AndPrepareParams extracts the params from the request to pass into rpc client.
+//	May also modify params etc. as needed for specific methods
+func (we *WalletExtension) extractAndPrepareParams(method string, reqJSONMap map[string]json.RawMessage) ([]interface{}, error) {
```
joeldudley, line 200: @BedrockSquirrel I don't find this bundling of two things very intuitive (extracting the params, and modifying them. I think it works better as two separate methods.

tools/walletextension/wallet_extension.go
```diff
@@ -153,105 +131,111 @@ func (we *WalletExtension) handleHTTPEthJSON(resp http.ResponseWriter, req *http
...
+	if err != nil {
+		return nil, fmt.Errorf("could not unmarshal params list from JSON-RPC request body: %w", err)
 	}
 
-	params, ok := reqJSONMap[reqJSONKeyParams].([]interface{})
-	if !ok {
-		return nil, fmt.Errorf("params for %s request were malformed", method)
+	// If an `eth_call` request doesn't have a `from` field, we won't be able to encrypt the response. In that case, we use
+	// the viewing key address as the `from` field to allow encryption and decryption.
+	if method == ReqJSONMethodCall {
```
joeldudley, line 210: @BedrockSquirrel I think for ease of reading, this check should be inside the method call to `ensureParamsHaveFromField`.

tools/walletextension/wallet_extension.go
```diff
@@ -153,105 +131,111 @@ func (we *WalletExtension) handleHTTPEthJSON(resp http.ResponseWriter, req *http
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to ensure params include from field: %w", err)
+		}
 	}
+
+	return params, nil
+}
+
+// ensureParamsHaveFromField is used for call methods whose initial params element needs the viewing key address as its 'from' field
+func (we *WalletExtension) ensureParamsHaveFromField(params []interface{}) ([]interface{}, error) {
```
joeldudley, line 221: @BedrockSquirrel Can we rename this to `ensureCallParamsHaveFromField` for clarity?

tools/walletextension/wallet_extension.go
```diff
@@ -153,105 +131,111 @@ func (we *WalletExtension) handleHTTPEthJSON(resp http.ResponseWriter, req *http
 		return
 	}
 
-	// We unmarshall the JSON request.
-	var reqJSONMap map[string]interface{}
+	// We unmarshall the JSON request
```
joeldudley, line 134: @BedrockSquirrel I'd be tempted to put all the unmarshalling into a method, that returns either the ID, the method and the params, or a struct containing all three. That will help prevent the existing method from getting too long.

go/rpcclientlib/obscuro_client.go
BedrockSquirrel: I've left this as-is for now because it comes out dirty anyway, 'all the processing' is just two casts and I'm hoping we'll get rid of it this week.

tools/walletextension/wallet_extension.go
```diff
@@ -33,18 +27,15 @@ const (
...
-	ReqJSONMethodSendRawTx    = "eth_sendRawTransaction"
-	ReqJSONMethodGetTxByHash  = "eth_getTransactionByHash"
-	respJSONKeyErr            = "error"
-	respJSONKeyMsg            = "message"
-	RespJSONKeyResult         = "result"
-	httpCodeErr               = 500
+	reqJSONKeyID             = "id"
+	reqJSONKeyMethod         = "method"
+	reqJSONKeyParams         = "params"
+	ReqJSONMethodCall        = "eth_call"
```
joeldudley, line 33: @BedrockSquirrel This and the constant below can be moved over to the integration test file (not used here).

tools/walletextension/wallet_extension.go
```diff
@@ -153,105 +134,80 @@ func (we *WalletExtension) handleHTTPEthJSON(resp http.ResponseWriter, req *http
...
+	respMap[resJSONKeyID] = rpcReq.id
+	respMap[resJSONKeyRPCVer] = jsonrpc.Version
 
-	// We report any errors from the request.
-	if respJSONMap[respJSONKeyErr] != nil {
-		logAndSendErr(resp, respJSONMap[respJSONKeyErr].(map[string]interface{})[respJSONKeyMsg].(string))
-		return
-	}
+	// respStr could be a plain string value or could be a json encoded object
+	respMap[RespJSONKeyResult] = respStr
```
joeldudley, line 155: @BedrockSquirrel I think it's clearer to have this as an else branch below, if the err isn't nil (so on line 163).

tools/walletextension/wallet_extension.go
```diff
@@ -78,23 +58,24 @@ type WalletExtension struct {
...
-	//  to indicate which viewing key they were encrypted with.
-	viewingPublicKeyBytes  []byte
-	viewingPrivateKeyEcies *ecies.PrivateKey
-	// The address associated with the last viewing key submitted. Used to set missing `from` fields in `eth_call` requests.
-	viewingPublicKeyAddress common.Address
-	server                  *http.Server
+	server           *http.Server
+}
+
+type RPCRequest struct {
```
joeldudley, line 64: @BedrockSquirrel This type can be private.

tools/walletextension/main/main.go
```diff
@@ -6,12 +6,16 @@ import (
 	"github.com/obscuronet/obscuro-playground/tools/walletextension"
 )
 
+const (
+	Localhost = "127.0.0.1"
```
joeldudley, line 10: @BedrockSquirrel Can be private.

go/host/in_mem_obscuro_client.go
```diff
@@ -35,12 +38,11 @@ func (c *inMemObscuroClient) Call(result interface{}, method string, args ...int
 		if len(args) != 1 {
 			return fmt.Errorf("expected 1 arg to %s, got %d", rpcclientlib.RPCSendRawTransaction, len(args))
 		}
-		tx, ok := args[0].(common.EncryptedParamsSendRawTx)
-		if !ok {
-			return fmt.Errorf("arg to %s was not of expected type EncryptedParamsSendRawTx", rpcclientlib.RPCSendRawTransaction)
+		b, err := json.Marshal(args)
```
joeldudley, line 41: @BedrockSquirrel Can we use `bytes` instead of `b` as the variable name, here and below? We also use `b` for blocks, it's too opaque.

go/host/in_mem_obscuro_client.go
```diff
@@ -140,3 +144,11 @@ func (c *inMemObscuroClient) Call(result interface{}, method string, args ...int
 func (c *inMemObscuroClient) Stop() {
 	// There is no RPC connection to close.
 }
+
+func (c *inMemObscuroClient) SetViewingKey(viewingKey *ecies.PrivateKey, pubKeyBytes []byte) {
```
joeldudley, line 148: @BedrockSquirrel If all the params aren't used, you can drop their names.

go/host/in_mem_obscuro_client.go
```diff
@@ -140,3 +144,11 @@ func (c *inMemObscuroClient) Call(result interface{}, method string, args ...int
...
+	panic("viewing key encryption/decryption is not currently supported by in-memory obscuro-client")
+}
+
+func (c *inMemObscuroClient) RegisterViewingKey(signerAddr gethcommon.Address, signature []byte) error {
```
joeldudley, line 152: @BedrockSquirrel If all the params aren't used, you can drop their names.

integration/walletextension/wallet_extension_test.go
```diff
@@ -546,19 +557,27 @@ func makeEthJSONReqAsJSON(t *testing.T, walletExtensionAddr string, method strin
 	return respBodyJSON
 }
 
+func generateAndSubmitWalletViewingKey(t *testing.T, walletExtensionAddr string, accountPrivateKey *ecdsa.PrivateKey) {
```
joeldudley, line 560: @BedrockSquirrel I think we can kill off this method? It only has one usage, and only wraps another call.

go/rpcclientlib/obscuro_client.go
```diff
@@ -19,39 +28,264 @@ const (
...
+	// todo: this is a convenience for testnet testing and will eventually be retrieved from the L1
+	enclavePublicKeyHex = "034d3b7e63a8bcd532ee3d1d6ecad9d67fca7821981a044551f0f0cbec74d0bc5e"
+
+	// ViewingKeySignedMsgPrefix is the prefix added when signing the viewing key in MetaMask using the personal_sign
+	// API. Why is this needed? MetaMask has a security feature whereby if you ask it to sign something that looks like
+	// a transaction using the personal_sign API, it modifies the data being signed. The goal is to prevent hackers
+	// from asking a visitor to their website to personal_sign something that is actually a malicious transaction (e.g.
+	// theft of funds). By adding a prefix, the viewing key bytes no longer looks like a transaction hash, and thus get
+	// signed as-is.
+	ViewingKeySignedMsgPrefix = "vk"
```
joeldudley, line 46: @BedrockSquirrel Can you review these constants please? Some are literally unused now, some are public but only used in this file.

go/rpcclientlib/obscuro_client.go
```diff
@@ -19,39 +28,264 @@ const (
...
+// The result must be a pointer so that package json can unmarshal into it. You
+// 	can also pass nil, in which case the result is ignored.
+func (c *networkClient) Call(result interface{}, method string, args ...interface{}) error {
+	if !isSensitive(method) {
+		// for non-sensitive methods or when viewing keys are disabled we just delegate directly to the geth RPC client
+		return c.rpcClient.Call(&result, method, args...)
+	}
+
+	// we setup a generic rawResult to receive the response (then we can decrypt it as necessary into the requested result type)
+	var err error
```
joeldudley, line 107: @BedrockSquirrel You can push this further down IMO, to where it's first used.

go/rpcclientlib/obscuro_client.go
```diff
@@ -19,39 +28,264 @@ const (
...
-		rpcClient: rpcClient,
+	return &networkClient{
+		rpcClient:        rpcClient,
+		enclavePublicKey: enclavePublicKey,
+	}
+}
+
+// Call handles JSON rpc requests - if the method is sensitive it will encrypt the args before sending the request and
+//		then decrypts the response before returning
+// The result must be a pointer so that package json can unmarshal into it. You
```
joeldudley, line 98: @BedrockSquirrel Some pretty random line lengths.

go/rpcclientlib/obscuro_client.go
```diff
@@ -19,39 +28,264 @@ const (
...
+
+	// we setup a generic rawResult to receive the response (then we can decrypt it as necessary into the requested result type)
+	var err error
+	var rawResult interface{}
+	if result == nil {
+		// we set result receiver to nil if caller set nil since no return needs to be set
+		rawResult = nil
+	}
+
+	args, err = c.sanitiseArgs(method, args)
```
joeldudley, line 114: @BedrockSquirrel I don't love this. Quite opaque. Can't we just inline the call for now?

```	if method == RPCCall {
		args, err = c.ensureCallParamsHaveFromAddress(method, args)
		if err != nil {
			return nil, err
		}
	}```

File: go/rpcclientlib/obscuro_client.go [hunk 39]
```diff
@@ -19,39 +28,264 @@ const (
...
+	encryptedParams, err := ecies.Encrypt(rand.Reader, c.enclavePublicKey, params, nil, nil)
+	if err != nil {
+		return nil, fmt.Errorf("could not encrypt request params with enclave public key: %w", err)
+	}
+	return encryptedParams, nil
+}
+
+func (c *networkClient) decryptResponse(resultBlob interface{}) ([]byte, error) {
+	encryptedResult := common.Hex2Bytes(resultBlob.(string))
+	if c.viewingPrivKey == nil {
```
joeldudley, line 189: @BedrockSquirrel We've already checked this above, before calling into this method.

PR #352, opened 2022-07-04
Adds tx hashes to block.

joeldudley:

To reply to `eth_getBlockByNumber` requests meaningfully, we need to include the hashes of the txs in the block (so we can know once they've been included).

Functional.

- Adds tx hashes to the RPC block response


tudor-malene (approved):
lgtm

PR #353, opened 2022-07-04
update geth deployment number

otherview:

- points to the newest geth network


BedrockSquirrel (approved):
lgtm

PR #356, opened 2022-07-04
Obscuro node uses Edgeless DB + testnet deploy

otherview:

- https://github.com/obscuronet/obscuro-internal/issues/592

- Testnet now uses SGX 
- Testnet now uses EdgelessDB
- Local testnet has a `use_sg


testnet/docker-compose.yml
```diff
@@ -24,25 +26,43 @@ services:
...
+    volumes:
+      - /dev/sgx:/dev/sgx
+    networks:
+      - default
+    environment:
+      EDG_EDB_CERT_DNS: edgelessdb
+    ports:
+      - "3306:3306"
+      - "8080:8080"
+    image: ghcr.io/edgelesssys/edgelessdb-sgx-1gb:latest
```
BedrockSquirrel, line 68: There's also a 4GB version of this (same path but with "4-gb" at the end. Maybe worth using that to buy us a bit more time before we have to think about archiving and optimisation or wait til it becomes a problem?

BedrockSquirrel (approved):
LGTM - optional comment for your consideration

PR #358, opened 2022-07-04
Further fixes to enable MetaMask and Remix

joeldudley:

Correctly formats the returned JSON for `eth_getTransactionReceipt`, so that MetaMask can detect a successful transaction.

Also adds the `eth_getBlockByHash` method that MetaMask needs.

Functional.

- As described above.


BedrockSquirrel (approved):
LGTM

PR #360, opened 2022-07-05
Adds eth_getCode method.

joeldudley:

Remix invokes the `eth_getCode` method after successful contract deployment.

Functional.

- Adds support for `eth_getCode` method
- Fixes setting of tx receipt status
- Fixes bug in getRollupByHeight where we didn't support `-1`, which is a valid Eth index


tudor-malene (approved):
lgtm.

Note: so much boilerplate.. We need to find a better way.

PR #361, opened 2022-07-05
Avoids panic if there is no head rollup.

joeldudley:

Previously, if there was no head rollup, `GetRollupByHeight` would panic.

Refactoring.

- Avoids panic


BedrockSquirrel (approved):
LGTM

PR #365, opened 2022-07-05
fix the toml unit test

otherview:

- Broke the unit test because of a flag change and I didn't spotted in the CI :(


joeldudley (approved):
Lgtm

PR #369, opened 2022-07-06
Wallet extension prints warning if MetaMask is not enabled.

joeldudley:

Previously, if MetaMask was not installed, the wallet extension would simply not respond to attempts to generate a viewing key.

Functional.

- Attempting to generate a viewing key without MetaMask enabled prints an error, telling the user to install and enable MetaMask.


BedrockSquirrel (approved):
LGTM

PR #371, opened 2022-07-06
Avoids a separate log file per wallet-extension test.

joeldudley:

Previously, a separate log was created for every wallet extension test, which was too heavy.

Functional.

- Consolidates the logs for the wallet extension tests.


go/common/log/logutil/logutil.go
```diff
@@ -9,14 +9,20 @@ import (
 )
 
 type TestLogCfg struct {
-	LogDir      string // directory for the log file
-	TestType    string // type of test (comes before timestamp in filename so sorted file list will block these together)
-	TestSubtype string // test subtype (comes after timestamp in filename so sorted file list will show latest of different subtypes together)
+	LogFile     *os.File // the log file to use, or nil if a new log file should be created
```
BedrockSquirrel, line 12: As much as I want to try to justify the contentious config block by adding a param to it, I don't think this should be in here haha. It's like calling `setupTestLog("actually  don't set one up")`

Can we just call log.OutputToFile in `setupWalletTestLog()` in the test file.

joeldudley: Fixed.

BedrockSquirrel (approved):
LGTM

PR #372, opened 2022-07-06
Removes incorrect todo.

joeldudley:

The todo being removed is incorrect. There is no need to stop the enclaves, since they are automatically pulled down by the hosts when they are shut down.

- list of changes


BedrockSquirrel (approved):
LGTM

PR #375, opened 2022-07-06
Explains why viewing key must be regenerated on restart.

joeldudley:

It is unclear in the docs why the viewing key must be regenerated upon restart.

Functional.

- Clarifies docs


GavT (approved):
LGTM

PR #377, opened 2022-07-06
Avoids panic on un-encodable message

joeldudley:

Previously, the node panicked if it received a message to broadcast that couldn't be encoded with RLP. This was an obvious attack vector.

Functional.

- Surfaces error instead of panicking when receiving un-encodable message
- Surfaces error when processing blocks
- Renames `Host.broadcastTx` to `Host.broadcastL1Tx` for clarity


go/host/host.go
```diff
@@ -302,11 +306,17 @@ func (a *Node) startProcessing() {
 		select {
 		case b := <-a.blockRPCCh:
 			roundInterrupt = triggerInterrupt(roundInterrupt)
-			a.processBlocks([]common.EncodedBlock{b.p, b.b}, roundInterrupt)
+			err := a.processBlocks([]common.EncodedBlock{b.p, b.b}, roundInterrupt)
+			if err != nil {
+				common.LogWithID(a.shortID, "Could not process block received via RPC. Cause: %v", err)
```
BedrockSquirrel, line 311: Tiny bit annoying to add `WarnWithID()` to logging.go but feels like maybe these should be warn level messages?

Getting dodgy blocks might not be fatal but feels like something operators would want flagged up (since atm I guess it likely will be fatal and this could help debug).

BedrockSquirrel (approved):
LGTM - I think the logging level change below is important but not a blocker, I'm sure we'll want to review all the host logging carefully before we onboard node operators.

PR #380, opened 2022-07-08
Fixing geth config + slowing down block production

otherview:

- Keeps the testnet up for longer

- fixed geth arguments
- changed the deployment from 1 blk/sec to 1 blk/min


tudor-malene (approved):
lgtm

PR #382, opened 2022-07-11
Surfaces errors during encoding and decoding of blocks.

joeldudley:

Encoding or decoding a block panics if there's an issue. This isn't always the correct behaviour. Often (e.g. when receiving blocks via RPC), it is better to handle the error (e.g. to prevent bad blocks from being sent over and crashing the node).

Functional.

- `EncodeBlock` now returns an error instead of panicking
- `DecodeBlock` now returns an error instead of panicking

### DIFF
--- a/go/common/l1_types.go
+++ b/go/common/l1_types.go
@@ -1,9 +1,8 @@
 package common
 
 import (
+	"fmt"
 	"math/big"
-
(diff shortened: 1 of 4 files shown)


BedrockSquirrel (approved):
LGTM

PR #383, opened 2022-07-11
Use `LogWithID` and similar everywhere

joeldudley:

We have introduced `LogWithID` and similar methods to automate prefixing a node's ID to a log message. But we used it inconsistently, in part because we didn't have similar methods for `log.Panic` and `log.Trace`.

Functional.

- Introduces `TraceWithID` and `LogWithID` everywhere
- Uses the logging methods everywhere they apply
- Creates a `nodeShortID` field on the enclave client to avoid endless recalculation of the node's short ID

### DIFF
--- a/go/common/logging.go
+++ b/go/common/logging.go
@@ -9,29 +9,29 @@ import (
 // LogWithID logs a message at INFO level with the aggregator's identity prepended.
 func LogWithID(nodeID uint64, msg string, args ...interface{}) {
 	formattedMsg := fmt.Sprintf(msg, args...)
-	log.Info(fmt.Sprintf(">   Agg%d: %s", nodeID, formattedMsg))
+	log.Info(">   Agg%d: %s", nodeID, formattedMsg)
 }
 
 // WarnWithID logs a message at WARN level with the aggregator's identity prepended.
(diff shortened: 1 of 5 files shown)


go/common/logging.go
```diff
@@ -18,8 +18,20 @@ func WarnWithID(nodeID uint64, msg string, args ...interface{}) {
 	log.Warn(fmt.Sprintf(">   Agg%d: %s", nodeID, formattedMsg))
 }
 
+// TraceWithID logs a message at TRACE level with the aggregator's identity prepended.
+func TraceWithID(nodeID uint64, msg string, args ...interface{}) {
+	formattedMsg := fmt.Sprintf(msg, args...)
+	log.Trace(fmt.Sprintf(">   Agg%d: %s", nodeID, formattedMsg))
```
BedrockSquirrel, line 24: Can we pull this pattern string out into a constant?

BedrockSquirrel (commented):
Minor comment, if you make that change then I think "Agg%d" should only be appearing once across the codebase if you got them all :)

go/common/logging.go
joeldudley: Done

PR #385, opened 2022-07-11
Do not panic when encoding/decoding

joeldudley:

In many cases, we panic when encoding or decoding a rollup or attestation. This is dangerous. The decision on whether to panic should be made further down the stack.

Functional.

- Removes implicit panics when encoding/decoding rollups/attestations
- Combines `Decode` and `DecodeBlock` methods - only one is needed
- Moves block encoding/decoding into the `encoding.go` file for consistency

### DIFF
--- a/go/common/l1_types.go
+++ b/go/common/l1_types.go
@@ -1,12 +1,10 @@
 package common
 
 import (
-	"fmt"
 	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
-	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/ethereum/go-ethereum/trie"
 )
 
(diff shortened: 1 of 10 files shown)


go/common/l1_types.go
```diff
@@ -15,31 +13,6 @@ var GenesisHash = common.HexToHash("00000000000000000000000000000000000000000000
...
-
-func EncodeBlock(b *types.Block) (EncodedBlock, error) {
-	encoded, err := rlp.EncodeToBytes(b)
-	if err != nil {
-		return nil, fmt.Errorf("could not encode block to bytes. Cause: %w", err)
-	}
-	return encoded, nil
-}
-
-func (eb EncodedBlock) Decode() (*types.Block, error) {
```
joeldudley, line 29: This was inlined into the `DecodeBlock` method, below. A single method is fine.

go/common/l1_types.go
```diff
@@ -15,31 +13,6 @@ var GenesisHash = common.HexToHash("00000000000000000000000000000000000000000000
 // GenesisBlock - this is a hack that has to be removed ASAP.
 var GenesisBlock = NewBlock(nil, common.HexToAddress("0x0"), []*types.Transaction{})
 
-// EncodedBlock the encoded version of an ExtBlock
-type EncodedBlock []byte
```
joeldudley, line 19: This logic was moved into the `encoding.go` file for consistency.

BedrockSquirrel (approved):
LGTM

PR #386, opened 2022-07-11
Moves test utils out of common, and into test-specific common.

joeldudley:

Some test utilities ended up in the `go/common/` package, where they were at risk of being used by non-test code. In this particular instance, the issue was that the functions panicked, but in general we should try to segregate our test vs production code.

Functional.

- Creates new `common` package inside `integration`
- Moves certain test utils to that package

### DIFF
--- a/integration/simulation/utils.go
+++ b/integration/simulation/utils.go
@@ -4,6 +4,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"math/big"
+	"time"
+
+	testcommon "github.com/obscuronet/obscuro-playground/integration/common"
 
 	gethcommon "github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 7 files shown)


integration/simulation/utils.go
```diff
@@ -201,3 +204,7 @@ func findRollupDups(list []common.L2RootHash) map[common.L2RootHash]int {
 	}
 	return dups
 }
+
+func SleepRndBtw(min time.Duration, max time.Duration) {
```
joeldudley, line 208: This ended up here since it was only used by the sims, so didn't need to go into the `integration/common/` package.

BedrockSquirrel (approved):
LGTM, good to keep a distinction between the safe-for-prod libs and this stuff

PR #387, opened 2022-07-11
Use random L2 client for transfers in sim, not always the first

joeldudley:

We always use the first L2 client for transfers in the simulation. This might hide issues.

Functional.

- Uses random L2 client for transfers in the simulation


BedrockSquirrel (approved):
LGTM

PR #389, opened 2022-07-11
Shrink TransactionBlobCrypto interface to not expose encrypt and decrypt methods.

joeldudley:

The `TransactionBlobCrypto` class exposes `Encrypt` and `Decrypt` methods, but these should be encapsulated and not used directly.

Functional.

- Removes `Encrypt` and `Decrypt` from `TransactionBlobCrypto` interface
- Fixes tests to work around the encapsulation of these methods

### DIFF
--- a/go/enclave/crypto/transaction_blob_crypto.go
+++ b/go/enclave/crypto/transaction_blob_crypto.go
@@ -24,8 +24,6 @@ const (
 
 // TransactionBlobCrypto handles the encryption and decryption of the transaction blobs stored inside a rollup.
 type TransactionBlobCrypto interface {
-	Encrypt(transactions core.L2Txs) common.EncryptedTransactions
-	Decrypt(encryptedTxs common.EncryptedTransactions) core.L2Txs
 	// ToExtRollup - Transforms an internal rollup as seen by the enclave to an external rollup with an encrypted payload
 	ToExtRollup(r *core.Rollup) common.ExtRollup
 	ToEnclaveRollup(r *common.EncryptedRollup) *core.Rollup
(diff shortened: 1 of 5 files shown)


go/enclave/crypto/transaction_blob_crypto.go
```diff
@@ -50,8 +48,28 @@ func NewTransactionBlobCryptoImpl() TransactionBlobCrypto {
 	}
 }
 
+func (t TransactionBlobCryptoImpl) ToExtRollup(r *core.Rollup) common.ExtRollup {
```
joeldudley, line 51: These are not new, I just hoisted them up the file (since the methods that were above them are no longer public).

go/enclave/rpc_server.go
```diff
@@ -37,6 +37,11 @@ func StartServer(
 	erc20ContractLib erc20contractlib.ERC20ContractLib,
 	collector StatsCollector,
 ) (func(), error) {
+	tomlConfig, err := toml.Marshal(enclaveConfig)
```
joeldudley, line 40: Unrelated but v. minor - it felt right to detect immediately if printing the config would fail, rather than after starting a server.

tools/obscuroscan/obscuroscan_test.go
```diff
@@ -39,6 +41,7 @@ func TestThrowsIfEncryptedRollupIsInvalid(t *testing.T) {
 
 // Generates an encrypted transaction blob in Base64 encoding.
 func generateEncryptedTxBlob(txs []*common.L2Tx) []byte {
-	txBlob := crypto.NewTransactionBlobCryptoImpl().Encrypt(txs)
+	rollup := core.Rollup{Header: &common.Header{}, Transactions: txs}
```
joeldudley, line 44: These changes allow us to get around the removal of the `Encrypt` method from the public API.

tudor-malene (approved):
lgtm

PR #390, opened 2022-07-11
Avoids panic if cannot accept P2P connection. Pushes certain msgs to warn level.

joeldudley:

Previously, the host would panic if a P2P connection failed to form. This was too drastic.

Functional.

- Warns instead of panics if P2P connection fails
- Pushes some info msgs to warn level
- Some refactoring


tudor-malene (approved):
lgtm

PR #391, opened 2022-07-11
Encapsulates decryption of the viewing key.

joeldudley:

Addresses a todo in the codebase. Viewing keys were previously decrypted by the enclave before being passed to the `RPCEncryptionManager`, but that wasn't a good separation of duties.

Functional.

- Encapsulates decryption of viewing key


tudor-malene (approved):
lgtm

PR #392, opened 2022-07-11
Removes a todo that incorrectly assumes that multiple MetaMask accounts can be returned.

joeldudley:

MetaMask's `eth_requestAccounts` method returns an array. I previously assumed that the user should be able to select any of these accounts, and left a todo to that effect. But actually, upon inspecting the APIs docs, it's clear that this array is always only contains one element.

Functional.

- Removes todo
- Adds clarifying comment


tudor-malene (approved):
lgtm

PR #394, opened 2022-07-12
Docs fix. Defaults to testnet node addresses. Better error message on failed request parse.

joeldudley:

- When the enclave fails to parse a JSON request, it doesn't provide sufficient information to diagnose it
- The wallet extension should default to the testnet nodes for ease of use, but doesn't
- There's a typo in the docs ("PRC" instead of "RPC")

Functional.

- Fixes the above


BedrockSquirrel (approved):
LGTM

PR #399, opened 2022-07-13
Better errors when ObscuroScan fails. Catches case where rollup hash is nil.

joeldudley:

Errors in ObscuroScan can be too opaque.

Functional.

- Provides error cause when getting head block or rollup fails
- Catches special case where rollup hash is nil


otherview (approved):
We should prolly change the payload to always have an error field.
Can you add a todo on that please ?

PR #400, opened 2022-07-14
Returns error if rollup cannot be found.

joeldudley:

Currently, attempting to retrieve a rollup by height or by hash from the enclave returns nil if the enclave isn't found. This leads to cases where we assume a rollup is found because we don't check whether fields are nil. It's clearer to return an error.

Functional.

- `GetRollup` returns an error if the rollup cannot be found.
- `GetRollupByHeight` returns an error if the rollup cannot be found.


tudor-malene (approved):
lgtm

PR #404, opened 2022-07-14
enables ifshort in the linter

otherview:

- Reenables the `ifshort` linter 

- Adds a comment that ignores ALL linters in the problem line


tudor-malene (approved):
lgtm

PR #405, opened 2022-07-14
Fix wallet extension tests

joeldudley:

The wallet extension tests had two bugs that were cancelling each other out:

- We were asserting that the result of an ERC20 getBalance call should be empty bytes, when it should be nil bytes
- We were not waiting for the ERC20 contract to be deployed, so we were calling a non-existent contract and getting empty bytes

Functional.

- Fixes the two bugs above

### DIFF
--- a/integration/walletextension/wallet_extension_test.go
+++ b/integration/walletextension/wallet_extension_test.go
@@ -400,6 +400,8 @@ func TestCanSubmitTxAndGetTxReceiptAndTxAfterSubmittingViewingKey(t *testing.T) 
 	}
 	sendTxJSON := makeEthJSONReqAsJSON(t, walletExtensionAddr, walletextension.ReqJSONMethodSendRawTx, []interface{}{txBinaryHex})
 
+	time.Sleep(6 * time.Second) // We wait for the deployment of the contract to the Obscuro network.
+
 	// We get the transaction receipt for the Obscuro ERC20 contract deployment.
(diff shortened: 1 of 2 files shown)


integration/walletextension/wallet_extension_test.go
```diff
@@ -632,7 +631,25 @@ func createObscuroNetwork(t *testing.T) (func(), error) {
...
+	if !ok {
+		panic("could not retrieve transaction hash from JSON result")
+	}
+	counter := 0
+	for {
+		if counter > 20 {
+			t.Fatalf("could not get receipt for Obscuro ERC20 deployment transaction after 20 seconds")
+		}
+		txReceiptResp := makeEthJSONReq(t, walletExtensionAddr, walletextension.ReqJSONMethodGetTxReceipt, []string{txHash})
+		if !strings.Contains(string(txReceiptResp), "could not retrieve transaction with hash") {
```
BedrockSquirrel, line 647: The exiting of this loop is slightly hard to wrap your head around on the first read-through.

Maybe worth a comment to break it down for those of us running on little sleep, along the lines of: "Breakout if we find the hash, or if we see an unexpected failure. If just hash not found we continue to retry."

BedrockSquirrel (approved):
Minor comment, LGTM, well-investigated

integration/walletextension/wallet_extension_test.go
joeldudley: Fixed.

PR #406, opened 2022-07-14
Setting the sugested node execution as is_genesis = true

otherview:

- Easy to forget to setup at least one node as the genesis
- Leads to confusing errors
- Made is_genesis=true as part of the default sugestion


moraygrieve (approved):
LGTM

PR #409, opened 2022-07-14
Management placeholder for ERC20

joeldudley:

When the EVM returns nothing in response to an `eth_call`, we need to encrypt _something_ to return (we cannot encrypt empty bytes). So we encrypt a placeholder. However, the old placeholder made Remix unhappy. We should replace it with a placeholder in a form Remix can understand.

Functional.

- Switches to placeholder Remix can handle


tudor-malene (approved):
ok. let's see if this brings us further

PR #419, opened 2022-07-18
Add viewing keys in tx injector

BedrockSquirrel:

- we want to make sure we test the viewing key encryption paths with the simulations
- we want to use a version of transaction injector to smoke test for testnet issues

- add ability to cache a viewing key in the wallet object
- have the tx injector to get a random client take a wallet as a param and set its viewing key on the client before request
- register a viewing key for each wallet as part of the simulation setup

- tweak the client methods so one lets you set priv viewing key with an address, and one lets you register the public key with a signed version
- update wallet extension for those client changes (which were added to better support the tx injector use case)

- transaction injector
- wallet extension


go/host/in_mem_obscuro_client.go
```diff
@@ -213,12 +213,14 @@ func (c *inMemObscuroClient) Stop() {
 	// There is no RPC connection to close.
 }
 
-func (c *inMemObscuroClient) SetViewingKey(_ *ecies.PrivateKey, _ []byte) {
-	panic("viewing key encryption/decryption is not currently supported by in-memory obscuro-client")
+// viewing key encryption must be disabled on the enclave if using the in-memory client, it doesn't support viewing keys yet
+func (c *inMemObscuroClient) SetViewingKey(_ gethcommon.Address, _ *ecies.PrivateKey) {
+	// viewing key encryption/decryption is not currently supported by in-memory obscuro-client
```
joeldudley, line 218: @BedrockSquirrel Here and below, either we should have a todo, or we should change the text to "is not supported by". Not clear to me whether we plan to add support or not, based on this.

go/rpcclientlib/obscuro_client.go
```diff
@@ -49,11 +49,11 @@ type Client interface {
 	Stop()
 
 	// SetViewingKey sets the current viewing key on the client to be used for all sensitive request decryption
-	SetViewingKey(viewingKey *ecies.PrivateKey, pubKeyBytes []byte)
+	SetViewingKey(viewingKeyAddr common.Address, viewingKey *ecies.PrivateKey)
 
 	// RegisterViewingKey takes a signature for the public key, it verifies the signed public key matches the currently set private viewing key
 	//	and then submits it to the enclave
-	RegisterViewingKey(signerAddr common.Address, signature []byte) error
+	RegisterViewingKey(signature []byte, viewingPubKeyBytes []byte) error
```
joeldudley, line 56: @BedrockSquirrel Can we rename this to `RegisterViewingKeyWithEnclave`? It wasn't clear to me what the difference was with the above.

joeldudley, line 56: @BedrockSquirrel It can just be added as a todo for now, but it seems wrong to me that we re-pass in the viewingPubKeyBytes here, when we've already set the viewing key, so we know what the viewing key is.

Looking at it now, it seems to me that it would be cleaner to get rid of the `SetViewingKey` method, and store the generated viewing key on the wallet side in the wallet extension until it can be signed by MetaMask and submitted to the enclave.

Thoughts?

go/wallet/wallet.go
```diff
@@ -29,6 +31,12 @@ type Wallet interface {
 
 	// PrivateKey returns the wallets private key
 	PrivateKey() *ecdsa.PrivateKey
+
+	// SetViewingPrivKey and GetViewingPrivKey and wallet can cache its current registered viewing private key here.
```
joeldudley, line 35: @BedrockSquirrel This sentence doesn't make sense.

go/wallet/wallet.go
```diff
@@ -29,6 +31,12 @@ type Wallet interface {
...
+	//
+	// In testing we may need to set VK on the rpc client before each request as clients are shared between many wallets
+	SetViewingPrivKey(viewingPrivKey *ecies.PrivateKey)
```
joeldudley, line 38: @BedrockSquirrel Can we add a todo to remove these methods? I'm imagining that once we get to having viewing keys in the sims, we'll need to have a dedicated L2 client per wallet (this is a more accurate representation of real-world usage anyway), in which case I don't believe we'll need this stuff?

integration/simulation/transaction_injector.go
```diff
@@ -171,6 +158,73 @@ func (ti *TransactionInjector) Start() {
...
+	}
+	viewingPrivateKeyEcies := ecies.ImportECDSA(viewingKeyPrivate)
+	w.SetViewingPrivKey(viewingPrivateKeyEcies)
+
+	viewingKeyBytes := crypto.CompressPubkey(&viewingKeyPrivate.PublicKey)
+	viewingKeyHex := hex.EncodeToString(viewingKeyBytes)
+
+	signature := signViewingKey(viewingKeyHex, w.PrivateKey())
+
+	err = ti.rndL2NodeClient(w).RegisterViewingKey(signature, viewingKeyBytes)
```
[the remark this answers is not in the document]
BedrockSquirrel: Yeah thanks, as discussed this is a fatal flaw, can't believe I didn't spot it! Second attempt on a fresh branch, shouldn't need to make the client interface changes and wallet extension changes hopefully.

PR #425, opened 2022-07-19
Removes second constructor for Geth client. Cleans up Geth client ID field.

joeldudley:

- There was a todo to be addressed re: the Geth client having two constructors, when one would do
- The Geth client's ID field was used confusingly. Sometimes it was not set, sometimes it was set to a random ID, sometimes it was set to the L2 node ID

Refactoring.

- Removes secondary constructor for Geth client
- Forces passing of L2 ID when constructing Geth client
- Clarifies role of L2 ID in Geth client
- Encapsulates ID field in ethereummock.Node
- Removes unused param in `Stats.NewRollup`

### DIFF
--- a/go/ethadapter/geth_rpc_client.go
+++ b/go/ethadapter/geth_rpc_client.go
@@ -21,11 +21,11 @@ import (
 // gethRPCClient implements the EthClient interface and allows connection to a real ethereum node
 type gethRPCClient struct {
 	client *ethclient.Client  // the underlying eth rpc client
-	id     gethcommon.Address // TODO remove the id common.Address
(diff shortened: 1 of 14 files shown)


joeldudley:
@BedrockSquirrel Good question. In practice, there's an assumption generally across the tests that there is an Eth client pr Obscuro node. Where this doesn't matter, you can always pass a random ID.

It's not perfect, but in the old structure it was worse, as it was unclear what the ID should be and whether it mattered. I think this is a step forward.

BedrockSquirrel (approved):
LGTM

PR #426, opened 2022-07-19
Removes L2Txs type, per todo.

joeldudley:

Addresses a todo by deleting a useless type alias.

Refactoring.

- Removes `L2Txs` type


BedrockSquirrel (approved):
LGTM, always thought that type looked odd.

FWIW I'm not sure we want a slice of pointers, feels non-standard but I couldn't easily find anything to back that up so best leave it for now.

PR #427, opened 2022-07-19
Uses shared chain config across EVM facade and rollup chain.

joeldudley:

The EVM facade and rollup chain both require the same chain config, but we have hardcoded it twice. We should use a single chain config that's passed around.

Refactoring.

- Uses a single chain config across the EVM facade and rollup chain


tudor-malene (approved):
lgtm

PR #428, opened 2022-07-19
Splits out command line flags for wallet extension.

joeldudley:

Previously, the wallet extension had two flags, one for the HTTP address and one for the websocket address. This didn't make too much sense, as the HTTP and websocket port are always on the same host.

Refactoring.

- Rationalises the wallet extension flags


BedrockSquirrel (approved):
lgtm

PR #429, opened 2022-07-19
Excludes built docsite.

joeldudley:

`docs/_site/` is autogenerated by Jekyll as part of the docs build, and shouldn't be committed. We should .gitignore it.

- Excludes `docs/_site/` from the repo


GavT (approved):
LGTM

PR #431, opened 2022-07-19
Documents supported and sensitive RPC methods

joeldudley:

We do not have any documentation currently on which RPC methods are supported, which ones are encrypted, and the visibility rules on who can decrypt them.

- Adds docs page explaining the above.

### DIFF
--- a/docs/testnet/handling-sensitive-data.md
+++ b/docs/testnet/handling-sensitive-data.md
@@ -1,0 +1,11 @@
+# Handling Sensitive Information
+
+Obscuro supports a subset of Ethereum's [JSON-RPC API](https://ethereum.org/en/developers/docs/apis/json-rpc/).
+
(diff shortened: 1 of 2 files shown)


docs/testnet/handling-sensitive-data.md
```diff
@@ -0,0 +1,49 @@
...
+* `eth_getBalance`
+* `eth_getBlockByNumber`
+* `eth_getBlockByHash`
+* `eth_gasPrice`
+* `eth_call`
+* `eth_getTransactionReceipt`
+* `eth_estimateGas`
+* `eth_sendRawTransaction`
+* `eth_getCode`
+* `eth_getTransactionCount`
```
moraygrieve, line 28: Could eth_getTransactionCount be deemed as sensitive out of curiosity?

moraygrieve (approved):
LGTM

PR #432, opened 2022-07-19
Add viewing keys in tx injector

BedrockSquirrel:

- we want to make sure we test the viewing key encryption paths with the simulations
- we want to use a version of transaction injector to smoke test for testnet issues

*functional*
- Create rpc client per wallet for testing, register viewing key for it
- Bundled together the different types of client that come out of create network

- transaction_injector.go (where the correct client for the wallet is now used with its viewing key for encryption)
- obscuro_node_utils.go (where the wallet clients are setup and viewing keys registered)
- network.Clients struct in network.go (where some convenience methods for interacting with the networks are provided)

### DIFF
--- a/integration/common/viewkey/viewkey.go
+++ b/integration/common/viewkey/viewkey.go
@@ -12,21 +12,36 @@ import (
 	"github.com/obscuronet/go-obscuro/go/wallet"
 )
 
-// GenerateAndRegisterViewingKey take an obscuro client and a wallet, it generates a keypair, simulates signing the key,
-//	configures it on the client and registers the viewing key with the enclave
+// This package contains viewing key utils for testing and simulations
+
+// GenerateAndRegisterViewingKey takes a wallet and the client that will be used for that wallet's transactions.
+//
+//	It sets up a viewing key for that client (without using a wallet extension) with the following steps:
+//	1. generate a "viewing" keypair
+//	2. simulates signing the key with metamask
+//	3. sets the private key on the client (to decrypt enclave responses)
+//	4. registers the public viewing key with the enclave (to encrypt enclave responses)
 func GenerateAndRegisterViewingKey(cli rpcclientlib.Client, wal wallet.Wallet) {
+	// generate an ECDSA key pair to encrypt sensitive communications with the obscuro enclave
 	vk, err := crypto.GenerateKey()
 	if err != nil {
 		panic(err)
 	}
-	viewingPrivateKeyEcies := ecies.ImportECDSA(vk)
-	viewingPubKey := crypto.CompressPubkey(&vk.PublicKey)
-	cli.SetViewingKey(viewingPrivateKeyEcies, viewingPubKey)
 
-	viewingKeyHex := hex.EncodeToString(viewingPubKey)
+	// get key in ECIES format
+	viewingPrivateKeyECIES := ecies.ImportECDSA(vk)
 
+	// encode public key as bytes
+	viewingPubKeyBytes := crypto.CompressPubkey(&vk.PublicKey)
+
+	// set key pair on the RPC client
+	cli.SetViewingKey(viewingPrivateKeyECIES, viewingPubKeyBytes)
+
+	// sign hex-encoded public key string with the wallet's private key
+	viewingKeyHex := hex.EncodeToString(viewingPubKeyBytes)
 	signature := signViewingKey(viewingKeyHex, wal.PrivateKey())
 
+	// submit the signed public key to the enclave so it can encrypt sensitive responses
 	err = cli.RegisterViewingKey(wal.Address(), signature)
 	if err != nil {
 		panic(err)
--- a/integration/common/viewkey/viewkey.go
+++ b/integration/common/viewkey/viewkey.go
@@ -1,0 +1,60 @@
+package viewkey
+
+import (
+	"crypto/ecdsa"
+	"encoding/hex"
+	"github.com/ethereum/go-ethereum/accounts"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/crypto/ecies"
+	"github.com/obscuronet/go-obscuro/go/enclave/rpcencryptionmanager"
+	"github.com/obscuronet/go-obscuro/go/rpcclientlib"
+	"github.com/obscuronet/go-obscuro/go/wallet"
+)
+
+// GenerateAndRegisterViewingKey take an obscuro client and a wallet, it generates a keypair, simulates signing the key,
+//	configures it on the client and registers the viewing key with the enclave
+func GenerateAndRegisterViewingKey(cli rpcclientlib.Client, wal wallet.Wallet) {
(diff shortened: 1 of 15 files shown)


integration/simulation/network/network.go
```diff
@@ -17,6 +20,26 @@ type Network interface {
...
 	// Return an error in case it cannot start for an expected reason. Otherwise it panics.
-	Create(params *params.SimParams, stats *stats.Stats) ([]ethadapter.EthClient, []rpcclientlib.Client, error)
+	Create(params *params.SimParams, stats *stats.Stats) (*Clients, error)
 	TearDown()
 }
+
+type Clients struct {
+	EthClients     []ethadapter.EthClient         // an eth client per eth node in the network
+	ObscuroClients []rpcclientlib.Client          // an obscuro client per obscuro node in the network
+	WalletClients  map[string]rpcclientlib.Client // an obscuro client per wallet (configured with viewing key where applicable)
```
joeldudley, line 30: @BedrockSquirrel I think this can be a private field?

integration/simulation/network/obscuro_node_utils.go
```diff
@@ -116,7 +144,66 @@ func startStandaloneObscuroNodes(params *params.SimParams, stats *stats.Stats, g
 		}
 	}
 
-	return obscuroClients
+	// round-robin the wallets onto the different obscuro nodes, register them each a viewing key
+	walletClients := make(map[string]rpcclientlib.Client)
+	var i int
+	for _, w := range params.Wallets.SimObsWallets {
+		walletClients[w.Address().String()] = rpcclientlib.NewClient(nodeRPCAddresses[i%len(nodeRPCAddresses)])
```
joeldudley, line 151: @BedrockSquirrel Why are we keying the map with a string, rather than with the wallet itself? More efficient?

integration/simulation/network/obscuro_node_utils.go
```diff
@@ -61,11 +69,31 @@ func startInMemoryObscuroNodes(params *params.SimParams, stats *stats.Stats, gen
...
+
+	walletClients := setupWalletClientsWithoutViewingKeys(params, obscuroClients)
+
+	return obscuroClients, walletClients
+}
+
+// setupWalletClientsWithoutViewingKeys will round-robin wallets onto the existing obscuro clients, no viewing keys registered.
+func setupWalletClientsWithoutViewingKeys(params *params.SimParams, obscuroClients []rpcclientlib.Client) map[string]rpcclientlib.Client {
+	walletClients := make(map[string]rpcclientlib.Client)
+	var i int
```
joeldudley, line 81: @BedrockSquirrel Smart, but I think it deserves an explanatory comment.

joeldudley (dismissed):
Generally looks great, assuming it works. A few comments.

integration/simulation/network/obscuro_node_utils.go
BedrockSquirrel: Tbh I did it because I couldn't remember the Go rules for what could safely be a map key and didn't want any gotchas haha. But also I could imagine being in part of the code that knew the wallet address it wanted to transact from but didn't have easy access to the wallet objects collection.

Can change it if you prefer? (Looks like the wallet struct would be a legit key)

BedrockSquirrel:
I've moved the logic to generate, simulate signing and register a viewing key to a simulation-level util so it can be re-used in the network manager transaction injector.

(I've also put the WalletClients collection back to being public so it can be populated in the struct from anywhere rather than adding a constructor and because I think there will be situations where we want to access it, the lookup by wallet is a convenience).

integration/common/viewkey/viewkey.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/ethereum/go-ethereum/accounts"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/crypto/ecies"
+	"github.com/obscuronet/go-obscuro/go/enclave/rpcencryptionmanager"
+	"github.com/obscuronet/go-obscuro/go/rpcclientlib"
+	"github.com/obscuronet/go-obscuro/go/wallet"
+)
+
+// GenerateAndRegisterViewingKey take an obscuro client and a wallet, it generates a keypair, simulates signing the key,
+//	configures it on the client and registers the viewing key with the enclave
```
otherview, line 16: // GenerateAndRegisterViewingKey takes an obscuro client and a wallet, generates a keypair, simulates signing the key,
//	configures the (which key?) key on the client and registers the viewing key with the enclave

Can this be tweaked a nudge ?

integration/common/viewkey/viewkey.go
```diff
@@ -0,0 +1,61 @@
...
+func GenerateAndRegisterViewingKey(cli rpcclientlib.Client, wal wallet.Wallet) {
+	vk, err := crypto.GenerateKey()
+	if err != nil {
+		panic(err)
+	}
+	viewingPrivateKeyEcies := ecies.ImportECDSA(vk)
```
otherview, line 22: These are pretty high level ops, can they have a few comments ?

integration/simulation/network/obscuro_node_utils.go
```diff
@@ -61,11 +63,34 @@ func startInMemoryObscuroNodes(params *params.SimParams, stats *stats.Stats, gen
...
+	for _, t := range params.Wallets.Tokens {
+		w := t.L2Owner
+		walletClients[w.Address().String()] = obscuroClients[i%len(obscuroClients)]
+		i++
+	}
+	return walletClients
 }
 
-func startStandaloneObscuroNodes(params *params.SimParams, stats *stats.Stats, gethClients []ethadapter.EthClient, enclaveAddresses []string) []rpcclientlib.Client {
+func startStandaloneObscuroNodes(params *params.SimParams, stats *stats.Stats, gethClients []ethadapter.EthClient, enclaveAddresses []string) ([]rpcclientlib.Client, map[string]rpcclientlib.Client) {
```
otherview, line 91: What if the starting an obscuro node and creating RPC Clients / Wallets was decoupled ?

otherview (commented):
Looks good. Personally, I'd start decoupling node setup and connection setup. But it's a matter of taste I guess.

integration/common/viewkey/viewkey.go
BedrockSquirrel: Yeah that was confusing, have added a bunch more context, might be overkill but better to be too helpful than too vague I guess haha.

integration/simulation/network/obscuro_node_utils.go
BedrockSquirrel: I agree in principle but I've just had a go and it's fiddly to make it come out cleanly (like you might need to pass the addresses back as well as the nodes). I think maybe this is a prime spot for a refactor PR (have left a todo).

tudor-malene (commented):
lgtm.
(after a few naming changes we discussed)

tudor-malene (approved):
lgtm

PR #439, opened 2022-07-22
Remove flag that allows disabling viewing keys

BedrockSquirrel:

- for backwards compatibility we were returning the raw result if trying to decrypt a response with no viewing key set
- now that all simulations can use RPC client have viewing key encryption it's not required
- we should error instead of returning raw response to avoid weird behaviour (like metamask trying to interpret encrypted balance as actual balance)

*functional*
- return error instead of unencrypted response when attempting to decrypt without a VK available
- enable viewing keys in all simulations that use the RPC client
- provide test validation with a matrix of rpc clients (one for each wallet-node combination) for verifying the transactions


tudor-malene (approved):
lgtm. 
Once we deploy this to the testnet, and try it out with metamask, we might have to make a change to the wallet extension as well. ( to return an error, or 0, or whatever makes metamask not crash)

BedrockSquirrel:
>

integration/simulation/network/network.go
```diff
@@ -31,9 +33,10 @@ type RPCHandles struct {
 	// an obscuro client per obscuro node in the network (used for things like validation rather than transactions on behalf of sim accounts)
 	ObscuroClients []rpcclientlib.Client
 
-	// an RPC client per wallet, with a viewing key setup (on the client and registered on its corresponding host enclave),
+	// an RPC client per node per wallet, with a viewing key setup (on the client and registered on its corresponding host enclave),
```
joeldudley, line 36: @BedrockSquirrel Can we add a TODO here that in the future, we can hopefully drop back to a single RPC client per node?

integration/simulation/network/network.go
```diff
@@ -31,9 +33,10 @@ type RPCHandles struct {
...
 	//	to mimic user acc interaction via a wallet extension
-	VirtualWalletExtensionClients map[string]rpcclientlib.Client // an obscuro client per wallet (configured with viewing key where applicable)
+	// map of address hex string -> slice of clients (indexed by node)
```
joeldudley, line 38: @BedrockSquirrel Not sure what this means. Can you clarify? The keys are the hex strings of whose/which addresses? Which clients are in the value slice? Should it be:

`// map of owner addresses to RPC clients for that owner (one RPC client per L2 node)`

integration/simulation/network/network.go
```diff
@@ -44,8 +47,15 @@ func (n *RPCHandles) RndObscuroClient() rpcclientlib.Client {
 	return n.ObscuroClients[rand.Intn(len(n.ObscuroClients))] //nolint:gosec
 }
 
-// ObscuroWalletClient fetches client for given wallet if it exists
-func (n *RPCHandles) ObscuroWalletClient(wallet wallet.Wallet) rpcclientlib.Client {
+// ObscuroWalletRndClient fetches a client for a given wallet connected to a random node
```
joeldudley, line 50: @BedrockSquirrel Maybe clearer as:

`// ObscuroWalletRndClient fetches an RPC client connected to a random L2 node for a given wallet`

? Sorry!

integration/simulation/network/obscuro_node_utils.go
```diff
@@ -148,35 +161,33 @@ func startStandaloneObscuroNodes(params *params.SimParams, stats *stats.Stats, g
...
 		w := t.L2Owner
-		client, err := rpcclientlib.NewViewingKeyNetworkClient(nodeRPCAddresses[i%len(nodeRPCAddresses)])
+		walletClients[w.Address().String()] = createAuthenticatedClientsForWallet(nodeRPCAddresses, w)
+	}
+
+	return obscuroClients, walletClients
+}
+
+// createAuthenticatedClientsForWallet takes a wallet and sets up a client for it for every node
+func createAuthenticatedClientsForWallet(nodeRPCAddresses []string, wal wallet.Wallet) []rpcclientlib.Client {
```
joeldudley, line 177: @BedrockSquirrel I found the naming of this confusing. It's really just `createInMemoryClientsForWallet`, but for standalone nodes, right? But the naming makes them sound different.

Can you place the two methods next to each other in the file, and give them more similar nodes - e.g. `createInMemoryClientsForWallet` and `createStandaloneClientsForWallet` (or `createRPCClientsForWallet`).

tools/networkmanager/inject_txs.go
```diff
@@ -183,10 +185,10 @@ func checkDepositsSuccessful(txInjector *simulation.TransactionInjector, l1Clien
 	}
 }
 
-func checkL2TxsSuccessful(l2Client rpcclientlib.Client, txInjector *simulation.TransactionInjector) {
+func checkL2TxsSuccessful(rpcHandles *network.RPCHandles, txInjector *simulation.TransactionInjector) {
```
joeldudley, line 188: @BedrockSquirrel Can you add a todo to investigate why we do some validation here, then repeat it in `validate_chain.go`?

integration/simulation/utils.go
```diff
@@ -139,18 +120,18 @@ func balance(client rpcclientlib.Client, address gethcommon.Address, l2ContractA
...
-		panic(fmt.Errorf("could not decode ERC20 balance response for node. Response was %s. Cause: %w", encryptedResponse, err))
+	if !strings.HasPrefix(response, "0x") {
+		panic(fmt.Errorf("expected hex formatted balance string but was: %s", response))
 	}
-	r := new(big.Int)
-	r = r.SetBytes(bytes)
-	return r.Uint64()
+
+	b := new(big.Int)
+	b.SetString(response[2:], 16)
```
joeldudley, line 133: @BedrockSquirrel Can you add a comment explaining this op, please?

integration/simulation/validate_chain.go
```diff
@@ -285,33 +278,46 @@ func checkBlockchainOfObscuroNode(
 }
 
 // FindNotIncludedL2Txs returns the number of transfers and withdrawals that were injected but are not present in the L2 blockchain.
-func FindNotIncludedL2Txs(l2Client rpcclientlib.Client, txInjector *TransactionInjector) (int, int) {
+func FindNotIncludedL2Txs(nodeIdx int, rpcHandles *network.RPCHandles, txInjector *TransactionInjector) (int, int) {
```
joeldudley, line 281: @BedrockSquirrel I think it makes sense to pass in the client, as we did before, rather than selecting the client (twice) in this method?

tools/networkmanager/inject_txs.go
BedrockSquirrel: This code is in the network manager tool (it's not the sim transaction injector code).

It re-uses a bit of the validate_chain logic from the sims to check the transactions behaved as successful but I don't think `validate_chain.go` gets called in full afterwards.

joeldudley: Ah thanks. Got confused between the two inject_txs.go's.

joeldudley (approved):

PR #440, opened 2022-07-25
Refactor contract deployer to do single contract deployments

BedrockSquirrel:

- we want to deploy L1 and L2 contracts as part of testnet setup, this is to pave the way for that
- I think most flexible way to be able to use this tool for that stuff is to have it be called once per contract it is deploying rather than having preset hardcoded bundles of contracts to deploy

*refactor*
- add 'contract' param that for now accepts 'MGMT' or 'ERC20' as a value, and then deploys the associated bytecode
- output deployed address as final line of output (to replace the json blob that was getting output before)

- `testnet-deploy-contracts.sh` for change in how it is called (and how the output is read)
- `contract_deployer.go` for whether refactor is sensible (it prepares us to be able to add an --L1Deployment boolean flag that will build the client etc. differently)

### DIFF
--- a/testnet/testnet-deploy-contracts.sh
+++ b/testnet/testnet-deploy-contracts.sh
@@ -52,20 +52,30 @@ fi
 # deploy contracts to the geth network
 echo "Deploying contracts to the geth network..."
 docker network create --driver bridge node_network || true
+
+# deploy Obscuro management contract
 docker run --name=contractdeployer \
     --network=node_network \
     --entrypoint /home/go-obscuro/tools/contractdeployer/main/main \
      testnetobscuronet.azurecr.io/obscuronet/obscuro_contractdeployer:latest \
-    --l1NodeHost=${l1host} \
-    --l1NodePort=${l1port} \
+    --nodeHost=${l1host} \
+    --nodePort=${l1port} \
+    --contract=MGMT \
     --privateKey=${pkstring}
-
-# storing the contract addresses to the .env file
-log_output=$(docker logs --tail 1 contractdeployer)
-json_output=$(echo ${log_output} | awk -F"[{}]" '{print "{"$2"}"}')
-mgmtContractAddr=$(echo "${json_output}"  | jq .MgmtContractAddr)
-erc20ContractAddr=$(echo "${json_output}" | jq .ERC20ContractAddr)
-
+# storing the contract address to the .env file (note: this first contract creates/overwrites the .env file)
+mgmtContractAddr=$(docker logs --tail 1 contractdeployer)
 echo "MGMTCONTRACTADDR=${mgmtContractAddr}" > "${testnet_path}/.env"
+
+# deploy ERC20 contract
+docker run --name=contractdeployer \
+    --network=node_network \
+    --entrypoint /home/go-obscuro/tools/contractdeployer/main/main \
+     testnetobscuronet.azurecr.io/obscuronet/obscuro_contractdeployer:latest \
+    --nodeHost=${l1host} \
+    --nodePort=${l1port} \
+    --contract=ERC20 \
+    --privateKey=${pkstring}
+# storing the contract address to the .env file
+mgmtContractAddr=$(docker logs --tail 1 contractdeployer)
 echo "ERC20CONTRACTADDR=${erc20ContractAddr}" >> "${testnet_path}/.env"
 
--- a/testnet/testnet-deploy-contracts.sh
+++ b/testnet/testnet-deploy-contracts.sh
@@ -54,7 +54,7 @@ echo "Deploying contracts to the geth network..."
 docker network create --driver bridge node_network || true
 
 # deploy Obscuro management contract
-docker run --name=contractdeployer \
+docker run --name=mgmtcontractdeployer \
     --network=node_network \
     --entrypoint /home/go-obscuro/tools/contractdeployer/main/main \
      testnetobscuronet.azurecr.io/obscuronet/obscuro_contractdeployer:latest \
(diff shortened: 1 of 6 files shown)


testnet/testnet-deploy-contracts.sh
```diff
@@ -52,20 +52,30 @@ fi
...
+# deploy Obscuro management contract
 docker run --name=contractdeployer \
     --network=node_network \
     --entrypoint /home/go-obscuro/tools/contractdeployer/main/main \
      testnetobscuronet.azurecr.io/obscuronet/obscuro_contractdeployer:latest \
-    --l1NodeHost=${l1host} \
-    --l1NodePort=${l1port} \
+    --nodeHost=${l1host} \
+    --nodePort=${l1port} \
+    --contract=MGMT \
```
otherview, line 63: Nit: Can this be wrapped in double quotes "MGMT" ? Make is clear that it's a hardcoded value.

testnet/testnet-deploy-contracts.sh
```diff
@@ -52,20 +52,30 @@ fi
...
 echo "MGMTCONTRACTADDR=${mgmtContractAddr}" > "${testnet_path}/.env"
+
+# deploy ERC20 contract
+docker run --name=contractdeployer \
+    --network=node_network \
+    --entrypoint /home/go-obscuro/tools/contractdeployer/main/main \
+     testnetobscuronet.azurecr.io/obscuronet/obscuro_contractdeployer:latest \
+    --nodeHost=${l1host} \
+    --nodePort=${l1port} \
+    --contract=ERC20 \
```
otherview, line 76: Nit: Similar here. `"ERC20"`

testnet/testnet-deploy-contracts.sh
```diff
@@ -52,20 +52,30 @@ fi
...
+    --nodeHost=${l1host} \
+    --nodePort=${l1port} \
+    --contract=MGMT \
     --privateKey=${pkstring}
-
-# storing the contract addresses to the .env file
-log_output=$(docker logs --tail 1 contractdeployer)
-json_output=$(echo ${log_output} | awk -F"[{}]" '{print "{"$2"}"}')
-mgmtContractAddr=$(echo "${json_output}"  | jq .MgmtContractAddr)
-erc20ContractAddr=$(echo "${json_output}" | jq .ERC20ContractAddr)
```
otherview, line 67: Did you remove the jq requirment ? Awesome :)

tools/contractdeployer/main/main.go
```diff
@@ -8,7 +8,12 @@ import (
 func main() {
 	log.SetLogLevel(log.DisabledLevel)
 	config := contractdeployer.ParseConfig()
-	deployer := contractdeployer.NewContractDeployer(config)
+	deployer, err := contractdeployer.NewContractDeployer(config)
+	if err != nil {
+		// todo: why is this log level stuff setup in this way (why not print here or use logs everywhere)
```
otherview, line 13: Good question.

Logs are disabled so that the only output is the contract deployer address.
If there are logs, then it complicated the retrieval of the contract address from the output logs.

Might be worth improving, for sure.

tools/contractdeployer/cli_flags.go
```diff
@@ -2,18 +2,18 @@ package contractdeployer
...
+	nodeHostUsage = "The host on which to connect the RPC client"
 
-	l1NodePortName  = "l1NodePort"
-	l1NodePortUsage = "The port on which to connect to the Ethereum client"
+	nodePortName  = "nodePort"
+	nodePortUsage = "The port on which to connect the RPC client"
 
-	mgmtContractPathName  = "mgmtContractPath"
-	mgmtContractPathUsage = "The path of the management contract path"
+	contractName  = "contract"
```
otherview, line 11: Sugestion: Would change this flag to `contractID` or `contractName`, maybe something else ? Just to make clear that it's and identifier of sorts ? WDYT ?

otherview (approved):
Minor stuff that can be swapped around, but lgtm.

testnet/testnet-deploy-contracts.sh
BedrockSquirrel: Cool, yeah will give it a go. If those quotes get passed into the flag args string then I'm not so sure though, don't want to have `"\"MGMT\""` as the Go constant ideally haha

tools/contractdeployer/cli_flags.go
BedrockSquirrel: Nice, yeah good idea, makes it clearer what the argument expects (and it will be fun to call this variable `contractNameName` haha

BedrockSquirrel:
Tested the script locally, seems to work nicely after the last commits:
Screenshot 2022-07-25 at 12 25 48

tools/contractdeployer/README.md
```diff
@@ -0,0 +1,18 @@
...
+Contract is a string value to select the contract bytecode to deploy, currently only ERC20 and MGMT are supported
+(an ERC20 contract and the obscuro management contract, respectively)
+
+* Example arguments to deploy an L2 contract:
+
+  `--nodeHost=<x> --nodePort=<x> --privateKey=<x> --contract=ERC20`
+
+* Example arguments to deploy an L1 contract:
+
+  `--nodeHost=<x> --nodePort=<x> --privateKey=<x> --contract=MGMT --l1Deployment`
```
otherview, line 18:
```suggestion
  `--nodeHost=<x> --nodePort=<x> --privateKey=<x> --contract=MGMT`
```

otherview (approved):

PR #442, opened 2022-07-26
re-ingests blocks

otherview:

- https://github.com/obscuronet/obscuro-internal/issues/659

- Functional
- After restarting the subscriber, reingests any missing blocks

### DIFF
--- a/go/ethadapter/geth_rpc_client.go
+++ b/go/ethadapter/geth_rpc_client.go
@@ -143,7 +143,7 @@ func (e *gethRPCClient) Nonce(account gethcommon.Address) (uint64, error) {
 }
 
 func (e *gethRPCClient) BlockListener() (chan *types.Header, ethereum.Subscription) {
-	ch := make(chan *types.Header, 1)
+	ch := make(chan *types.Header, 100)
 	sub, err := e.client.SubscribeNewHead(context.Background(), ch)
 	if err != nil {
 		log.Panic("could not subscribe for new head blocks. Cause: %s", err)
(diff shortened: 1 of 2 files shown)


go/host/host.go
```diff
@@ -667,9 +668,36 @@ func (a *Node) monitorBlocks() {
...
-			// todo this is a very simple way of reconnecting the node, it might need catching up logic
+			// it's fine to immediately restart the listener, any incoming blocks will be on hold in the queue
 			listener, subs = a.ethClient.BlockListener()
 
+			// catch up any missed blocks
+			var lastBlkNumber *big.Int
+			// get the latest known blk
+			lastBlk, err := a.ethClient.BlockByNumber(lastBlkNumber)
+			if err != nil {
+				log.Panic("%s", err)
```
BedrockSquirrel, line 679: Bit of context on this might be good since we're panicking. "unable to catch up on missed blocks, last known block not found" or something, I know we'll get the line number but might help to point investigator in the right direction.

go/ethadapter/geth_rpc_client.go
```diff
@@ -143,7 +143,7 @@ func (e *gethRPCClient) Nonce(account gethcommon.Address) (uint64, error) {
 }
 
 func (e *gethRPCClient) BlockListener() (chan *types.Header, ethereum.Subscription) {
-	ch := make(chan *types.Header, 1)
+	ch := make(chan *types.Header, 100)
```
BedrockSquirrel, line 146: A comment on this might be nice to say what the channel is used for (source and destination) and why it has a buffer.

Thinking along the lines of `// this channel holds blocks that have been received from the geth network but not yet processed by the enclave, the buffer provides resilience in case of intermittent RPC or processing issues`

BedrockSquirrel (commented):
LGTM, couple of minor requests

go/host/host.go
otherview: Good point 👍

go/ethadapter/geth_rpc_client.go
otherview: That's great 👌

BedrockSquirrel (approved):

PR #445, opened 2022-07-27
Debugging l1 reconnect

otherview:

- Same issue with 
`Jul 27 05:13:10.250 ERR Restarting L1 block monitoring. Errored with: read tcp 172.18.0.2:42280->20.26.38.211:9000: i/o timeout` follows a `Jul 27 05:13:10.295 WRN >   Agg1: Could not process block received via RPC. Cause: did not ingest block b_15886852837258715195. Cause: Block parent not stored.`

- Debugging changes


go/host/host.go
```diff
@@ -716,6 +699,47 @@ func (a *Node) monitorBlocks() {
 	subs.Unsubscribe()
 }
 
+func (a *Node) catchupMissedBlocks(lastKnownBlkHash gethcommon.Hash) {
+	var lastBlkNumber *big.Int
+	var reingestBlocks []*types.Block
+
+	// get the tip block
```
BedrockSquirrel, line 706: 'tip block' isn't super clear, I'd say 'last processed block' or 'last ingested block' maybe?

go/host/host.go
```diff
@@ -716,6 +699,47 @@ func (a *Node) monitorBlocks() {
...
+		log.Panic("catching up on missed blocks, unable to fetch tip block - reason: %s", err)
+	}
+
+	if lastBlk.Hash().Hex() == lastKnownBlkHash.Hex() {
+		// no new blocks have been issued
+		return
+	}
+	reingestBlocks = append(reingestBlocks, lastBlk)
+
+	// iterate from the tip (last block produced) to the last block ingested by the node
```
BedrockSquirrel, line 718: perhaps 'processed' or 'ingested' instead of 'produced'?

go/host/host.go
```diff
@@ -716,6 +699,47 @@ func (a *Node) monitorBlocks() {
...
+	for lastBlk.Hash().Hex() != lastKnownBlkHash.Hex() {
```
BedrockSquirrel, line 719: I think one of these should be called latestBlock maybe? I guess one of them is the latest block we have received from the L1 but it's not clear to me which it is. Some renaming or comment to clarify perhaps (sorry, Friday afternoon Matt is not the sharpest, it's probably not as confusing as I'm making out)

go/host/host.go
```diff
@@ -716,6 +699,47 @@ func (a *Node) monitorBlocks() {
...
+		blockParent, err := a.ethClient.BlockByHash(lastBlk.ParentHash())
+		if err != nil {
+			log.Panic("catching up on missed blocks, could not fetch block's parent with hash %s. Cause: %s", lastBlk.ParentHash(), err)
+		}
+
+		reingestBlocks = append(reingestBlocks, blockParent)
+		lastBlk = blockParent
+	}
+
+	// make to have the last ingested block available
```
BedrockSquirrel, line 729: this sentence doesn't quite make sense, think maybe words have gone missing after 'make'?

BedrockSquirrel (commented):
i think it seems sensible but I got quite confused - maybe just some tweaking of the names/comments to make it clear what the process is that it's going through

go/host/host.go
otherview: I was trying to avoid that, because it gets confusing. 

The last canonical block is usually called the tip of the block chain. Changing to that, wdyt ?

otherview:
```suggestion
	// get the blockchain tip block
```

go/host/host.go
otherview:
```suggestion
	// iterate from the blockchain tip to the last block sucessfully ingested by the node
```

Wdyt @BedrockSquirrel ?

go/host/host.go
otherview:
```suggestion
	// make sure to have the last ingested block available for ingestion (because we always ingest ( blk, blk_parent)
```

BedrockSquirrel (approved):
LGTM

PR #446, opened 2022-07-27
ObscuroScan can retrieve specific rollups and their blocks, not just the latest

joeldudley:

Previously, ObscuroScan only showed the latest block and rollup. This wasn't useful, and changed too often to be inspected.

Functional.

- ObscuroScan now allows you to retrieve a rollup and its associated block by number, rather than just showing the latest
- Added RPC methods needed to support new functionality
- Killed off some old methods on the host that were created to support RPC functionality, but weren't used


BedrockSquirrel (approved):
Nice, yeah looks good to me - I can't test against testnet because of the missing rpc endpoint but the errors showed up nicely :)

If this does become something we use going forward it should attempt to decrypt the transaction blob as part of the rollup request and display that too on the block_monitor page imo but that can wait :)

joeldudley:
@BedrockSquirrel You can test easily against a the full-network sim network if you bump up the time and use port `127.0.0.1:37400`.

joeldudley:
Agreed re: decrypting tx blob on that page, just don't want to do too much work here if it's getting paused.

PR #447, opened 2022-07-27
adding enclave debug mode

otherview:

- Part of https://github.com/obscuronet/obscuro-internal/issues/596
- Adds an debugger mode to the local testnet
- Some infos in the scripts were incorrect
- Reduces the docker images with the .dockerignore

- Adds a new docker image, docker compose file and minor changes to the scripts

### DIFF
--- a/dockerfiles/enclave.debug.Dockerfile
+++ b/dockerfiles/enclave.debug.Dockerfile
@@ -1,0 +1,26 @@
+FROM golang:1.17-alpine
+
+# Debug image which installs and runs delve
+#
+# on the container:
+#   /data                    contains working files for the enclave
+#   /home/obscuro/go-obscuro contains the src
+#
+# Note: ego uses a virtual file system mount to map data directory to /data inside the enclave,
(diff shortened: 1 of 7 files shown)


dockerfiles/enclave.debug.Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM golang:1.17-alpine
+
+# Debug image which installs and runs delve
+#
+# on the container:
+#   /data                    contains working files for the enclave
+#   /home/obscuro/go-obscuro contains the src
+#
+# Note: ego uses a virtual file system mount to map data directory to /data inside the enclave,
+#   for this non-ego build I'm using /data as the data dir to preserve /data folder in paths inside enclave
```
joeldudley, line 10: Maybe clarify further up that this dockerfile runs the enclave outside of EGo?

joeldudley (approved):

dockerfiles/enclave.debug.Dockerfile
otherview: yep 👍

PR #449, opened 2022-07-28
Block explorer design

joeldudley:

There is no formal design for the block explorer. We should have a document that captures key decisions made in the past (e.g. fork vs build our own).

Functional.

- Adds design doc


CaisManai (approved):
TODOs:

**Other Metadata** - block number and status of transaction. Once It's left the mempool it can still have several types of status e.g. success and failed

**Any contracts that have been deployed** - Etherscan can display the bytecode of every contract, and the source code too where confirmed by providing a Github repo, but we should drop this requirement from TN1.

**Decryption capabilities** It sounds like the capability is already built, so yes, let's integrate it into the block explorer

Let's discuss the private explorer over a call

PR #450, opened 2022-07-28
removes setting both poststate and status fields

otherview:
From my understanding, there's either postState (root) set as the state Hash or Status field set as true/false.

That's from and interpretation from https://github.com/ethereum/EIPs/blob/master/EIPS/eip-658.md: 
`For blocks where block.number >= BYZANTIUM_FORK_BLKNUM, the intermediate state root is replaced by a status code, 0 indicating failure (due to any operation that can cause the transaction or top-level call to revert) and 1 indicating success.`

And it seems like that is what hardhat thinks should happen.
```
  Formatter.prototype.receipt = function (value) {
        var result = Formatter.check(this.formats.receipt, value);
        // RSK incorrectly implemented EIP-658, so we munge things a bit here for it
        if (result.root != null) {
            if (result.root.length <= 4) {
                // Could be 0x00, 0x0, 0x01 or 0x1
                var value_1 = bignumber_1.BigNumber.from(result.root).toNumber();
                if (value_1 === 0 || value_1 === 1) {
                    // Make sure if both are specified, they match
                    if (result.status != null && (result.status !== value_1)) {
                        logger.throwArgumentError("alt-root-status/status mismatch", "value", { root: result.root, status: result.status });
                    }
                    result.status = value_1;
                    delete result.root;
                }
                else {
                    logger.throwArgumentError("invalid alt-root-status", "value.root", result.root);
                }
            }
            else if (result.root.length !== 66) {
                // Must be a valid bytes32
                logger.throwArgumentError("invalid root hash", "value.root", result.root);
            }
        }
        if (result.status != null) {
            result.byzantium = true;
        }
        return result;
    };
```

### DIFF
--- a/go/enclave/enclave.go
+++ b/go/enclave/enclave.go
@@ -371,11 +371,6 @@ func (e *enclaveImpl) GetTransactionReceipt(encryptedParams common.EncryptedPara
 		return nil, fmt.Errorf("could not retrieve transaction receipt in eth_getTransactionReceipt request. Cause: %w", err)
 	}
 
-	// If the poststate is set, the status should be set to successful.
-	if len(txReceipt.PostState) == 32 {
-		txReceipt.Status = types.ReceiptStatusSuccessful
-	}
-
 	encryptedTxReceipt, err := e.rpcEncryptionManager.EncryptTxReceiptWithViewingKey(viewingKeyAddress, txReceipt)
 	if err != nil {
 		return nil, fmt.Errorf("enclave could not respond securely to eth_getTransactionReceipt request. Cause: %w", err)


joeldudley:
@otherview Have you tested this via Remix?

otherview:
> @otherview Have you tested this via Remix?

I've tested it with hardhat, by running the below. 
Happy to test it with remix, is there a specific test that I can run ?


```js
describe("DeployTokens", function () {
    const mintedAmount = ethers.utils.parseEther("1234567891");
    let wethContract;
    let pedroTokenContract;
    let obxTokenContract;
    let USDCTokenContract;
    beforeEach(async function () {
        const [owner] = await ethers.getSigners();
        let signValue = ""
        await axios
            .get('http://127.0.0.1:3001/generateviewingkey')
            .then(res => {
                signValue = res.data;
            })
            .catch(error => {
                console.error(error);
            });

        let signed_msg = await owner.signMessage("vk" + signValue);

        await axios
            .post('http://127.0.0.1:3001/submitviewingkey/',
                JSON.stringify({"address": owner.address.toString(), "signature": signed_msg})
                , {
                headers: {
                    'Content-Type': 'application/json'
                }
            })
            .catch(error => {
                console.error(error);
            });

        const Weth = await ethers.getContractFactory("WETHToken");
        wethContract = await Weth.deploy(mintedAmount, {gasLimit: 50000000000});

        const PedroToken = await ethers.getContractFactory("PedroToken");
        pedroTokenContract = await PedroToken.deploy(mintedAmount, {gasLimit: 50000000000});

        const Obx = await ethers.getContractFactory("OBXToken");
        obxTokenContract = await Obx.deploy(mintedAmount, {gasLimit: 50000000000});

        const USDC = await ethers.getContractFactory("USDC");
        USDCTokenContract = await USDC.deploy(mintedAmount, {gasLimit: 50000000000});

        console.log("Owner Address of tokens: ");
        console.log(owner.address);
    });

    describe("Deployment", function () {
        it("Should deploy the contracts correctly", async function () {
            console.log(await wethContract.symbol())
            expect(await wethContract.symbol()).to.equal("WETH");
            expect(await wethContract.totalSupply()).to.equal(mintedAmount);
            console.log("weth address: ")
            console.log(wethContract.address);

            expect(await pedroTokenContract.symbol()).to.equal("PTK");
            expect(await pedroTokenContract.totalSupply()).to.equal(mintedAmount);
            console.log("PedroToken address: ")
            console.log(pedroTokenContract.address);

            expect(await obxTokenContract.symbol()).to.equal("OBX");
            expect(await obxTokenContract.totalSupply()).to.equal(mintedAmount);
            console.log("obxTokenContract address: ")
            console.log(obxTokenContract.address);

            expect(await USDCTokenContract.symbol()).to.equal("USDC");
            expect(await USDCTokenContract.totalSupply()).to.equal(mintedAmount);
            console.log("USDCTokenContract address: ")
            console.log(USDCTokenContract.address);

            console.log("done")
        });

    });
});

``

joeldudley:
@otherview If you can just try deploying a standard contract via Remix (e.g. the default storage one). I only ask because we originally made this change because it was breaking Remix.

otherview:
@joeldudley works well :)
image

joeldudley (approved):
Awesome

PR #453, opened 2022-07-28
ObscuroScan allows rollups to be searched by transaction hash

joeldudley:

Previously, rollups could only be queried by number. But this wasn't user-friendly, as no-one knows which rollup their transaction is in.

Functional.

- Adds the ability in ObscuroScan to query for the rollup containing a given transaction hash


BedrockSquirrel (approved):
Tested locally with network sim - LGTM

PR #455, opened 2022-07-28
Auto-decrypt encrypted tx blob in ObscuroScan

joeldudley:

Currently, in ObscuroScan, the encrypted transaction blob for a rollup has to be pasted into a separate window. This is bad UX - we should be decrypting it for them automatically.

Functional.

- Auto-decrypts the tx blob for any fetched rollup
- (Minor) Jump down to show the results after fetching
- (Minor) Uses h5 headers instead of text so it looks better


tools/obscuroscan/static/index.js
```diff
@@ -35,8 +34,6 @@ const initialize = () => {
     }, 1000);
 
     document.getElementById(idFormGetRollup).addEventListener(typeSubmit, async (event) => {
-        event.preventDefault();
```
joeldudley, line 38: Removed because we actually want the form to perform its action (jumping down the page) now.

BedrockSquirrel (approved):
Very nice, tested locally and all seems to be working as expected.

PR #458, opened 2022-07-29
Clear-up wallet extension docs

joeldudley:

There were a number of issues with the wallet extension docs:

* We did not document the flags for the wallet extension
* We defaulted the wallet extension to point to testnet, which is a bad silent default
* The "wallet extension" and "handling sensitive info" docs were under the "Testnet" header, but were not testnet specific

Functional.

* Documents the flags
* Defaults to localhost, making choice of testnet explicit
* Moves the "wallet extension" and "handling sensitive info" sections


docs/wallet-extension/wallet-extension.md
```diff
@@ -27,15 +27,27 @@ unencrypted. If the data is not particularly sensitive, it can also be run in an
...
+   * `nodeHost` (default: `127.0.0.1`): The host machine for the Obscuro node
+   * `nodePortHTTP` (default: `13000`): The Obscuro node's HTTP RPC port
+   * `nodePortWS` (default: `13001`): The Obscuro node's websockets RPC port
+   * `logPath` (default: `wallet_extension_logs.txt`): The path for the wallet extension's logs
+
+   The wallet extension is now listening on the specified host and port.
+
+3. Sign in to MetaMask.
 
 4. In MetaMask, add a new custom network. Point the network at the wallet extension by using `http://127.0.0.1:3000/` as
```
moraygrieve, line 52: Should this be stated as <nodeHost> to make it explicit it needs to match nodeHost given on startup?

Same is true for viewing keys in section 5 under usage

tools/walletextension/main/cli.go
```diff
@@ -14,7 +14,7 @@ const (
 	walletExtensionPortUsage   = "The port on which to serve the wallet extension"
 
 	nodeHostName    = "nodeHost"
-	nodeHostDefault = "testnet.obscu.ro"
```
moraygrieve, line 17: Is the URL for testnet in any of the other readmes under docs? Might be worth adding the testnet URL into the Usage section of wallet-extension.md

joeldudley: That's covered under https://docs.obscu.ro/testnet/essentials.html. One of the goals of the PR is to write a network-agnostic wallet extension docs, which isn't just about connecting to testnet.

docs/wallet-extension/wallet-extension.md
joeldudley: Good catch. The host should be always be localhost (it's where the wallet extension is hosted, not the node), but the port may vary.

moraygrieve: ah yes - of course!

moraygrieve (approved):
LGTM

PR #460, opened 2022-07-29
Waits for host when starting wallet extension.

joeldudley:

See https://github.com/obscuronet/obscuro-internal/issues/665.

Previously, the wallet extension would immediately start taking requests and trying to send them to the host. These requests would eventually lead to timeouts, but it wasn't always clear whether it was a genuine timeout, or the host just couldn't be reached.

Functional.

- Wallet extension pauses at start-up for 30 seconds to try and connect to host, and exits if it cannot connect

### DIFF
--- a/tools/walletextension/main/main.go
+++ b/tools/walletextension/main/main.go
@@ -16,6 +16,8 @@ const (
 
 func main() {
 	config := parseCLIArgs()
+	jsonConfig, _ := json.MarshalIndent(config, "", "  ")
+	fmt.Printf("Wallet extension starting with following config: \n%s\n", string(jsonConfig))
 
 	// We wait thirty seconds for a connection to the node. If we cannot establish one, we exit the program.
 	fmt.Printf("Waiting up to thirty seconds for connection to host at %s...\n", config.NodeRPCHTTPAddress)
@@ -31,7 +33,7 @@ func main() {
 
 		counter--
 		if counter <= 0 {
-			fmt.Printf("Could not establish connection to host at %s. Exiting.", config.NodeRPCHTTPAddress)
+			fmt.Printf("Exiting. Could not establish connection to host at %s. Cause: %s\n", config.NodeRPCHTTPAddress, err)
 			return
 		}
 		time.Sleep(time.Second)
@@ -41,9 +43,7 @@ func main() {
 	defer walletExtension.Shutdown()
 	walletExtensionAddr := fmt.Sprintf("%s:%d", localhost, config.WalletExtensionPort)
 	go walletExtension.Serve(walletExtensionAddr)
-	s, _ := json.MarshalIndent(config, "", "  ")
-	fmt.Printf("Wallet extension started with following config: \n%s\n\n", string(s))
-	fmt.Printf("💡 Visit http://%s/viewingkeys/ to generate an ephemeral viewing key.\n", walletExtensionAddr)
+	fmt.Printf("💡 Wallet extension started - visit http://%s/viewingkeys/ to generate an ephemeral viewing key.\n", walletExtensionAddr)
 
 	select {}
 }
--- a/tools/walletextension/main/main.go
+++ b/tools/walletextension/main/main.go
@@ -3,25 +3,47 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"net"
+	"time"
 
 	"github.com/obscuronet/go-obscuro/tools/walletextension"
 )
 
 const (
+	tcp       = "tcp"
 	localhost = "127.0.0.1"
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


tools/walletextension/main/main.go
```diff
@@ -3,25 +3,47 @@ package main
...
+		if conn != nil {
+			conn.Close()
+		}
+		if err == nil {
+			break
+		}
+
+		counter--
+		if counter <= 0 {
+			fmt.Printf("Could not establish connection to host at %s. Exiting.", config.NodeRPCHTTPAddress)
```
moraygrieve, line 34: Need a newline character

tools/walletextension/main/main.go
```diff
@@ -3,25 +3,47 @@ package main
...
 	localhost = "127.0.0.1"
 )
 
 func main() {
 	config := parseCLIArgs()
+
+	// We wait thirty seconds for a connection to the node. If we cannot establish one, we exit the program.
+	fmt.Printf("Waiting up to thirty seconds for connection to host at %s...\n", config.NodeRPCHTTPAddress)
+	counter := 30
+	for {
```
moraygrieve, line 23: This hides other issues now e.g. if the port is in use it will just spin for 30 seconds and then exit as below;

```
Waiting up to thirty seconds for connection to host at 127.0.0.1:13000...
Could not establish connection to host at 127.0.0.1:13000. Exiting
```

... with no indication of what the problem would be. If the port was in use previously you would have seen the error as below;

```
Morays-MacBook-Pro:main moraygrieve$ ./wallet_extension 
Wallet extension config: 
{
  "WalletExtensionPort": 3000,
  "NodeRPCHTTPAddress": "127.0.0.1:13000",
  "NodeRPCWebsocketAddress": "127.0.0.1:13001",
  "LogPath": "wallet_extension_logs.txt"
}
Wallet extension started.
💡 Visit http://127.0.0.1:3000/viewingkeys/ to generate an ephemeral viewing key.
panic: listen tcp 127.0.0.1:3000: bind: address already in use
```

tools/walletextension/main/main.go
```diff
@@ -3,25 +3,47 @@ package main
...
 	walletExtension := walletextension.NewWalletExtension(config)
 	defer walletExtension.Shutdown()
-
 	walletExtensionAddr := fmt.Sprintf("%s:%d", localhost, config.WalletExtensionPort)
 	go walletExtension.Serve(walletExtensionAddr)
 	s, _ := json.MarshalIndent(config, "", "  ")
-	fmt.Printf("Wallet extension config: \n%s", string(s))
-	fmt.Println()
-	fmt.Printf("Wallet extension started.\n💡 Visit http://%s/viewingkeys/ to generate an ephemeral viewing key.\n", walletExtensionAddr)
+	fmt.Printf("Wallet extension started with following config: \n%s\n\n", string(s))
```
moraygrieve, line 45: You might want to dump out the config it is _trying_ to start with, e.g. if it can't connect at least you can see what config it was trying to connect with

moraygrieve (changes requested):
See comments

tools/walletextension/main/main.go
joeldudley: If you give it a spin, don't you automatically get a new line added on program return? I did on my machine.

tools/walletextension/main/main.go
joeldudley: Those are two separate errors. One is the host being unavailable, the other is the local port you want to bind to serve up the wallet extension being unavailable. I guess my change means that you don't detect the latter error, but only because you're detecting the former. Are you saying we should add an extra check here to see if the local port is also free?

tools/walletextension/main/main.go
joeldudley: Yeah, good point.

tools/walletextension/main/main.go
moraygrieve: I didn't on mine;

```
Waiting up to thirty seconds for connection to host at 127.0.0.1:13000...
Could not establish connection to host at 127.0.0.1:13000. Exiting.Morays-MacBook-Pro:main moraygrieve$ ls -l
```

joeldudley: Ah weird, will add, thanks.

tools/walletextension/main/main.go
moraygrieve: Just that it hides the error response in the implementation ... i.e. before you would have seen the port was in use and so would fix the issue quickly, whereas now that information is lost to the user and they don't immediately realise the underlying cause. Yes you can't establish connection to the server, but the issue is not on the server side but on your side.

moraygrieve (approved):

PR #462, opened 2022-07-29
Return nil rather than error for non-existent receipt

joeldudley:

Geth returns `nil` if you request a transaction receipt that does not exist via RPC. Our RPC API returns an error instead. We should follow their approach.

Functional.

- Return `nil` receipt for transaction receipt if transaction cannot be found
- Skip decryption in RPC client if result is `nil`

### DIFF
--- a/go/enclave/db/interfaces.go
+++ b/go/enclave/db/interfaces.go
@@ -74,7 +74,7 @@ type TransactionStorage interface {
 	GetReceiptsByHash(hash gethcommon.Hash) types.Receipts
 
 	// GetTransaction - returns the positional metadata of the tx by hash
-	GetTransaction(txHash gethcommon.Hash) (*types.Transaction, gethcommon.Hash, uint64, uint64, error)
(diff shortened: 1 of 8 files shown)


go/enclave/db/interfaces.go
```diff
@@ -74,7 +74,7 @@ type TransactionStorage interface {
 	GetReceiptsByHash(hash gethcommon.Hash) types.Receipts
 
 	// GetTransaction - returns the positional metadata of the tx by hash
-	GetTransaction(txHash gethcommon.Hash) (*types.Transaction, gethcommon.Hash, uint64, uint64, error)
+	GetTransaction(txHash gethcommon.Hash) (*types.Transaction, gethcommon.Hash, uint64, uint64)
```
joeldudley, line 77: The `error` return was never actually set on any codepath, so I deleted it.

BedrockSquirrel (approved):
LGTM - only thing I wanted to check is if the empty string expected rather than a json `null` value in the `"result"` field?

joeldudley:
@BedrockSquirrel I think empty string is right, since the JSON serialisation is done by the Geth RPC layer and not by us. Remix seems happy with it.

PR #463, opened 2022-07-31
Avoids case where the text for a single response is awaited twice, causing an error.

joeldudley:

ObscuroScan's front-end code can sometimes await the text for a single response twice, causing an error.

Functional.

- Stores response text in a constant, and reuses that


BedrockSquirrel (approved):
LGTM

PR #464, opened 2022-08-01
Return nil rather than error for unfound transactions and blocks

joeldudley:

External tooling will work better if we align ourselves to the Geth API, which returns nil rather than error for the following API methods:

* `GetBlockByNumber`
* `GetBlockByHash`
* `GetTransactionByHash`

Functional.

- Methods above now return `nil` instead of error


BedrockSquirrel (approved):
LGTM

PR #465, opened 2022-08-01
Avoids endless HTTP hops when retrieving latest transactions.

joeldudley:

Currently, to retrieve the five latest transactions, ObscuroScan makes an RPC request for each rollup in turn, and checks its transactions. This is fine when there is reasonable network traffic, since only a few rollups will be checked. However, if the network is quiet (e.g. during a weekend initially, or before opening to external devs), you can have to work through thousands of rollups to find five transactions. This is very slow.

Refactoring.

- There's a single HTTP request between ObscuroScan and the host to get all the latest transactions, rather than one per rollup


BedrockSquirrel (approved):
LGTM

PR #466, opened 2022-08-01
Prevents potential Slowloris attack.

joeldudley:

Linter was complaining about Slowloris attack, because ObscuroScan and the wallet extension never timeout requests.

Functional.

- Adds a ten-second timeout for requests to ObscuroScan and the wallet extension


moraygrieve (approved):
LGTM

PR #467, opened 2022-08-01
Efficiently retrieve a rollup by number

joeldudley:

ObscuroScan requires the ability to retrieve a rollup by number. Currently, we have to walk the entire chain back until we find the rollup we want. This is very inefficient.

Refactoring.

- KV store now has mapping from rollup number to rollup hash, providing an efficient lookup

### DIFF
--- a/go/host/client_api_obscuro.go
+++ b/go/host/client_api_obscuro.go
@@ -53,27 +53,17 @@ func (api *ObscuroAPI) GetRollupHeader(hash gethcommon.Hash) *common.Header {
 
 // GetRollupHeaderByNumber returns the header for the rollup with the given number.
 func (api *ObscuroAPI) GetRollupHeaderByNumber(number *big.Int) (*common.Header, error) {
-	err := fmt.Errorf("no rollup with number %d is stored", number.Int64())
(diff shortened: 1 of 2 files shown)


go/host/client_api_obscuro.go
```diff
@@ -53,27 +53,17 @@ func (api *ObscuroAPI) GetRollupHeader(hash gethcommon.Hash) *common.Header {
...
-		}
-
-		rollupHeader = api.host.nodeDB.GetRollupHeader(rollupHeader.ParentHash)
-		if rollupHeader == nil {
-			return nil, err
-		}
+	if rollupHash == nil {
+		return nil, fmt.Errorf("no rollup with number %d is stored", number.Int64())
 	}
+	rollupHeader := api.host.nodeDB.GetRollupHeader(*rollupHash)
```
joeldudley, line 61: To minimise the increase in the data stored, we don't store a mapping from rollup number -> rollup header, but just from rollup number -> rollup hash, which can then be used to look up the header.

BedrockSquirrel (approved):
LGTM

PR #471, opened 2022-08-01
Efficient retrieval of rollup given a tx hash

joeldudley:

ObscuroScan needs the ability to retrieve rollups based on a transaction hash. Currently, it does this by walking the entire chain back to find the transaction hash. This will be very slow as the number of rollups grows.

Refactoring.

- Greatly speeds up the recovery of the rollup for a given tx hash


BedrockSquirrel (approved):
LGTM

PR #472, opened 2022-08-01
Was converting wrong array to JSON.

joeldudley:

Previously, the wrong JSON array was returned to the front-end, leading to fields of `Undefined` rather than `N/A`.

Functional.

- Fixes bug where latest transactions initially shows `Undefined`
- Removes info on causes from front-end (this is pushed to the logs now)

### DIFF
--- a/tools/obscuroscan/obscuroscan.go
+++ b/tools/obscuroscan/obscuroscan.go
@@ -178,7 +178,7 @@ func (o *Obscuroscan) getLatestTxs(resp http.ResponseWriter, _ *http.Request) {
 		}
 	}
 
-	jsonTxHashes, err := json.Marshal(txHashes)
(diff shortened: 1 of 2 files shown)


tools/obscuroscan/obscuroscan.go
```diff
@@ -178,7 +178,7 @@ func (o *Obscuroscan) getLatestTxs(resp http.ResponseWriter, _ *http.Request) {
 		}
 	}
 
-	jsonTxHashes, err := json.Marshal(txHashes)
+	jsonTxHashes, err := json.Marshal(txHashStrings)
```
joeldudley, line 181: This is the bug fix. I was returning the non-padded array.

BedrockSquirrel (approved):
LGTM

PR #473, opened 2022-08-02
update new erc20 local contract addr in docs

BedrockSquirrel:

- Moray noticed mistake where the OBX erc20 address was duplicated across both parameters for the example commands in documentation

- fix the ETH erc20 contract address to match what you get when it runs locally atm


moraygrieve (changes requested):
A minor point but probably worth adding another where part for the ether20addr ... at the moment further down from this section it states;

"0xC0370e0b5C1A41D447BDdA655079A1B977C71aA9 is the address of the ERC20 contract which is known a-priori as a nonce of 1 is used"

so it would be good to expand and say this is specifically for OBX, and add another line for the ETH contract etc

BedrockSquirrel:
Ah yeah, good spot Moray - that wording seem ok?

moraygrieve (approved):
LGTM

PR #476, opened 2022-08-02
Resets transaction blob view and scrolls screen correctly on rollup lookup failure.

joeldudley:

Previously, if looking up a rollup or transaction failed, it wouldn't reset the transaction blob or scroll the screen to display the result.

Functional.

- Resets transaction blob view and scrolls screen correctly on rollup lookup failure


BedrockSquirrel (approved):
LGTM

PR #480, opened 2022-08-03
Separate view key layer from low-level node client

BedrockSquirrel:

- getting rid of viewing key enabled flag in the enclave requires we support VK enc/dec in the in-memory client
- separating the view key enc/dec logic means we won't duplicate it

- separated the viewing key layer from the RPC layer in our client 
- ViewingKeyClient now wraps an underlying client and in a follow-up I'll have it wrap the in-memory version for the tests

- rpcclientlib changes

### DIFF
--- a/go/rpcclientlib/client.go
+++ b/go/rpcclientlib/client.go
@@ -28,15 +28,3 @@ type Client interface {
 	// Stop closes the client.
 	Stop()
 }
-
-func NewViewingKeyNetworkClient(rpcAddress string) (*ViewingKeyClient, error) {
-	rpcClient, err := NewNetworkClient(rpcAddress)
-	if err != nil {
-		return nil, err
-	}
-	vkClient, err := NewViewingKeyClient(rpcClient)
-	if err != nil {
-		return nil, err
-	}
-	return vkClient, nil
-}
--- a/go/rpcclientlib/client.go
+++ b/go/rpcclientlib/client.go
@@ -1,0 +1,42 @@
+package rpcclientlib
+
+const (
+	RPCGetID                   = "obscuro_getID"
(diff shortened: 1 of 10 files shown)


go/rpcclientlib/client.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Client is used by client applications to interact with the Obscuro node
+type Client interface {
+	// Call executes the named method via RPC.
+	Call(result interface{}, method string, args ...interface{}) error
+	// Stop closes the client.
+	Stop()
+}
+
+func NewViewingKeyNetworkClient(rpcAddress string) (*ViewingKeyClient, error) {
```
joeldudley, line 32: @BedrockSquirrel Feels like this lives more naturally in `viewing_key_client.go`?

integration/simulation/network/obscuro_node_utils.go
```diff
@@ -150,25 +150,25 @@ func startStandaloneObscuroNodes(params *params.SimParams, stats *stats.Stats, g
 	walletClients := make(map[string]rpcclientlib.Client)
 	var i int
 	for _, w := range params.Wallets.SimObsWallets {
-		cli, err := rpcclientlib.NewClient(nodeRPCAddresses[i%len(nodeRPCAddresses)])
+		cli, err := rpcclientlib.NewViewingKeyNetworkClient(nodeRPCAddresses[i%len(nodeRPCAddresses)])
```
joeldudley, line 153: @BedrockSquirrel Not your fault, but any chance you could rename `cli` to `client`? `cli` is reserved for CLIs IMO

BedrockSquirrel: Probably was my fault originally tbf haha

PR #482, opened 2022-08-03
Adds dummy fee history method.

joeldudley:

MetaMask expects the `eth_feeHistory` endpoint to exist. Not having it is not an issue, but causes log noise.

Functional.

- Adds dummy `eth_feeHistory` endpoint


moraygrieve (approved):
LGTM (don't have too much context around this area, but don't see anything to raise concerns on)

PR #485, opened 2022-08-03
Add L2 contract deployments to local testnet readme

BedrockSquirrel:

- missed important step for local testnet to have parity with main testnet


moraygrieve (approved):
LGTM

PR #486, opened 2022-08-03
Fix the container name

moraygrieve:

- There is a conflict in the container names used and so for the guessing game it needs to be unique

- Made the container name unique


tudor-malene (approved):
Lgtm

BedrockSquirrel (approved):
Good catch, LGTM

PR #487, opened 2022-08-04
Removes todo around generating randomness in the enclave

joeldudley:

We left a todo to check that the way we were generating randomness in the enclave was secure. Upon further research, it turned out it was.

Refectoring.

- Removes todo
- Uses helper to automate length check


BedrockSquirrel (approved):
LGTM thanks

PR #488, opened 2022-08-04
Documents the rollup encryption key.

joeldudley:

See https://github.com/obscuronet/obscuro-internal/issues/685: "Rather than use ObscuroScan a Testnet participant might want to independently decrypt the encrypted transaction blob. To achieve this they will need to use the shared secret used in Testnet. Make the shared secret available on the docsite for Testnet."

Functional.

- Adds the shared secret to the docsite


GavT (approved):
LGTM

PR #490, opened 2022-08-04
Eliminate drift between CLI flags, file-based config and actual config

joeldudley:

See https://github.com/obscuronet/obscuro-internal/issues/628. Currently, we have no mechanism to ensure that the .toml file based config and CLI flags for a host or enclave align with the actual config objects they use. We repeatedly introduce new config options that are left out of file based config especially.

Refactoring.

- Adds tests to ensure .toml config fields align with actual config fields
- Adds tests to ensure that CLI flags align with actual config fields
  - These tests are disabled for now, as making them pass would require renaming CLI flags, which should be done at a less disruptive time
- Adds missing host .toml fields


BedrockSquirrel (approved):
LGTM

PR #491, opened 2022-08-04
Adds special handling for the fact that non-existent tx receipts now returns empty string.

joeldudley:

Requesting the receipt for a non-existent transaction now returns the empty string, not an error. The simulation utils need to be updated to account for that.

Refactoring.

- Handles the special case of the receipt being empty


moraygrieve (approved):
LGTM

PR #494, opened 2022-08-04
Sets time field in rollup headers.

joeldudley:

Needed so that we can report average rollup time easily in ObscuroScan. It's also an expected field in headers.

Functional.

- Sets the time in rollup headers


tudor-malene (approved):
lgtm.
pls add a todo, or raise a ticket that we need to think about how this timestamp correlates with the timestamp of the ethereum block used as proof. And what are the implications, etc.
That's the reason why I was reluctant to add this, but if it's needed for stats, it's ok

PR #497, opened 2022-08-04
update guessing game web assets

BedrockSquirrel:

- we added some links etc. to make guessing game a bit easier to navigate/troubleshoot
- any change on https://github.com/obscuronet/number-guessing-game requires an asset change here for now

After changes, it looked like this running locally:
Screenshot 2022-08-04 at 21 18 59


GavT (approved):
LGTM

PR #498, opened 2022-08-05
Include deploying-a-smart-contract-programmatically

moraygrieve:

- Includes the section about deployment a smart contract programatically 

- Updated docs/_data/navigation.yml

- The include md file containst a link to a file which is contained in the docs/testnet directory (deploying-a-smart-contract-programmatically.py). I'm not sure if this file will automatically be bundled up or included in the site when served, so would be good to check this?


GavT (approved):
LGTM. The html is automatically constructed from the md by a GH Action when this PR is merged into main.

PR #500, opened 2022-08-05
Cleanup of go/host

joeldudley:

A few niggling structure/clean-up issues related to `go/host`

Refactoring.

- Clean-up of various sorts. See my comments


go/host/host.go
```diff
@@ -77,7 +88,7 @@ func NewHost(
 		EnclaveClient: enclaveClient,
 
 		// statistics and metrics
-		stats: collector,
+		stats: stats,
```
joeldudley, line 91: It's clearer IMO to align the field and param names.

go/host/host.go
```diff
@@ -102,7 +113,27 @@ func NewHost(
 	}
 
 	if config.HasClientRPCHTTP || config.HasClientRPCWebsockets {
-		host.rpcServer = NewRPCServer(config, host)
+		rpcAPIs := []gethrpc.API{
```
joeldudley, line 116: The separation of concerns is better if the RPC server is provided a list of RPC APIs to install, rather than having them hardcoded.

go/host/host.go
```diff
@@ -226,11 +257,6 @@ func (a *Node) ReceiveTx(tx common.EncryptedTx) {
 	a.txP2PCh <- tx
 }
 
-// DB returns the DB of the node
-func (a *Node) DB() *DB {
```
joeldudley, line 230: This getter was useless - it was only used within this file, where the field is directly accessible.

go/host/interfaces.go
```diff
@@ -21,12 +21,6 @@ type P2P interface {
 	BroadcastTx(tx common.EncryptedTx) error
 }
 
-// RPCServer is the layer responsible for handling RPC requests from Obscuro client applications.
-type RPCServer interface {
```
joeldudley, line 25: This had to be moved to avoid an import cycle.

go/host/p2p/p2p.go
```diff
@@ -18,18 +18,18 @@ import (
 
 // TODO - Provide configurable timeouts on P2P connections.
 
-// Type indicates the type of a P2P message.
-type Type uint8
+// msgType indicates the type of a P2P message.
```
joeldudley, line 21: I've made the various types and constants private where suitable.

integration/simulation/p2p/in_mem_obscuro_client.go
```diff
@@ -1,11 +1,13 @@
-package host
+package p2p
```
joeldudley, line 1: This struct was only used in tests, so I moved it out of `go/host`, which should be for production code only.

moraygrieve (approved):
Don't have too much context around this code at the moment, but nothing looks glaringly wrong to me, so LGTM

PR #501, opened 2022-08-05
Removing the 'privacy matters' section. It's wrong and repeats a lot …

CaisManai:
…of what is already on the homepage.

Adding diagram to show wallet extension flows.


docs/_data/navigation.yml
```diff
@@ -1,8 +1,6 @@
 pages:
   - title: What is Obscuro?
     children:
-      - title: Privacy Matters
-        url: what-is-obscuro/privacy-matters.html
       - title: Technology
```
GavT, line 4: Seems odd to have a privacy L2 and yet not have a page which focuses on that at all on the docsite. I'd rather see this rewritten than removed entirely.

GavT: I think its ok to repeat the homepage (you mean obscu.ro?) because some people will just go straight to the docsite knowing the website is not where the deeper tech content is

GavT (changes requested):

docs/_data/navigation.yml
CaisManai: I think that's the point though, people are going to the doc site to find the deeper tech not learn about "why privacy matters". What I can do instead is talk about privacy use-cases but the existing stuff was both wrong and marketing focused

CaisManai: @GavT I've readded the section. Can you have another look, please

GavT (approved):
Lgtm

PR #502, opened 2022-08-05
Adds timeouts for host to host P2P connections

joeldudley:

See https://github.com/obscuronet/obscuro-internal/issues/704.

Currently, host P2P connections never time out (possibly the OS eventually cuts them off, not sure). In addition, P2P messages are not broadcast in parallel. So a single unreachable node can block broadcast for the entire network.

Functional.

- Adds timeout for P2P connections
- Makes timeout configurable
- Broadcasts P2P messages in parallel


tudor-malene (approved):
lgtm.

The ultimate solution is to introduce a gossiping library, though

PR #503, opened 2022-08-05
Separation of duties for transaction blob crypto.

joeldudley:

`TransactionBlobCrypto` is currently aware of specifics of rollups and how they're mapped back and forth. Its only responsibility should be encrypting and decrypting tx blobs.

Refactoring.

- Moves logic for creating rollups and ext rollups out of `TransactionBlobCrypto`
- Makes a few methods and constants non-public

### DIFF
--- a/go/enclave/crypto/crypto.go
+++ b/go/enclave/crypto/crypto.go
@@ -19,11 +19,11 @@ const (
 	// obscuroPrivateKeyHex is the private key used for sensitive communication with the enclave.
 	// TODO - Replace this fixed key with a key derived from the master seed.
 	obscuroPrivateKeyHex = "81acce9620f0adf1728cb8df7f6b8b8df857955eb9e8b7aed6ef8390c09fc207"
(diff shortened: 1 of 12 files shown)


go/enclave/crypto/crypto.go
```diff
@@ -28,57 +33,57 @@ func GetObscuroKey() *ecdsa.PrivateKey {
...
 		log.Panic("could not generate secret. Cause: %s", err)
 	}
-	var temp [core.SharedSecretLen]byte
+	var temp [sharedSecretLen]byte
 	copy(temp[:], secret)
 	return temp
 }
 
-// EncryptWithPublicKey encrypts data with public key
-func EncryptWithPublicKey(msg []byte, pub *ecdsa.PublicKey) ([]byte, error) {
```
joeldudley, line 42: These should be internal/private.

moraygrieve (approved):
LGTM

PR #507, opened 2022-08-05
Moray/programatic deployment

moraygrieve:

- Shorten the title bar for programmatic deployment


GavT (approved):
lgtm

PR #510, opened 2022-08-08
Displays attestation report via ObscuroScan.

joeldudley:

Displaying the attestation report via ObscuroScan will be useful to show the node can be trusted.

Note that I couldn't test this fully, since I don't have an SGX node, and the existing testnet nodes don't have the `obscuro_attestation` RPC method yet (it's introduced by this PR). I tested that it could return and display a dummy report fine.

Functional.

- Adds the obscuro_attestation RPC endpoint
- Displays the attestation and associated report at `/attestation.html`


tudor-malene (approved):
lgtm

PR #515, opened 2022-08-09
Adds menu to ObscuroScan

joeldudley:

We are starting to add pages to ObscuroScan (the guessing game, and in the near future the attestation page).

Functional.

- Adds dropdown menu to ObscuroScan

Here's what it looks like:

Screenshot 2022-08-09 at 11 21 17


BedrockSquirrel (approved):
LGTM, could potentially go with "Block explorer" or "ObscuroScan" instead of "Home"? But I think Home's perfectly clear too, it is the root after all.

PR #516, opened 2022-08-09
Verbose linting.

joeldudley:

No downsides, and may help us detect linter issues when the linter fails.

Functional.

- Adds verbose flag to linter


otherview (approved):
Not 100% on what this outputs, but can't hurt I guess.

PR #518, opened 2022-08-09
Handle case where JSON can't be marshalled in attestation

joeldudley:

There is an edgecase where the JSON fails to be marshalled when verifying an attestation. This case should be caught and flagged.

Functional.

- Catch JSON marshalling error in attestation verification


BedrockSquirrel (approved):
LGTM

PR #520, opened 2022-08-09
ObscuroScan: Update Docker image to produce attestation report

joeldudley:

The existing ObscuroScan Docker image does not allow it to verify attestations on the `/attestation` page.

Functional.

- Modify Docker image and script so that it can verify remote reports


BedrockSquirrel (approved):
LGTM

PR #524, opened 2022-08-10
Pulls code for identifying which viewing key to use into separate utils

joeldudley:

In the future, the wallet extension will support multiple viewing keys. This will require the ability to ascertain which viewing key should be used for a given request. This can be established based on the contents of the request, but the logic to do so is currently spread out across the enclave code.

This PR pulls that code out so it can be reused cleanly.

Refactoring.

- Pulls code for parsing RPC requests, including establishing which viewing key to use, into a separate utils class
- Moves any non-encryption code out of RPCEncryptionManager, and into these new utils


go/enclave/rollupchain/rollup_chain.go
```diff
@@ -667,57 +668,19 @@ func (rc *RollupChain) ExecuteOffChainTransaction(encryptedParams common.Encrypt
 	return encryptedResult, nil
 }
 
-// Extracts and validates the relevant parameters in a Call request.
-func extractCallParams(decryptedParams []byte) (gethcommon.Address, gethcommon.Address, []byte, error) {
```
joeldudley, line 671: This was moved across into the RPC utils file, and broken into three separate methods.

go/enclave/rpc/rpc_encryption_manager.go
```diff
@@ -100,48 +95,3 @@ func (rpc *RPCEncryptionManager) EncryptWithViewingKey(address gethcommon.Addres
 
 	return encryptedBytes, nil
 }
-
-// ExtractTxHash - Returns the transaction hash from a common.EncryptedParamsGetTxReceipt object.
-func (rpc *RPCEncryptionManager) ExtractTxHash(encryptedParams common.EncryptedParamsGetTxReceipt) (gethcommon.Hash, error) {
```
joeldudley, line 105: Stuff like this isn't strictly encryption related, and has been moved off into the utils class.

go/enclave/rpc/rpc_utils.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		return gethcommon.Address{}, fmt.Errorf("could not parse JSON params in eth_getBalance request. JSON "+
+			"params are: %s. Cause: %w", string(balanceParams), err)
+	}
+	// The first argument is the address, the second the block.
+	return gethcommon.HexToAddress(paramsJSONMap[0]), nil
+}
+
+// ExtractCallParamTo extracts and parses the `to` field of an eth_call request.
+func ExtractCallParamTo(callParams []byte) (gethcommon.Address, error) {
```
BedrockSquirrel, line 76: Could we have a CallParams struct and then just a single `ExtractCallParams` method that returns all of them bundled together. Would clear up the duplication here both in code and in CPU work,

BedrockSquirrel (commented):
Generally LGTM, minor thought about the call data extraction

go/enclave/rpc/rpc_utils.go
joeldudley: I would do that, but on the WE side, we only need to extract one of them (the `from`), so I thought this was the best.

BedrockSquirrel: Personally I'd rather extract the fields unnecessarily for the WE than duplicate the code and work in the enclave but don't feel strongly.

BedrockSquirrel (approved):
LGTM

PR #526, opened 2022-08-10
Simplifies ViewingKeyClient.Call logic.

joeldudley:

Minor. There's an assignment we can skip, since variable declarations are `nil` by default.

- list of changes

### DIFF
--- a/go/rpcclientlib/viewing_key_client.go
+++ b/go/rpcclientlib/viewing_key_client.go
@@ -68,18 +68,11 @@ func (c *ViewingKeyClient) Call(result interface{}, method string, args ...inter
 		return c.obscuroClient.Call(result, method, args...)
 	}
 
-	// we setup a generic rawResult to receive the response (then we can decrypt it as necessary into the requested result type)
-	var rawResult interface{}
-	if result == nil {
-		// we set result receiver to nil if caller set nil since no return needs to be set
-		rawResult = nil
-	}
-
 	var err error
 	if method == RPCCall {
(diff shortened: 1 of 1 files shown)


go/rpcclientlib/viewing_key_client.go
```diff
@@ -68,18 +68,11 @@ func (c *ViewingKeyClient) Call(result interface{}, method string, args ...inter
 		return c.obscuroClient.Call(result, method, args...)
 	}
 
-	// we setup a generic rawResult to receive the response (then we can decrypt it as necessary into the requested result type)
-	var rawResult interface{}
-	if result == nil {
-		// we set result receiver to nil if caller set nil since no return needs to be set
-		rawResult = nil
```
joeldudley, line 75: This is `nil` anyway at this point.

go/rpcclientlib/viewing_key_client.go
```diff
@@ -68,18 +68,11 @@ func (c *ViewingKeyClient) Call(result interface{}, method string, args ...inter
...
-	}
-
 	var err error
 	if method == RPCCall {
 		// RPCCall is a sensitive method that requires a viewing key lookup but the 'from' field is not mandatory in geth
 		//	and is often not included from metamask etc. So we ensure it is populated here.
-		args, err = c.ensureCallParamsHaveFromAddress(method, args)
+		args, err = c.addFromAddressToCallParamsIfMissing(method, args)
```
joeldudley, line 75: Thought this name was better, to make it clear that it mutated the args (instead of performing an assertion).

BedrockSquirrel (approved):

PR #529, opened 2022-08-10
Moves setting of VK address earlier in lifecycle. Clean up.

joeldudley:

Currently, when using the viewing key client, the viewing key's address is only passed when submitting the viewing key, not when generating it.

As we move towards supporting multiple viewing keys in a single wallet extension, we'll need to pass the address at generation time, so we can store the generated viewing key against the address.

- list of changes

### DIFF
--- a/go/rpcclientlib/viewing_key_client.go
+++ b/go/rpcclientlib/viewing_key_client.go
@@ -175,14 +175,13 @@ func (c *ViewingKeyClient) Stop() {
 	c.obscuroClient.Stop()
 }
 
-func (c *ViewingKeyClient) SetViewingKey(viewingKey *ecies.PrivateKey, viewingPubKeyBytes []byte) {
+func (c *ViewingKeyClient) SetViewingKey(viewingKey *ecies.PrivateKey, signerAddress common.Address, viewingPubKeyBytes []byte) {
 	c.viewingPrivKey = viewingKey
 	c.viewingPubKey = viewingPubKeyBytes
+	c.viewingKeyAddr = signerAddress
 }
 
-func (c *ViewingKeyClient) RegisterViewingKey(signerAddr common.Address, signature []byte) error {
-	c.viewingKeyAddr = signerAddr
-
+func (c *ViewingKeyClient) RegisterViewingKey(signature []byte) error {
 	// We encrypt the viewing key bytes
 	encryptedViewingKeyBytes, err := ecies.Encrypt(rand.Reader, c.enclavePublicKey, c.viewingPubKey, nil, nil)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


go/rpcclientlib/viewing_key_client.go
```diff
@@ -175,14 +175,13 @@ func (c *ViewingKeyClient) Stop() {
...
 
-func (c *ViewingKeyClient) SetViewingKey(viewingKey *ecies.PrivateKey, viewingPubKeyBytes []byte) {
+func (c *ViewingKeyClient) SetViewingKey(viewingKey *ecies.PrivateKey, signerAddress common.Address, viewingPubKeyBytes []byte) {
 	c.viewingPrivKey = viewingKey
 	c.viewingPubKey = viewingPubKeyBytes
+	c.viewingKeyAddr = signerAddress
 }
 
-func (c *ViewingKeyClient) RegisterViewingKey(signerAddr common.Address, signature []byte) error {
-	c.viewingKeyAddr = signerAddr
```
joeldudley, line 184: This is set at generation time now.

go/rpcclientlib/viewing_key_client.go
```diff
@@ -220,10 +219,13 @@ func (c *ViewingKeyClient) addFromAddressToCallParamsIfMissing(method string, ar
 		// We only modify `eth_call` requests where the `from` field is not set.
 		return args, nil
 	}
+
+	if c.viewingKeyAddr == (common.Address{}) {
```
joeldudley, line 223: This check was missing before. Best to check this, to not accidentally leave the from field as nil and send a wasted request.

tools/walletextension/static/javascript.js
```diff
@@ -23,23 +23,29 @@ const initialize = () => {
...
         const viewingKey = await viewingKeyResp.text();
 
-        const accounts = await ethereum.request({method: metamaskRequestAccounts});
-        if (accounts.length === 0) {
-            statusArea.innerText = "No MetaMask accounts found."
-            return
-        }
-        // The array returns "An array of a single, hexadecimal Ethereum address string.")
-        // We use the last created account for the viewing key since it makes most sense for testing.
-        const account = accounts[accounts.length - 1];
```
joeldudley, line 41: This didn't make sense to me. Accounts is always a single item, it says above. So just grab the first.

tools/walletextension/wallet_extension.go
```diff
@@ -269,18 +284,19 @@ func (we *WalletExtension) handleSubmitViewingKey(resp http.ResponseWriter, req
 		return
 	}
 
-	// We drop the leading "0x", and transform the V from 27/28 to 0/1.
```
joeldudley, line 272: Half of the comment was moved further down, to where the transformation was actually performed.

BedrockSquirrel (approved):
LGTM - should make sure Moray is happy before merging if you haven't spoken to him already. Don't think there are any other users of that interface atm

moraygrieve (approved):
LGTM

PR #533, opened 2022-08-11
Use error for RPC nil response

BedrockSquirrel:

- we can't populate the typed pointer in the rpc client when we have a nil response
- we return a specific Nil Response error so that caller can check for it (no more onerous than if they had to nil-check their data type, and this is more idiomatic Go)

- add nil response to client
- add nil transaction response to higher-level client

### Evidence
The contract deployment now succeeds against testnet:
```
Preparing contract deployer with config: {
  "NodeHost": "obscuronode-0-testnet-1.uksouth.cloudapp.azure.com",
  "NodePort": 13000,
  "IsL1Deployment": false,
  "PrivateKey": "6e384a07a01263518a09a5424c7b6bbfc3604ba7d93f47e3a455cbdd7f9f0682",
  "ChainID": 777,
  "ContractName": "L2ERC20",
  "ConstructorParams": [
    "JAM",
    "JAM",
    "1000000000000000000000"
  ]
}
0xdDb7b4973DfC4F151BF7DEf4fD3ABFd899Bc1778
```

### DIFF
--- a/go/ethadapter/obscuro_rpc_client.go
+++ b/go/ethadapter/obscuro_rpc_client.go
@@ -15,9 +15,7 @@ import (
 	"github.com/obscuronet/go-obscuro/integration/common/viewkey"
 )
 
-var (
-	ErrReceiptNotFound = errors.New("receipt not found, received nil response")
-)
+var ErrReceiptNotFound = errors.New("receipt not found, received nil response")
 
 // obscuroWalletRPCClient implements the EthClient interface, it's bound to a single wallet (viewing key and address)
 //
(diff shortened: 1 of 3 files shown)


go/ethadapter/obscuro_rpc_client.go
```diff
@@ -62,6 +67,9 @@ func (c *obscuroWalletRPCClient) TransactionReceipt(hash gethcommon.Hash) (*type
 	var r types.Receipt
 	err := c.client.Call(&r, rpcclientlib.RPCGetTxReceipt, hash)
 	if err != nil {
+		if err == rpcclientlib.ErrNilResponse {
+			return nil, ErrReceiptNotFound
```
BedrockSquirrel, line 71: Not 100% necessary to but feel it will be cleaner to have a specific error from this client, can be checked for by caller without depending on the upstream

joeldudley (approved):

tudor-malene (approved):
lgtm

PR #535, opened 2022-08-11
Allow multiple viewing keys to be stored (still a single key used for decryption)

joeldudley:

Another step in enabling multiple viewing keys. We still use a single key for decryption, but we allow registration of multiple viewing keys now.

Functional.

- Allows multiple viewing keys to be stored, keyed by address

### DIFF
--- a/go/rpcclientlib/viewing_key_client.go
+++ b/go/rpcclientlib/viewing_key_client.go
@@ -61,7 +61,7 @@ type ViewingKeyClient struct {
 	obscuroClient Client
 
 	enclavePublicKey *ecies.PublicKey
-	// todo - joel - remove this field
+	// TODO - Use multiple keys below to perform decryption.
 	viewingPrivKey *ecies.PrivateKey // private viewing key to use for decrypting sensitive requests
 
 	viewingKeysPrivate map[common.Address]*ecies.PrivateKey // Maps an address to its private viewing key.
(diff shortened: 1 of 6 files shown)


tools/walletextension/static/javascript.js
```diff
@@ -56,7 +56,7 @@ const initialize = () => {
             return
         }
 
-        const signedViewingKeyJson = {"signature": signature}
+        const signedViewingKeyJson = {"signature": signature, "address": account}
```
BedrockSquirrel, line 59: Need to mention this to Moray again if you haven't already, in case he stopped sending it

go/rpcclientlib/viewing_key_client.go
```diff
@@ -181,13 +187,14 @@ func (c *ViewingKeyClient) Stop() {
...
 	c.viewingPrivKey = viewingKey
-	c.viewingPubKey = viewingPubKeyBytes
-	c.viewingKeyAddr = signerAddress
+
+	c.viewingKeysPrivate[signerAddress] = viewingKey
+	c.viewingKeysPublic[signerAddress] = viewingPubKeyBytes
 }
 
-func (c *ViewingKeyClient) RegisterViewingKey(signature []byte) error {
+func (c *ViewingKeyClient) RegisterViewingKey(signature []byte, signerAddress common.Address) error {
```
joeldudley, line 195: We need to repass this now, since we don't know which address the signature is for.

BedrockSquirrel (approved):
LGTM - we should make sure these flows are tested in the integration tests, maybe a test in wallet_ext_tests that doesn't set a from field or maybe we just ask Moray to cover it in his test suites since he is mimicking the users.

tools/walletextension/static/javascript.js
joeldudley: 👍

PR #536, opened 2022-08-11
Decrypt with correct viewing key in wallet extension

joeldudley:

Currently, the wallet extension supports multiple viewing keys, but always decrypts with the same one.

Functional.

- Wallet extension decrypts with correct viewing key, even if multiple are registered
- Integration test of wallet extension decrypting successfully when multiple viewing keys are present


BedrockSquirrel (approved):
LGTM

PR #537, opened 2022-08-11
Uses helper to reduce duplication in WE tests.

joeldudley:

Pulls registering an account into a helper in the wallet extension tests.

Refactoring.


moraygrieve (approved):
LGTM

PR #541, opened 2022-08-11
Break up RPC APIs

joeldudley:

Currently, we have every custom API lumped into the `obscuro_` namespace. But there are actually three kinds of APIs:

* Ones required by a running production Obscuro node
* Ones only required to power ObscuroScan
* Ones only required for testing

Splitting up these APIs makes things cleaner, and enables us to potentially disable test APIs in production nodes in the future.

Refactoring.

- Creates a new API namespace, `obscuroscan_`, for ObscuroScan API methods
- Creates a new API namespace, `test_`, for methods only required in tests
- Removes `obscuro_nonce`, since this duplicates `eth_getTransactionCount`


BedrockSquirrel:
Just a minor thought cos this PR caught my eye. The methods you've moved to the `obscuroscan_` namespace feel quite generic to me, like might make sense to put them in that `net_` namespace as generic network-level operations or some other more general name.

Was just thinking that, once people are using these APIs, we won't be able to rename them easily and you could imagine half a dozen tools springing up that use some of the obscuro scan methods. And then we'll be explaining to a new hire in 6 months that such-and-such method is under the `obscuroscan_` name "for legacy reasons".

joeldudley:
@BedrockSquirrel So firstly, I don't want to touch the `net_` namespace - that's an Ethereum/Geth one.

My intention here was to avoid the ObscuroScan and tests methods being picked up by third-party devs as an official part of the platform. They may have utility to third-party developers, but not much work has gone into working out if they're the correct APIs. They may not even be correct for ObscuroScan in the next iteration.

For the test APIs, we can just turn them off with a flag. But for the ObscuroScan ones, I'm not sure - I see your point that if ppl are going to start using them anyway, they should be in a good namespace.

I propose instead that we rename the `obscuroscan_` namespace to `internal_`, with a note in the function docs. Unlikely to prevent use, but will hopefully discourage a little bit?

@tudor-malene Thoughts?

tudor-malene (approved):
lgtm

PR #547, opened 2022-08-12
Adds method to retrieve a rollup by height.

joeldudley:

The API for `eth_getBalance` requires that we can check the balance at any block. Currently, there is no API to grab the rollup with a specific height.

Functional.

- Adds API to get rollup with a specific height

### DIFF
--- a/go/enclave/db/storage.go
+++ b/go/enclave/db/storage.go
@@ -56,6 +56,19 @@ func (s *storageImpl) FetchGenesisRollup() *core.Rollup {
 	return r
 }
 
+func (s *storageImpl) FetchRollupByHeight(height uint64) *core.Rollup {
+	if height == 0 {
+		return s.FetchGenesisRollup()
+	}
+
+	hash := obscurorawdb.ReadCanonicalHash(s.db, height)
+	if hash == (gethcommon.Hash{}) {
+		return nil
+	}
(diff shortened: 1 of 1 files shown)


go/enclave/db/storage.go
```diff
@@ -56,6 +56,19 @@ func (s *storageImpl) FetchGenesisRollup() *core.Rollup {
...
+func (s *storageImpl) FetchRollupByHeight(height uint64) *core.Rollup {
+	if height == 0 {
+		return s.FetchGenesisRollup()
+	}
+
+	hash := obscurorawdb.ReadCanonicalHash(s.db, height)
+	if hash == (gethcommon.Hash{}) {
+		return nil
+	}
+	r, _ := s.FetchRollup(hash)
```
BedrockSquirrel, line 68: Should we also return boolean for success from this method?

BedrockSquirrel (commented):
Minor thought that might help users of the code and seems consistent with other methods in storage code

go/enclave/db/storage.go
joeldudley: Yeah. Sorry, was copying `FetchGenesisRollup`, but that always succeeds.

BedrockSquirrel (approved):
LGTM

PR #553, opened 2022-08-16
Correctly handles returned value from eth_getTransactionCount request.

joeldudley:

Previously, we tried to get the result directly as a uint64, but it is returned as a string.

Functional.

- Correctly converts result from string


tudor-malene (approved):
lgtm

PR #563, opened 2022-08-17
Fix rollup existence check in eth_getCode

joeldudley:

There was a bug - we only retrieved the code if the rollup did not exist, rather than if it did.

Functional.

- Fix bug

- ...

- [ ] Merged into dev-testnet 


moraygrieve (approved):
LGTM

PR #565, opened 2022-08-17
Uses t.Cleanup to simplify wallet extension tests.

joeldudley:

Previously, we used `defer` instead of `t.Cleanup` in the wallet extension tests. This forced us to write lots of duplicated codes so that the `defer`s were called in the right places.

Refactoring.

- ...

- [ ] Merged into dev-testnet 


BedrockSquirrel (approved):
LGTM, that Cleanup method is handy, new to me

PR #566, opened 2022-08-17
Returns an error instead of panicking if cannot submit block.

joeldudley:

I've started seeing test failures where a block submission happens after we've shut down the node. Shutting down the node causes it to break its connection to the enclave, which causes panics in subsequent RPC calls to submit further blocks (since the connection is broken) if these happen before the test concludes.

Here is the failure message:

```
Aug 17 12:35:49.990 INF Stopped monitoring for l1 blocks
Aug 17 12:35:49.990 PNC >   Agg0: Failed to submit block. Cause: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp 127.0.0.1:31301: connect: connection refused"
```

You can see by the "Stopped monitoring for l1 blocks" message that this is happening at the end of the test, during shutdown.

Functional.

- `Enclave.SubmitBlock` now returns an error if the block cannot be submitted, instead of panicking.

- ...

- [ ] Merged into dev-testnet 

### DIFF
--- a/go/common/enclave.go
+++ b/go/common/enclave.go
@@ -41,7 +41,7 @@ type Enclave interface {
 	// it is the responsibility of the host to gossip the returned rollup
 	// For good functioning the caller should always submit blocks ordered by height
 	// submitting a block before receiving a parent of it, will result in it being ignored
-	SubmitBlock(block types.Block) BlockSubmissionResponse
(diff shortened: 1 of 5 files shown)


tudor-malene (approved):
lgtm

PR #567, opened 2022-08-17
Allows enclaves to configure the minimum gas amount they'll accept

joeldudley:

Currently, enclaves accept any gas price. They should be able to choose to reject txs with insufficient gas.

Functional.

- Enclave can configure minimum gas price for accepting transaction
- Transactions with an insufficient gas price are rejected

- ...

- [ ] Merged into dev-testnet 


tudor-malene (approved):
lgtm

PR #568, opened 2022-08-17
Relies on enclave config defaults where suitable.

joeldudley:

In the integration tests, we use a mishmash of setting enclave config and leaving other fields unset. I've moved it to using the default config as a base, updating it where necessary.

Refactoring.

- Uses enclave config defaults where suitable

- ...

- [ ] Merged into dev-testnet 

### DIFF
--- a/integration/simulation/network/network_utils.go
+++ b/integration/simulation/network/network_utils.go
@@ -78,20 +78,16 @@ func createInMemObscuroNode(
 		HasClientRPCHTTP:    false,
 	}
 
-	enclaveConfig := config.EnclaveConfig{
-		HostID:                 hostConfig.ID,
-		L1ChainID:              integration.EthereumChainID,
-		ObscuroChainID:         integration.ObscuroChainID,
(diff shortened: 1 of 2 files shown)


integration/simulation/network/network_utils.go
```diff
@@ -78,20 +78,16 @@ func createInMemObscuroNode(
...
+	enclaveConfig.HostID = hostConfig.ID
+	enclaveConfig.ValidateL1Blocks = validateBlocks
+	enclaveConfig.GenesisJSON = genesisJSON
+	enclaveConfig.ERC20ContractAddresses = wallets.AllEthAddresses()
+
 	enclaveClient := enclave.NewEnclave(enclaveConfig, mgmtContractLib, stableTokenContractLib, stats)
 
 	// create an in memory obscuro node
-	node := node.NewHost(
+	inMemNode := node.NewHost(
```
joeldudley, line 90: I renamed these because they conflicted with a package name.

PR #569, opened 2022-08-18
refactor wallet extension to use a map of clients

BedrockSquirrel:

- Pre-requisite for the go-obs client
- (hopefully) makes the RPC client logic layers easier to reason about
- The EncRPCClient introduced here is used in *both* the wallet extension and the AuthClient that's coming.

This change also moves the multi-acc fuzzy stuff out of the core client and just into wallet extension now for cleanliness.

- Add viewing key type per Tudor's suggestion to wrap account with viewing key and signature
- Rename ViewingKeyClient -> EncRPCClient, a VK-encrypting/decrypting client which is bound to a single account for its lifespan
- rework wallet extension to use a map of acc addresses -> EncRPCClients, with logic to pick the right one or retry them all

Note: this change means we will round-robin all registered clients when it's not obvious which to use, we should monitor and optimise this over time.

Note2: as Joel pointed out this means we find and cast the params map twice (in wallet ext to deduce the correct client, in EncRPCClient to make sure the 'from' is set)

EncRPCClient, WalletExtension

- [ ] Merged into dev-testnet 

### DIFF
--- a/go/rpcclientlib/encrypted_client.go
+++ b/go/rpcclientlib/encrypted_client.go
@@ -5,15 +5,13 @@ import (
 	"encoding/json"
 	"fmt"
 
-	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
+
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto/ecies"
 )
 
 const (
-	http           = "http://"
-	reqJSONKeyFrom = "from"
-
 	// todo: this is a convenience for testnet testing and will eventually be retrieved from the L1
 	enclavePublicKeyHex = "034d3b7e63a8bcd532ee3d1d6ecad9d67fca7821981a044551f0f0cbec74d0bc5e"
 )
@@ -21,11 +19,11 @@ const (
 // for these methods, the RPC method's requests and responses should be encrypted
 var sensitiveMethods = []string{RPCCall, RPCGetBalance, RPCGetTxReceipt, RPCSendRawTransaction, RPCGetTransactionByHash}
 
-type ViewingKey struct {
-	Account    *common.Address   // Account address that this private key is bound to
-	PrivateKey *ecies.PrivateKey // private viewing key
-	PublicKey  []byte            // public viewing key in bytes to share with enclave
-	SignedKey  []byte            // public viewing key signed by the Account's private key
+// EncRPCClient is a Client wrapper that implements Client but also has extra functionality for managing viewing key registration and decryption
+type EncRPCClient struct {
+	obscuroClient    Client
+	enclavePublicKey *ecies.PublicKey // Used to encrypt messages destined to the enclave.
+	viewingKey       *ViewingKey
 }
 
 // NewEncRPCClient sets up a client with a viewing key for encrypted communication (this submits the VK to the enclave)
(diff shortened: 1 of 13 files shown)


go/rpcclientlib/encrypted_client.go
```diff
@@ -4,19 +4,15 @@ import (
...
 	"github.com/ethereum/go-ethereum/crypto/ecies"
 )
 
 const (
-	http                = "http://"
-	reqJSONKeyFrom      = "from"
-	reqJSONKeyData      = "data"
-	ethCallPaddedArgLen = 64
-	ethCallAddrPadding  = "000000000000000000000000"
+	http           = "http://"
```
tudor-malene, line 14: This file needs to have a single responsibility, which is the encClient.
All the other stuff should be removed.
It should start with the `EncRPCClient` struct, and the comment that it's a wrapper..

For example these will not be needed after the suggested refactorings.

tudor-malene (changes requested):
Thanks Matt. I think this is definitely much clearer. 
I left a few suggestions, which should clarify and simplify further.

go/rpcclientlib/encrypted_client.go
```diff
@@ -216,7 +209,7 @@ func (c *ViewingKeyClient) RegisterViewingKey(signature []byte, signerAddress co
 // The enclave requires the `from` field to be set so that it can encrypt the response, but sources like MetaMask often
 // don't set it. So we check whether it's present; if absent, we walk through the arguments in the request's `data`
 // field, and if any of the arguments match our viewing key address, we set the `from` field to that address.
-func (c *ViewingKeyClient) setFromFieldIfMissing(args []interface{}) ([]interface{}, error) {
+func (c *EncRPCClient) setFromFieldIfMissing(args []interface{}) ([]interface{}, error) {
```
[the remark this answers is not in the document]
BedrockSquirrel: This does set it to the address of the current viewing key. But perhaps it shouldn't check (so having from set is required) and the wallet extension should make sure it sets it if needed before each attempted request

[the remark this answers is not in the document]
tudor-malene: ah, ok. then the comment is wrong.
I'd say we add a check that the from field, if set, is the current vk.account. If not set, we set it to that

tools/walletextension/wallet_extension.go
```diff
@@ -162,8 +160,25 @@ func (we *WalletExtension) handleHTTPEthJSON(resp http.ResponseWriter, req *http
 		return
 	}
 
+	// for obscuro RPC requests it is important we know the sender account for the viewing key encryption/decryption
```
tudor-malene, line 163: feels like all this logic from 163 to 181 belong in a function in the multi_acc_helper

tudor-malene (commented):
looks good. One small thing

tools/walletextension/multi_acc_helper.go
```diff
@@ -0,0 +1,169 @@
...
+	var err error
+	if suggestedClient != nil {
+		// todo: if we have a suggested client, should we still loop through the other clients if it fails?
+		// 		The call data guessing won't often be wrong but there could be edge-cases there
+		err = executeCall(suggestedClient, rpcReq, &rpcResp)
+	} else {
+		// we attempt the request with every client until we have a successful execution
+		log.Info("appropriate client not found, attempting request with up to %d clients", len(accClients))
+		for _, client := range accClients {
+			err = executeCall(client, rpcReq, &rpcResp)
```
tudor-malene, line 160: one last thing.
executeCall should prob be in the same file.

tudor-malene (approved):
lgtm.
one minor function move, and good to merge.

tools/walletextension/multi_acc_helper.go
BedrockSquirrel: Oops, yeah good spot :)

PR #573, opened 2022-08-18
Waits for receipt when submitting tx in wallet extension tests

joeldudley:

With non-zero gas prices becoming mandatory, we need to do faucet transfers in the wallet extension tests.

As a precursor to that, I wanted to introduce logic in the tests to wait for a tx's receipt (this will be useful to wait for the preallocation). This helped refactor out some other duplicated code.

Refactoring.

- Introduces method to wait for transaction receipt
- General clean-up

- ...

### DIFF
--- a/integration/walletextension/wallet_extension_test.go
+++ b/integration/walletextension/wallet_extension_test.go
@@ -49,8 +49,9 @@ const (
 	reqJSONKeyTo      = "to"
 	reqJSONKeyFrom    = "from"
 	reqJSONKeyData    = "data"
-	respJSONKeyResult = "result"
+	respJSONKeyStatus = "status"
 	latestBlock       = "latest"
+	statusSuccess     = "0x1"
 	errInsecure       = "enclave could not respond securely to %s request"
 
 	networkStartPort = integration.StartPortWalletExtensionTest + 1
@@ -93,8 +94,8 @@ func TestCanMakeNonSensitiveRequestWithoutSubmittingViewingKey(t *testing.T) {
 
 	respJSON := makeEthJSONReqAsJSON(t, walletExtensionAddr, rpcclientlib.RPCChainID, []string{})
 
-	if respJSON[respJSONKeyResult] != l2ChainIDHex {
-		t.Fatalf("Expected chainId of %s, got %s", l2ChainIDHex, respJSON[respJSONKeyResult])
+	if respJSON[walletextension.RespJSONKeyResult] != l2ChainIDHex {
+		t.Fatalf("Expected chainId of %s, got %s", l2ChainIDHex, respJSON[walletextension.RespJSONKeyResult])
 	}
 }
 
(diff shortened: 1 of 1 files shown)


integration/walletextension/wallet_extension_test.go
```diff
@@ -537,46 +533,59 @@ func createObscuroNetwork(t *testing.T) {
...
 
-	// Verify the Obscuro ERC20 contract deployed successfully
-	_, ok := sendTxJSON[walletextension.RespJSONKeyResult].(string)
+	// Verify the transaction was successful.
+	txHash, ok := sendTxJSON[walletextension.RespJSONKeyResult].(string)
 	if !ok {
 		panic("could not retrieve transaction hash from JSON result, failed to deploy ERC20")
 	}
 
-	// We wait ten seconds for the first rollup to be published, to ensure the network is ready.
```
joeldudley, line 560: I don't think we need this wait for the first rollup, now that we wait for the transaction receipt instead.

integration/walletextension/wallet_extension_test.go
```diff
@@ -391,15 +387,15 @@ func waitForWalletExtension(t *testing.T, walletExtensionAddr string) {
...
+func makeEthJSONReq(walletExtensionAddr string, method string, params interface{}) []byte {
 	reqBodyBytes, err := json.Marshal(map[string]interface{}{
 		"jsonrpc": "2.0",
 		"method":  method,
 		"params":  params,
 		"id":      "1",
 	})
 	if err != nil {
-		t.Fatal(err)
+		panic(err)
```
BedrockSquirrel, line 398: I'm not sure about this. The [Uber style-guide](https://github.com/uber-go/guide/blob/master/style.md#dont-panic) says: `Even in tests, prefer t.Fatal or t.FailNow over panics to ensure that the test is marked as failed.`

I do sometimes see a sim test show as green when it clearly broke (e.g. if I leave a debugger point for too long then when I eventually quit it it shows test passed).

integration/walletextension/wallet_extension_test.go
```diff
@@ -537,46 +533,59 @@ func createObscuroNetwork(t *testing.T) {
...
 		panic("could not retrieve transaction hash from JSON result, failed to deploy ERC20")
 	}
 
-	// We wait ten seconds for the first rollup to be published, to ensure the network is ready.
-	firstRollupIdx := 1
 	counter := 0
 	for {
 		if counter > 10 {
-			t.Fatalf("first rollup had not been published after 10 seconds")
+			panic("could not get ERC20 receipt after 10 seconds")
```
BedrockSquirrel, line 575: Is 10 seconds necessary? There's 12 tests in here, in worst case scenario does this add 2mins to the time for build to fail?

BedrockSquirrel (commented):
Couple of thoughts but the waiting for receipt bit seems good to me

integration/walletextension/wallet_extension_test.go
joeldudley: No, I think you'd fail after the first test (10 secs). Plus I'm not worried about the tests being slow when they're broken. It's them being slow when they pass that worries me more.

integration/walletextension/wallet_extension_test.go
joeldudley: I've also heard arguments in the opposite direction, because `panic` gives you a stacktrace, whereas t.Fatal is opaque (no indication of stack at time of failure).

BedrockSquirrel (approved):
LGTM

PR #575, opened 2022-08-18
WE tests: More tidying. Retrieves nonce via wallet extension, not directly via clients

joeldudley:

More clearing of the way to enable setting of a gas price in the WE tests.

Functional.

- WE tests now retrieve nonce via wallet extension, rather than going directly via clients (which is not what we're trying to test here)
- Some more clean-up

- ...


BedrockSquirrel (approved):
LGTM

PR #579, opened 2022-08-18
Makes sure Geth nodes are terminated if there are panics in the constructor.

joeldudley:

Currently, there are various panics in the `NewGethNetwork` constructor, where we do not clean up properly before exiting.

Functional.

- Tries to stop all node processes if panicking during `NewGethNetwork`

- ...


BedrockSquirrel (approved):
LGTM

PR #581, opened 2022-08-18
Contract deployer fixes

joeldudley:

Various changes are required to the contract deployer to support preallocs.

Functional.

- Logs better error messages
- Makes sure prealloc tx succeeds
- Proper handling of default constructor params

- ...

### DIFF
--- a/tools/contractdeployer/cli.go
+++ b/tools/contractdeployer/cli.go
@@ -6,6 +6,11 @@ import (
 	"strings"
 )
 
+const (
+	chainIDPlaceholder           = 0
+	constructorParamsPlaceholder = ""
+)
+
 var (
 	defaultL1ChainID = big.NewInt(1337)
 	defaultL2ChainID = big.NewInt(777)
@@ -45,8 +50,8 @@ func ParseConfig() *Config {
 	contractName := flag.String(contractNameName, defaultConfig.ContractName, contractNameUsage)
 	privateKeyStr := flag.String(privateKeyName, defaultConfig.PrivateKey, privateKeyUsage)
 	// if this flag has a non-zero value it will be used instead of the default chain IDs
-	overrideChainID := flag.Int64(chainIDName, 0, chainIDUsage)
-	constructorParams := flag.String(constructorParamsName, "", constructorParamsUsage)
+	overrideChainID := flag.Int64(chainIDName, chainIDPlaceholder, chainIDUsage)
+	constructorParams := flag.String(constructorParamsName, constructorParamsPlaceholder, constructorParamsUsage)
 
 	flag.Parse()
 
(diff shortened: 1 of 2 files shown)


tools/contractdeployer/cli.go
```diff
@@ -6,6 +6,11 @@ import (
 	"strings"
 )
 
+const (
+	chainIDPlaceholder           = 0
```
BedrockSquirrel, line 10: Little comment here might be good to say these are the go default values used for comparison to see if an argument value was passed.

tools/contractdeployer/contract_deployer.go
```diff
@@ -203,25 +206,22 @@ func signAndSendTxWithReceipt(wallet wallet.Wallet, client ethadapter.EthClient,
 	}
 
 	var start time.Time
-	var receipt *types.Receipt
-	var contractAddr *common.Address
 	for start = time.Now(); time.Since(start) < timeoutWait; time.Sleep(retryInterval) {
-		receipt, err = client.TransactionReceipt(signedTx.Hash())
+		receipt, err := client.TransactionReceipt(signedTx.Hash())
+		if err != nil {
+			log.Info(err.Error())
```
BedrockSquirrel, line 212: Maybe add something to this like `"Attempt failed, retrying... %w"` so if anyone does look at these logs they don't get too worried by the semi-spam

tools/contractdeployer/contract_deployer.go
```diff
@@ -203,25 +206,22 @@ func signAndSendTxWithReceipt(wallet wallet.Wallet, client ethadapter.EthClient,
...
+			return &receipt.ContractAddress, nil
 		}
 
 		log.Info("Contract deploy tx has not been mined into a block after %s...", time.Since(start))
 	}
-	if contractAddr == nil {
-		return nil, fmt.Errorf("failed to mine contract deploy tx %s into a block after %s. Aborting", signedTx.Hash(), time.Since(start))
-	}
-	return contractAddr, nil
+	return nil, fmt.Errorf("failed to mine contract deploy tx %s into a block after %s. Aborting", signedTx.Hash(), time.Since(start))
```
BedrockSquirrel, line 224: I think the idiomatic Go style tries to keep the happy path as the final return but this is a weird one and I don't feel strongly.

BedrockSquirrel (commented):
few thoughts

tools/contractdeployer/cli.go
joeldudley: Added.

tools/contractdeployer/contract_deployer.go
joeldudley: Left in by accident. Removed.

tools/contractdeployer/contract_deployer.go
joeldudley: Cool. I'll leave it for now

BedrockSquirrel (approved):
LGTM

PR #584, opened 2022-08-19
rpc.BlockNumber now accepts latests and earliest strings

otherview:

- getCode was not working as expected, latest block was not fetchable
- part of https://github.com/obscuronet/obscuro-internal/issues/596

- Latest and earliest is now acceptable

### DIFF
--- a/go/host/rpc/clientapi/client_api_eth.go
+++ b/go/host/rpc/clientapi/client_api_eth.go
@@ -6,12 +6,12 @@ import (
 	"fmt"
 	"math/big"
 
-	"github.com/obscuronet/go-obscuro/go/host"
-
-	gethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/rpc"
 	"github.com/obscuronet/go-obscuro/go/common"
+	"github.com/obscuronet/go-obscuro/go/host"
+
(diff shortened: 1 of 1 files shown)


go/host/rpc/clientapi/client_api_eth.go
```diff
@@ -112,7 +97,7 @@ func (api *EthereumAPI) GetTransactionReceipt(_ context.Context, encryptedParams
 // EstimateGas is a placeholder for an RPC method required by MetaMask/Remix.
 func (api *EthereumAPI) EstimateGas(_ context.Context, _ interface{}, _ *rpc.BlockNumberOrHash) (hexutil.Uint64, error) {
 	// TODO - Return a non-dummy gas estimate.
-	return 0, nil
+	return 5_000_000_000, nil
```
joeldudley, line 100: @otherview Can we revert this one for now?

joeldudley (changes requested):
Just don't want to muddle up the change to gas estimation with this.

go/host/rpc/clientapi/client_api_eth.go
otherview: This is actually a pain point.

Having 0 means I need to find all Tx calls when using hardhat and manually hardcode them to some gas.
At the same time I must do the same when doing transactions with MetaMask.

I believe it's better to have it set at a high number rather than 0.
At that's the experience I'm having with HH and MM. WDYT ?

joeldudley (approved):

PR #587, opened 2022-08-19
Assert for initialized pointer in rpc client Call

BedrockSquirrel:

- common cause of confusing errors during development is not giving the correct result pointer to encRPCClient.Call()

- perform assert on the result to make sure:
1. It is a pointer
2. It is initialized


joeldudley (approved):
Lgtm

PR #590, opened 2022-08-19
Waits for host shutdown, instead of spinning off goroutine.

joeldudley:

It's clear from CI that hosts are often not being shut down properly.

Functional.

- Previously, host shutdown was run on an unawaited goroutine, so shutdown would end as soon as the process exited

- ...


otherview (approved):

otherview:
Is it worth looking for more places where we fire-and-forget shutdown procedures ?

PR #593, opened 2022-08-22
Fixes node shutdown.

joeldudley:

Previously, we had to perform the node RPC server shutdown in a goroutine, and we weren't sure why.

Functional.

- Adds explanatory comment explaining why using a goroutine is required
- Waits for host RPC ports to free up, to ensure the RPC server has actually stopped
- Panics instead of logging an error if the ports aren't freed, to highlight issues in the future

- ...

### DIFF
--- a/integration/gethnetwork/geth_network.go
+++ b/integration/gethnetwork/geth_network.go
@@ -276,7 +276,7 @@ func (network *GethNetwork) StopNodes() {
 	for idx, process := range network.nodesProcs {
 		if process != nil {
 			wg.Add(1)
-			go func(process *os.Process) {
+			go func(process *os.Process, nodeNumber int) {
 				defer wg.Done()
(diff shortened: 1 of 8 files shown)


integration/gethnetwork/geth_network.go
```diff
@@ -276,7 +276,7 @@ func (network *GethNetwork) StopNodes() {
 	for idx, process := range network.nodesProcs {
 		if process != nil {
 			wg.Add(1)
-			go func(process *os.Process) {
+			go func(process *os.Process, nodeNumber int) {
```
joeldudley, line 279: Fixed a bug here, where the node number would be updated in the for loop, so it would look like the final node stopped a bunch of times.

BedrockSquirrel (approved):
LGTM

PR #594, opened 2022-08-22
Add context support to EncClient

BedrockSquirrel:

- we want our Go ObscuroClient to support contexts like the geth client does, so we need to add support on the EncClient for it

- add CallContext methods through the rpc clients (not used by anything yet)

- is the EncClient implementation clear with its `internalCall()`? I wanted to re-use the Call() logic, maybe there's a neater way

### DIFF
--- a/go/rpcclientlib/encrypted_client.go
+++ b/go/rpcclientlib/encrypted_client.go
@@ -18,8 +18,6 @@ const (
 	enclavePublicKeyHex = "034d3b7e63a8bcd532ee3d1d6ecad9d67fca7821981a044551f0f0cbec74d0bc5e"
 )
 
-type callExecutor func(result interface{}, method string, args ...interface{}) error
-
 // for these methods, the RPC method's requests and responses should be encrypted
 var sensitiveMethods = []string{RPCCall, RPCGetBalance, RPCGetTxReceipt, RPCSendRawTransaction, RPCGetTransactionByHash}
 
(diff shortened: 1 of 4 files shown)


go/rpcclientlib/encrypted_client.go
```diff
@@ -49,14 +52,28 @@ func NewEncRPCClient(client Client, viewingKey *ViewingKey) (*EncRPCClient, erro
...
+// callInternal is the main logic to execute JSON-RPC requests:
+// - if the method is sensitive it will encrypt the args before sending the request and then decrypts the response before returning
+// - result must be a pointer so that package json can unmarshal into it. You can also pass nil, in which case the result is ignored.
+// - callExec handles the delegated call, allows EncClient to use the same code for calling with or without a context
+func (c *EncRPCClient) callInternal(callExec callExecutor, result interface{}, method string, args ...interface{}) error {
 	assertResultIsPointer(result)
 	if !isSensitive(method) {
 		// for non-sensitive methods or when viewing keys are disabled we just delegate directly to the geth RPC client
-		return c.obscuroClient.Call(result, method, args...)
+		return callExec(result, method, args...)
```
joeldudley, line 76: > “Your scientists were so preoccupied with whether or not they could, they didn't stop to think if they should.”

I think this is pretty wacky tbh. Can't you just pass a context object into `callInternal`, then have an if-statement where if the context is nil, you use `Call`, and otherwise you use `CallContext`? Seems like it would be much easier to follow.

joeldudley (approved):
LGTM

PR #598, opened 2022-08-22
Changes to allow Uniswap tooling to work

otherview:

- Changes to allow Uniswap tooling to work

- Allows for no-account calls
- fixed gasprice reply of 1
- reduced default log level


tools/walletextension/multi_acc_helper.go
```diff
@@ -142,30 +142,39 @@ func parseParams(args []interface{}) (map[string]interface{}, error) {
...
 			}
 		}
-	}
+		// every attempt errored
+		return err
 
-	return err
+	default: // no clients registered, use a temporary one
+		// TODO: We might consider having a permanent, no-account client to avoid the overhead of connection setup
+		client, err := rpcclientlib.NewNetworkClient(we.hostAddr)
```
BedrockSquirrel, line 171: yeah I'd say move this to the WE initialisation, put it on the struct as `unauthClient` or something.

tools/walletextension/multi_acc_helper.go
```diff
@@ -142,30 +142,39 @@ func parseParams(args []interface{}) (map[string]interface{}, error) {
...
+		if err != nil {
+			return fmt.Errorf("unable to create temporary client for request - %w", err)
+		}
+
```
BedrockSquirrel, line 175: I think it would be better to check whether this is going to fail before calling the host, like:

```
if isSensitive(method) {
    return Error("method %s cannot be called with an unauthorised client but no signed viewing keys found")
}
```

You'll need to make the `sensitiveMethods` list public in the `rpcclientlib/client.go` to make that isSensitive func.

go/enclave/bridge/bridge.go
```diff
@@ -163,7 +163,7 @@ func (bridge *Bridge) ExtractRollups(b *types.Block, blockResolver db.BlockResol
 			// In case of L1 reorgs, rollups may end published on a fork
 			if blockResolver.IsBlockAncestor(b, r.Header.L1Proof) {
 				rollups = append(rollups, obscurocore.ToEnclaveRollup(r, bridge.TransactionBlobCrypto))
-				common.LogWithID(bridge.NodeID, "Extracted Rollup r_%d from block b_%d",
+				common.TraceWithID(bridge.NodeID, "Extracted Rollup r_%d from block b_%d",
```
BedrockSquirrel, line 166: Do we have trace level logging in DataDog atm? If not we should probably check with Tudor that it isn't one of his go-to debugging log msgs.

BedrockSquirrel (commented):
Few minor thoughts

tools/walletextension/multi_acc_helper.go
otherview: Done 👍

tools/walletextension/multi_acc_helper.go
otherview: Yep, makes sense 👍

go/enclave/bridge/bridge.go
otherview: Good point, checked with Tudor, looks good 👍

BedrockSquirrel (commented):
I am an idiot, this was sat there as "Pending"

tools/walletextension/wallet_extension.go
```diff
@@ -64,6 +64,7 @@ type WalletExtension struct {
 	accountClients   map[common.Address]*rpcclientlib.EncRPCClient // an encrypted RPC client per registered account
 	unsignedVKs      map[common.Address]*rpcclientlib.ViewingKey   // map temporarily holding VKs that have been generated but not yet signed
 	server           *http.Server
+	unAuthedClient   rpcclientlib.Client // The default, un-authenticated client.
```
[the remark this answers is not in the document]
otherview: Good point, what about `nonAuthedClient` ?

BedrockSquirrel (approved):
LGTM

tools/walletextension/wallet_extension.go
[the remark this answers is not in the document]
otherview: Back to `unauthedClient` after chat :)

PR #600, opened 2022-08-24
Prep work for enabling L1 tx broadcast retries

joeldudley:

Currently, we panic if a transaction cannot be sent to the L1 node on the first try. Instead, we should have retries, and surface an error if all retries fail instead of panicking.

This is prep work ahead of actually implementing the retries.

Functional.

- Renames the L1 connection config flag, to also cover RPC timeouts
- Adds an error return from `a.broadcastL1Tx`, and adjusts the code accordingly
- Surfaces errors instead of panicking in a few other host methods

- ...


go/host/node/node.go
```diff
@@ -178,27 +178,15 @@ func (a *Node) Start() {
 
 	// todo: we should try to recover the key from a previous run of the node here? Before generating or requesting the key.
 	if a.config.IsGenesis {
-		common.LogWithID(a.shortID, "Node is genesis node. Broadcasting secret.")
```
joeldudley, line 181: This logic still exists, but is encapsulated in `a.brodcastSecret` now.

BedrockSquirrel (approved):
LGTM

PR #603, opened 2022-08-24
Avoids race condition in Geth shutdown.

joeldudley:

There was a race condition in the Geth start-up code (I think).

We needed to wait for all the miner-creation goroutines to return before shutting down the nodes if there were any errors. Otherwise, there's a possible race condition whereby if the creation of one node fails, we start shutting down the nodes from that goroutine, but new nodes are being spun up on other goroutines, and these are missed by the shutdown process.

Refactoring.

- Avoids race condition

- ...


BedrockSquirrel (approved):
LGTM

PR #607, opened 2022-08-26
Adds an appendix to the events design.

joeldudley:

This is useful information that will guide our implementation of Obscuro events.


BedrockSquirrel (approved):
LGTM

PR #617, opened 2022-08-31
Removes host method only used by tests

joeldudley:

We have a `MockHost` interface that is implemented by the host, and exposes methods required by the in-memory integration tests. Where possible, we should be removing these methods, so that we're testing the "real" host API.

Refactoring.

- Removes the `MockHost.P2P` method

- ...

### DIFF
--- a/integration/simulation/network/inmemory.go
+++ b/integration/simulation/network/inmemory.go
@@ -43,6 +43,11 @@ func (n *basicNetworkOfInMemoryNodes) Create(params *params.SimParams, stats *st
 	dummyETHAddress := common.HexToAddress("BB")
 	params.Wallets.Tokens[bridge.POC].L1ContractAddress = &dummyETHAddress
 
+	p2pLayers := make([]*p2p.MockP2P, params.NumberOfNodes)
+	for i := 0; i < params.NumberOfNodes; i++ {
+		p2pLayers[i] = p2p.NewMockP2P(params.AvgBlockDuration, params.AvgGossipPeriod)
+	}
+
 	for i := 0; i < params.NumberOfNodes; i++ {
 		isGenesis := i == 0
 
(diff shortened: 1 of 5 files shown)


integration/simulation/network/inmemory.go
```diff
@@ -36,6 +36,7 @@ func (n *basicNetworkOfInMemoryNodes) Create(params *params.SimParams, stats *st
 	n.ethNodes = make([]*ethereum_mock.Node, params.NumberOfNodes)
 	obscuroNodes := make([]host.MockHost, params.NumberOfNodes)
 	n.obscuroClients = make([]rpcclientlib.Client, params.NumberOfNodes)
+	p2pLayers := make([]*p2p.MockP2P, params.NumberOfNodes)
```
BedrockSquirrel, line 39: I don't really understand the name `layers`, not sure what the word should be though, something like p2pConnections or p2pAccessor perhaps?

BedrockSquirrel (commented):
Minor naming confusion

integration/simulation/network/inmemory.go
joeldudley: I don't want connections, as these aren't connections, they're ways to make the connections.

Think accessors is too meaningless.

I used layers because this is the P2P layer of the node that's used for communication.

BedrockSquirrel: Fair, maybe it's the pluralisation throwing me. I'm picturing layers and I'm picturing p2p connections and struggling to marry those images up in mind but it's just one 'layer' per node.

BedrockSquirrel (approved):
LGTM

integration/simulation/network/inmemory.go
joeldudley: Let's call it `p2psLayer` instead :D

BedrockSquirrel: Don't hate the p2pLayer, hate the p2gAme

PR #621, opened 2022-08-31
Puts the Docker and Geth sims behind an environment variable.

joeldudley:

The CI times are very long. A large contributor is the large number of simulations we run.

Since we now have the test-obscuro tests and a running testnet, we no longer need as many sims to gain assurance that everything is working fine.

Functional.

- Moves the Docker and Geth sims behind an environment flag

- ...


BedrockSquirrel (approved):
lgtm

PR #624, opened 2022-09-01
Bridge code should use big.Int for transfer quantities in wei

BedrockSquirrel:

The bridge code was using int64's for quantity in bridge deposits/withdrawals, but ERC20s denominated in 10^-18 "wei" units need to support bigger quantities.

- Change bridge/withdraw/deposit code to use big.Int instead of uint64 for quantities
- Update the network validation tests with appropriate units


go/common/rpc/converters.go
```diff
@@ -69,7 +69,7 @@ func ToRollupHeaderMsg(header *common.Header) *generated.HeaderMsg {
 	var headerMsg generated.HeaderMsg
 	withdrawalMsgs := make([]*generated.WithdrawalMsg, 0)
 	for _, withdrawal := range header.Withdrawals {
-		withdrawalMsg := generated.WithdrawalMsg{Amount: withdrawal.Amount, Recipient: withdrawal.Recipient.Bytes(), Contract: withdrawal.Contract.Bytes()}
+		withdrawalMsg := generated.WithdrawalMsg{Amount: withdrawal.Amount.Bytes(), Recipient: withdrawal.Recipient.Bytes(), Contract: withdrawal.Contract.Bytes()}
```
BedrockSquirrel, line 72: protobuf doesn't support big.Int directly but we can easily read/write big.Ints from a protobuf `bytes` field.

go/ethadapter/erc20contractlib/erc20_contract_lib.go
```diff
@@ -54,7 +54,7 @@ func NewERC20ContractLib(mgmtContractAddr *gethcommon.Address, contractAddrs ...
 }
 
 func (c *erc20ContractLibImpl) CreateDepositTx(tx *ethadapter.L1DepositTx, nonce uint64) types.TxData {
-	data, err := c.contractABI.Pack("transfer", &tx.To, big.NewInt(int64(tx.Amount)))
+	data, err := c.contractABI.Pack("transfer", &tx.To, tx.Amount)
```
otherview, line 57: It's fine to pack big.Int, yes ?

BedrockSquirrel: More than fine, it's required haha. See how before the int64 value was getting converted to big.Int for packing

integration/datagenerator/common.go
```diff
@@ -5,6 +5,8 @@ import (
 	"math"
 	"math/big"
 
+	intcommon "github.com/obscuronet/go-obscuro/integration/common"
```
otherview, line 8: I'd move this to the go/common, but it's more of a preference.

integration/simulation/simulation.go
```diff
@@ -27,10 +27,11 @@ import (
 )
 
 const (
-	initialBalance  = 5000
 	allocObsWallets = 750000000000000 // The amount the faucet allocates to each Obscuro wallet.
 )
 
+var initialBalance = big.NewInt(5000*10 ^ 18) // balance in wei (1 HOC = 10^18 HOC-wei)
```
otherview, line 33: Should the converter be used here ?

integration/simulation/validate_chain.go
```diff
@@ -353,7 +358,7 @@ func extractWithdrawals(t *testing.T, nodeClient rpc.Client, nodeAddr uint64) (t
 			return
 		}
 		for _, w := range r.Withdrawals {
-			totalSuccessfullyWithdrawn += w.Amount
+			totalSuccessfullyWithdrawn.Add(totalSuccessfullyWithdrawn, w.Amount)
```
otherview, line 361: Noob question, but don't you have to assign it here ?
`totalSuccessfullyWithdrawn = totalSuccessfullyWithdrawn.Add(totalSuccessfullyWithdrawn, w.Amount)` ?

otherview (commented):
Looks good, just a few comments before merging I guess

integration/simulation/validate_chain.go
BedrockSquirrel: Apparently not and seems they recommend not to

otherview (approved):

PR #627, opened 2022-09-05
Remove address param from AuthClient's NonceAt and BalanceAt

BedrockSquirrel:

- Per Joel's suggestion it makes this client a bit more coherent. The client is bound to a single account, it should populate the address in requests (and not allow addresses to be passed in that can never be successful).

- store account on AuthObsClient
- remove address from NonceAt method
- add BalanceAt method, also without an address field

- My main concern was wondering if we'd ever support Nonce/Balance lookups for contracts from other accounts? In which case we'd need that option. But I think current design means we won't be able to do that (contract devs would need to expose it through methods if they wanted to, and that could be permissioned)

### DIFF
--- a/go/obsclient/authclient.go
+++ b/go/obsclient/authclient.go
@@ -19,7 +19,7 @@ import (
 // The methods in this client are analogous to the methods in geth's EthClient and should behave the same unless noted otherwise.
 type AuthObsClient struct {
 	ObsClient
-	Account *common.Address
+	account common.Address
 }
 
 // NewAuthObsClient constructs an AuthObsClient for sensitive communication with an enclave.
@@ -29,9 +29,9 @@ type AuthObsClient struct {
 func NewAuthObsClient(client *rpc.EncRPCClient) *AuthObsClient {
 	return &AuthObsClient{
 		ObsClient: ObsClient{
-			RPCClient: client,
+			rpcClient: client,
 		},
-		Account: client.Account(),
+		account: *client.Account(),
 	}
 }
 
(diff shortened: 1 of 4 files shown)


go/obsclient/authclient.go
```diff
@@ -19,6 +19,7 @@ import (
 // The methods in this client are analogous to the methods in geth's EthClient and should behave the same unless noted otherwise.
 type AuthObsClient struct {
 	ObsClient
+	Account *common.Address
```
joeldudley, line 22: @BedrockSquirrel Can't this be private?

go/obsclient/authclient.go
```diff
@@ -30,6 +31,7 @@ func NewAuthObsClient(client *rpc.EncRPCClient) *AuthObsClient {
 		ObsClient: ObsClient{
 			RPCClient: client,
```
joeldudley, line 32: @BedrockSquirrel I think this field can be private too (if you use the `NewObsClient` constructor in the tests)?

go/obsclient/authclient.go
```diff
@@ -62,9 +64,11 @@ func (ac *AuthObsClient) TransactionReceipt(ctx context.Context, txHash common.H
 	return &receipt, err
 }
 
-func (ac *AuthObsClient) NonceAt(ctx context.Context, account common.Address, blockNumber *big.Int) (uint64, error) {
+// NonceAt retrieves the nonce for the account registered on this client (due to obscuro privacy restrictions,
+// nonce cannot be requested for other accounts)
+func (ac *AuthObsClient) NonceAt(ctx context.Context, blockNumber *big.Int) (uint64, error) {
 	var result hexutil.Uint64
-	err := ac.RPCClient.CallContext(ctx, &result, rpc.RPCNonce, account, toBlockNumArg(blockNumber))
+	err := ac.RPCClient.CallContext(ctx, &result, rpc.RPCNonce, ac.Account.Hex(), toBlockNumArg(blockNumber))
```
joeldudley, line 71: @BedrockSquirrel Why have we switched from passing a `common.Address` to a `string` here?

go/obsclient/authclient.go
```diff
@@ -77,3 +81,15 @@ func (ac *AuthObsClient) CallContract(ctx context.Context, msg ethereum.CallMsg,
 func (ac *AuthObsClient) SendTransaction(ctx context.Context, signedTx *types.Transaction) error {
 	return ac.RPCClient.CallContext(ctx, nil, rpc.RPCSendRawTransaction, encodeTx(signedTx))
 }
+
+// BalanceAt retrieves the native balance for the account registered on this client (due to obscuro privacy restrictions,
+// balance cannot be requested for other accounts)
+func (ac *AuthObsClient) BalanceAt(ctx context.Context, blockNumber *big.Int) (*big.Int, error) {
+	var result string
+	err := ac.RPCClient.CallContext(ctx, &result, rpc.RPCGetBalance, ac.Account.Hex(), toBlockNumArg(blockNumber))
```
joeldudley, line 89: @BedrockSquirrel Same q as above re: conversion of account to hex.

joeldudley (approved):
Thanks for fixing that up.

PR #628, opened 2022-09-05
Fixes nil point except

otherview:

- Found when working on https://github.com/obscuronet/obscuro-internal/issues/911
- If the `acc` is nil, then the wallet blows up with nil point exception
- Found when runnign the following curl cmd and pointing at testnet
```
curl --location --request POST 'http://127.0.0.1:3001' \
--header 'Content-Type: application/json' \
--data-raw '{
    "jsonrpc":"2.0",
    "method":"eth_call",
    "params":[{
        "from": "0x38a47fe0c945900d2b29db7edae05178cdba3d15",
        "to": null,
        "gas": "null",
        "gasPrice": "null",
        "value": "0x3635c9adc5dea00000",
        "data": ""
    }, "latest"],
    "id":1
}'
```

### DIFF
--- a/tools/walletextension/multi_acc_helper.go
+++ b/tools/walletextension/multi_acc_helper.go
@@ -46,7 +46,7 @@ func suggestAccountClient(req *rpcRequest, accClients map[common.Address]*rpc.En
 	if req.method == rpc.RPCCall {
 		// Otherwise, we search the `data` field for an address matching a registered viewing key.
 		addr, err := searchDataFieldForAccount(paramsMap, accClients)
-		if err == nil {
+		if err == nil && addr != nil {
 			return accClients[*addr]
 		}
 	}
--- a/tools/walletextension/multi_acc_helper.go
+++ b/tools/walletextension/multi_acc_helper.go
@@ -46,7 +46,7 @@ func suggestAccountClient(req *rpcRequest, accClients map[common.Address]*rpc.En
 	if req.method == rpc.RPCCall {
 		// Otherwise, we search the `data` field for an address matching a registered viewing key.
 		addr, err := searchDataFieldForAccount(paramsMap, accClients)
-		if err == nil && addr != nil {
+		if err == nil {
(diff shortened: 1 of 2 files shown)


tools/walletextension/multi_acc_helper.go
```diff
@@ -46,7 +46,7 @@ func suggestAccountClient(req *rpcRequest, accClients map[common.Address]*rpc.En
 	if req.method == rpc.RPCCall {
 		// Otherwise, we search the `data` field for an address matching a registered viewing key.
 		addr, err := searchDataFieldForAccount(paramsMap, accClients)
-		if err == nil {
+		if err == nil && addr != nil {
```
BedrockSquirrel, line 49: I don't think this is quite the right solution, instead `searchDataFieldForAccount` should never return `nil,nil` because that breaks the Go convention. If there was no Account then it's an error imo.

I would change line 92 to return an error instead of nil,nil

BedrockSquirrel (changes requested):
Think we should have errored rather than allow a nil address with no error

BedrockSquirrel (approved):
LGTM

tools/walletextension/multi_acc_helper.go
otherview: Yes, I have a pet peeve with Not Found being an error.  But I don't think it applies here.

Good suggestion, done 👍

tools/walletextension/multi_acc_helper_test.go
```diff
@@ -58,8 +58,8 @@ func TestErrorsWhenDataFieldIsMissing(t *testing.T) {
 func TestGracefulWhenDataFieldTooShort(t *testing.T) {
```
joeldudley, line 58: @otherview The test name needs to change - it's the opposite of graceful now.

otherview: Fixed 👍

joeldudley (approved):

PR #629, opened 2022-09-05
Store logs for each block to not have to recalculate them

joeldudley:

Currently, we recalculate the logs for the entire chain each time we get a new block. This isn't sustainable.

Refactoring.

- Pushes the logs for each block to the DB
- Retrieves the logs for each block from the DB

- ...

### DIFF
--- a/go/enclave/db/storage.go
+++ b/go/enclave/db/storage.go
@@ -225,7 +225,7 @@ func (s *storageImpl) FetchBlockState(hash common.L1RootHash) (*core.BlockState,
 	return nil, nil, false
 }
 
-func (s *storageImpl) SaveNewHead(state *core.BlockState, rollup *core.Rollup, receipts []*types.Receipt) {
+func (s *storageImpl) SaveNewHead(state *core.BlockState, rollup *core.Rollup, receipts []*types.Receipt, logs map[uuid.UUID][]*types.Log) {
 	batch := s.db.NewBatch()
 
 	if state.FoundNewRollup {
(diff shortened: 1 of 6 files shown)


go/enclave/db/storage.go
```diff
@@ -213,15 +215,17 @@ func (s *storageImpl) Proof(r *core.Rollup) *types.Block {
 	return v
 }
 
-func (s *storageImpl) FetchBlockState(hash common.L1RootHash) (*core.BlockState, bool) {
+func (s *storageImpl) FetchBlockState(hash common.L1RootHash) (*core.BlockState, map[uuid.UUID][]*types.Log, bool) {
 	bs := obscurorawdb.ReadBlockState(s.db, hash)
-	if bs != nil {
-		return bs, true
+	logs := obscurorawdb.ReadBlockLogs(s.db, hash)
```
BedrockSquirrel, line 220: I guess at some point this thing would pop the memory, if it can grown unbounded in two dimensions (number of subscriptions open and log accumulation over time).

I guess we'll just have to put a reasonable cap on number of open subscriptions, and have a chat with Tudor about how many rollups back we keep logs for.

go/enclave/rollupchain/rollup_chain.go
```diff
@@ -270,6 +261,8 @@ func (rc *RollupChain) updateState(b *types.Block) (*obscurocore.BlockState, map
 		subscribedLogs[subscriptionID] = logsForID
 	}
 
+	rc.storage.SaveNewHead(bs, head, receipts, subscribedLogs)
```
BedrockSquirrel, line 264: Feels like we should store the logs *before* we append the parent logs but I guess not if we want the full growing history.

Trying to convince myself this never results in us accidentally appending the same ancestor logs multiple times in the same log history, and think it's fine but one to watch out for if it spirals out of control faster than expected maybe.

BedrockSquirrel (approved):
I think it's fine? Couple of comments for concerns about things growing out of control but think it's all sensible in what it's storing, we can just keep an eye on it.

Ancestors would be store multiple times in the db which isn't v efficient but never multiple times against the same hash I think so that's just for lookup-ability.

PR #639, opened 2022-09-08
Switches network client to websockets.

joeldudley:

Currently, our network client uses HTTP. This does not allow subscriptions, which we need for logs.

- Switches the network client to using WS
- Fixes tests as needed


go/config/host_config.go
```diff
@@ -72,7 +72,7 @@ func DefaultHostConfig() HostConfig {
 		GossipRoundDuration:    8333,
 		HasClientRPCHTTP:       true,
 		ClientRPCPortHTTP:      13000,
-		HasClientRPCWebsockets: false,
+		HasClientRPCWebsockets: true,
```
BedrockSquirrel, line 75: Am I right in thinking this is already enabled in all our testnets etc. because the config parsing code hardcodes it to true?

BedrockSquirrel (commented):
LGTM - just want to check that we're not going to impact anyone with this change

go/config/host_config.go
joeldudley: I think it's not enabled atm, but will become enabled due to this default.

BedrockSquirrel: So after this change goes out the WalletExtension on main won't work against Testnet until we release next week? At the very least worth checking if that's going to disrupt people probably.

Would it help to hardcode it in the docker YAML files? I guess probably not because they'll go out the same time as these changes anyway.

joeldudley: Yeah, that's correct. Not sure how we should approach this - just let ppl know? In general, breaking changes are fine between Testnet releases, right?

joeldudley: Actually, provided the WE is configured correctly, it should be fine. The WE was always supposed to be configured with an HTTP and a WS address, we've just switched to using the latter AFAIK.

BedrockSquirrel (approved):
LGTM - just have to keep everyone in the loop about client issues against old versions

PR #643, opened 2022-09-08
Update default RPC ports to use WS

BedrockSquirrel:

- Some things have hardcoded URLs/ports for the RPC connections that need updating after we switched the clients to be WS-only (this was noticed by Moray using the testnet L2 contract deployment script)

- Searched for '13000' in the code and updated the ones that were being used as RPC addresses rather than HTTP ports

- Did I change them correctly, did I miss any?


joeldudley (approved):
LGTM. I guess more may emerge as we deploy testnet, but this is a good start.

PR #644, opened 2022-09-08
Allows sub creation to be handled by the wallet extension.

joeldudley:

Creation of subscriptions is encrypted, and needs special handling via the wallet extension

- Adds a `Subscribe` method to all clients
- Implements `Subscribe` for the encrypted client
- Encrypts outbound subscription creation requests via the wallet extension
- Adds a very basic wallet extension test, showing that creating a subscription via the wallet extension doesn't blow up

### DIFF
--- a/go/common/types.go
+++ b/go/common/types.go
@@ -24,12 +24,12 @@ type (
 	EncryptedTx           []byte // A single transaction, encoded as a JSON list of transaction binary hexes and encrypted using the enclave's public key
 	EncryptedTransactions []byte // A blob of encrypted transactions, as they're stored in the rollup, with the nonce prepended.
 
-	EncryptedParamsGetBalance   []byte // The params for an RPC getBalance request, as a JSON object encrypted with the public key of the enclave.
-	EncryptedParamsCall         []byte // As above, but for an RPC call request.
-	EncryptedParamsGetTxByHash  []byte // As above, but for an RPC getTransactionByHash request.
-	EncryptedParamsGetTxReceipt []byte // As above, but for an RPC getTransactionReceipt request.
-	EncryptedParamsLogs         []byte // As above, but for an RPC logs subscription request.
-	EncryptedParamsSendRawTx    []byte // As above, but for an RPC sendRawTransaction request.
+	EncryptedParamsGetBalance      []byte // The params for an RPC getBalance request, as a JSON object encrypted with the public key of the enclave.
+	EncryptedParamsCall            []byte // As above, but for an RPC call request.
+	EncryptedParamsGetTxByHash     []byte // As above, but for an RPC getTransactionByHash request.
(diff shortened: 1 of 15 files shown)


integration/walletextension/wallet_extension_test.go
```diff
@@ -385,7 +399,7 @@ func waitForWalletExtension(walletExtensionAddr string) error {
 }
 
 // Makes an Ethereum JSON RPC request and returns the response body.
-func makeEthJSONReq(walletExtensionAddr string, method string, params interface{}) []byte {
```
joeldudley, line 388: This first param wasn't needed, because we use a file constant for the WE address.

tools/walletextension/multi_acc_helper.go
```diff
@@ -175,11 +176,27 @@ func proxyRequest(rpcReq *rpcRequest, rpcResp *interface{}, we *WalletExtension)
 }
 
 func executeCall(client *rpc.EncRPCClient, req *rpcRequest, resp *interface{}) error {
-	var err error
+	if req.method == rpc.RPCSubscribe {
+		if len(req.params) == 0 {
```
BedrockSquirrel, line 180: This branch is completely different from the rest of the executeCall code, might be better to `return executeSubscribe(...)`  and move that block in there. And maybe even consider if that branch of Call vs Subscribe could have happened earlier. (Could maybe add an `execute()` method that just branches between `Call` and `Subscribe` and gets used from `proxyRequest`)

go/common/types.go
```diff
@@ -26,9 +26,10 @@ type (
 
 	EncryptedParamsGetBalance   []byte // The params for an RPC getBalance request, as a JSON object encrypted with the public key of the enclave.
 	EncryptedParamsCall         []byte // As above, but for an RPC call request.
+	EncryptedParamsGetTxByHash  []byte // As above, but for an RPC getTransactionByHash request.
 	EncryptedParamsGetTxReceipt []byte // As above, but for an RPC getTransactionReceipt request.
+	EncryptedParamsLogs         []byte // As above, but for an RPC logs subscription request.
```
BedrockSquirrel, line 31: Maybe `EncryptedParamsSubscribeLogs` or `EncryptedParamsLogsSubscription` - I know they're a bit long but it's useful when looking at these methods to be reminded it's a subscription and the the other Params types have a verb in them.

BedrockSquirrel (approved):
LGTM generally, couple of minor quibbles

BedrockSquirrel (changes requested):
Sorry didn't mean to approve straight away, worth discussing or addressing the comments

go/common/types.go
joeldudley: Fixed.

BedrockSquirrel (approved):
LGTM, thanks

PR #648, opened 2022-09-09
Wallet extension - introduce abstraction layer to allow both HTTP and WS connections to be handled

joeldudley:

Currently, the WE only handles HTTP connections. To unblock the events work, it also needs to handle WS connections.

- Introduce an abstraction layer, `ReadWriter`, that hides whether a HTTP or WS connection is being used
- Adds a check and associated test that you cannot subscribe using an HTTP connection

### DIFF
--- a/integration/walletextension/wallet_extension_test.go
+++ b/integration/walletextension/wallet_extension_test.go
@@ -6,6 +6,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"github.com/ethereum/go-ethereum/eth/filters"
 	"io"
 	"math/big"
 	"net/http"
@@ -13,8 +14,6 @@ import (
 	"strings"
 	"testing"
 	"time"
-
-	"github.com/ethereum/go-ethereum/eth/filters"
 
 	"github.com/obscuronet/go-obscuro/go/enclave/rollupchain"
 
@@ -56,6 +55,7 @@ const (
 	latestBlock       = "latest"
 	statusSuccess     = "0x1"
 	errInsecure       = "enclave could not respond securely to %s request"
+	errSubscribeFail  = "received an eth_subscribe request but the connection does not support subscriptions"
 
 	networkStartPort = integration.StartPortWalletExtensionTest + 1
 	nodeRPCHTTPPort  = integration.StartPortWalletExtensionTest + 1 + network.DefaultHostRPCHTTPOffset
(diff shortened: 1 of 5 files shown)


integration/walletextension/wallet_extension_test.go
```diff
@@ -300,31 +301,6 @@ func TestCannotSubmitTxFromAnotherAddressAfterSubmittingViewingKey(t *testing.T)
 	}
 }
 
-// TODO - #453 - Build out this test as we expand the subscription functionality.
-func TestCanSubscribeForLogs(t *testing.T) {
```
joeldudley, line 304: I removed this test for now, as it no longer worked (since attempts to subscribe via HTTP are now rejected, as they should be). I added the test below (subscribing over HTTP should fail), and I'll readd this test in a future PR, as part of the events work.

tools/walletextension/readwriter/read_writer.go
```diff
@@ -0,0 +1,102 @@
...
+func (h HTTPReadWriter) HandleError(msg string) {
+	logAndSendErr(h.resp, msg)
+}
+
+func (h HTTPReadWriter) SupportsSubscriptions() bool {
+	return false
+}
+
+func (w WSReadWriter) ReadRequest() ([]byte, error) {
+	panic("not implemented")
```
joeldudley, line 83: I'll implement the websocket reading/writing in a future PR as part of the events work.

integration/walletextension/wallet_extension_test.go
```diff
@@ -367,6 +343,16 @@ func TestCanDecryptSuccessfullyAfterRestartingWalletExtension(t *testing.T) {
 	}
 }
 
+func TestCannotSubscribeOverHTTP(t *testing.T) {
+	createWalletExtension(t)
+
+	respBody := makeEthJSONReq(rpc.RPCSubscribe, []interface{}{rpc.RPCSubscriptionTypeLogs, filters.FilterCriteria{}})
+
+	if !strings.Contains(string(respBody), errSubscribeFail) {
```
otherview, line 351: Should this use errors.is (I think that's the package name) ?

tools/walletextension/readwriter/read_writer.go
```diff
@@ -0,0 +1,102 @@
...
+			return &WSReadWriter{
+				conn: conn,
+			}, nil
+		}
+	}
+
+	return &HTTPReadWriter{resp: resp, req: req}, nil
+}
+
+func (h HTTPReadWriter) ReadRequest() ([]byte, error) {
```
otherview, line 58: Should these be instance methods ?
`func (h HTTPReadWriter) ReadRequest() ([]byte, error)` to `func (h *HTTPReadWriter) `

tools/walletextension/readwriter/read_writer.go
```diff
@@ -0,0 +1,102 @@
...
+
+func (w WSReadWriter) HandleError(msg string) {
+	panic("not implemented")
+}
+
+func (w WSReadWriter) SupportsSubscriptions() bool {
+	return true
+}
+
+func logAndSendErr(resp http.ResponseWriter, msg string) {
```
otherview, line 98: Should this be a specific http logAndSendErr ?

tools/walletextension/readwriter/read_writer.go
```diff
@@ -0,0 +1,102 @@
...
+	req  *http.Request
+}
+
+// WSReadWriter is a ReadWriter over websockets.
+type WSReadWriter struct {
+	conn *websocket.Conn
+}
+
+func NewReadWriter(resp http.ResponseWriter, req *http.Request) (ReadWriter, error) {
+	for _, header := range req.Header[headerUpgrade] {
```
otherview, line 41: Would it be useful to add some comments on the header upgrade operations?

otherview (commented):
Few comments but the abstraction looks good to me 👍

tools/walletextension/readwriter/read_writer.go
joeldudley: Yep, fixed.

tools/walletextension/readwriter/read_writer.go
joeldudley: Fixed.

integration/walletextension/wallet_extension_test.go
joeldudley: I don't want to expose the error from the `readwriter` package just to use it way over here in the tests. I don't forsee it being used outside of tests.

tools/walletextension/readwriter/read_writer.go
joeldudley: Yep, added.

integration/walletextension/wallet_extension_test.go
otherview: I think it's weird we do strings contain to check if it's an error. But I see your point, and agree that it doens't need to be imported.

otherview (approved):

PR #652, opened 2022-09-12
Adds WS port to wallet extension server

joeldudley:

Previously, we tried to handle both HTTP and WS connections on a single port. To be consistent with Geth, and to enable us to use lower-cost HTTP clients where possible, we should have two ports.

- Adds a port to the wallet extension dedicated to WS connections
- Adds some todos, about creating WS clients selectively and lazily


otherview (approved):
Not sure about the addition of the CORS, but everything else looks grand 👍

PR #655, opened 2022-09-12
Add change log entry for WebSockets

moraygrieve:

- Adding a change log entry for support with WebSockets

- The change log

- The change log entry - specifically can you update if we support concurrent HTTP and WS connections etc?


joeldudley (approved):
Thanks

PR #656, opened 2022-09-12
Does not hardcode use of WS or HTTP for Obscuro clients

joeldudley:

Using WS everywhere may hog too many resources.

- Does not add WS protocol to every attempt to dial the client
- Fix-ups in tests

### DIFF
--- a/go/rpc/network_client.go
+++ b/go/rpc/network_client.go
@@ -3,12 +3,14 @@ package rpc
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/ethereum/go-ethereum/rpc"
 )
 
 const (
-	ws = "ws://"
+	ws   = "ws://"
+	http = "http://"
 )
 
 // networkClient is a Client implementation that wraps Geth's rpc.Client to make calls to the obscuro node
@@ -31,10 +33,13 @@ func NewEncNetworkClient(rpcAddress string, viewingKey *ViewingKey) (*EncRPCClie
 
 // NewNetworkClient returns a client that can make RPC calls to an Obscuro node
 func NewNetworkClient(address string) (Client, error) {
-	// TODO - Allow instantiator to choose between websockets and HTTP
-	rpcClient, err := rpc.Dial(ws + address)
+	if !strings.HasPrefix(address, http) && !strings.HasPrefix(address, ws) {
+		return nil, fmt.Errorf("clients for Obscuro only support the %s and %s protocols", http, ws)
+	}
+
+	rpcClient, err := rpc.Dial(address)
 	if err != nil {
-		return nil, fmt.Errorf("could not create RPC client on %s. Cause: %w", ws+address, err)
+		return nil, fmt.Errorf("could not create RPC client on %s. Cause: %w", address, err)
 	}
 
 	return &networkClient{
--- a/integration/simulation/network/obscuro_node_utils.go
+++ b/integration/simulation/network/obscuro_node_utils.go
@@ -265,7 +265,7 @@ func isAddressAvailable(address string) bool {
 	if ln != nil {
 		err = ln.Close()
 		if err != nil {
-			log.Error("could not close listener when checking if address %w was available", address)
+			log.Error("could not close listener when checking if address %s was available", address)
 		}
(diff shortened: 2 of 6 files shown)


go/rpc/network_client.go
```diff
@@ -31,10 +33,13 @@ func NewEncNetworkClient(rpcAddress string, viewingKey *ViewingKey) (*EncRPCClie
 
 // NewNetworkClient returns a client that can make RPC calls to an Obscuro node
 func NewNetworkClient(address string) (Client, error) {
-	// TODO - Allow instantiator to choose between websockets and HTTP
-	rpcClient, err := rpc.Dial(ws + address)
+	if !strings.HasPrefix(address, http) && !strings.HasPrefix(address, ws) {
```
joeldudley, line 36: I added this check because if no protocol is provided, it attempts to default to ipc, so I thought it was best to catch that cases explicitly.

integration/simulation/network/obscuro_node_utils.go
```diff
@@ -248,12 +249,19 @@ func waitTimeout(wg *sync.WaitGroup, timeout time.Duration) bool {
...
-	if err != nil {
-		return false
+	// `net.Listen` requires us to strip the protocol, if it exists.
+	addressNoProtocol := address
+	splitAddress := strings.Split(address, "://")
+	if len(splitAddress) == 2 {
+		addressNoProtocol = splitAddress[1]
 	}
+
+	ln, err := net.Listen("tcp", addressNoProtocol)
```
otherview, line 259: Does this server ever get closed ?

tools/contractdeployer/obscuro_deployer.go
```diff
@@ -126,5 +126,5 @@ func (o *obscuroDeployer) TransactionReceipt(hash gethcommon.Hash) (*types.Recei
 }
 
 func getURL(cfg *Config) string {
-	return fmt.Sprintf("%s:%d", cfg.NodeHost, cfg.NodePort)
+	return fmt.Sprintf("ws://%s:%d", cfg.NodeHost, cfg.NodePort)
```
otherview, line 129: Why hardcoding to ws ?

integration/simulation/network/obscuro_node_utils.go
joeldudley: Well, that's the purpose of this method. It's here to check that the server closes (see the surrounding method - it gives up after 10 secs, with an err msg).

tools/contractdeployer/obscuro_deployer.go
joeldudley: For simplicity - everything currently is set up to point to the host's WS port. Since there will only be one contract deployer execution, there's no cost to keeping with WS.

go/rpc/network_client.go
```diff
@@ -3,12 +3,14 @@ package rpc
...
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/ethereum/go-ethereum/rpc"
 )
 
 const (
-	ws = "ws://"
+	ws   = "ws://"
```
otherview, line 12: How do you feel about https://github.com/uber-go/guide/blob/master/style.md#prefix-unexported-globals-with-_ ?

joeldudley: Fine by me, but would be hard to enforce (maybe we can get the linter to complain?). In any case, don't think it makes sense to fix it in this PR for a single constant.

integration/simulation/network/obscuro_node_utils.go
otherview: Ah, yeah I see. I didn't see how this was working.

That ignore of the close was throwing me out. Would be nice to log the `_ = ln.Close()` error if there is one.

joeldudley: Cool, fixed.

tools/contractdeployer/obscuro_deployer.go
otherview: Fair enough 👍

otherview (approved):
Would make those proposed changes, other than that looks good 👍

PR #657, opened 2022-09-12
WE: Splits out persistence into separate component

joeldudley:

The WE was getting very bloated. This change moves the persistence logic into a separate package.

- Moves the persistence logic into a separate package

### DIFF
--- a/tools/walletextension/multi_acc_helper.go
+++ b/tools/walletextension/multi_acc_helper.go
@@ -224,3 +224,23 @@ func executeCall(client *rpc.EncRPCClient, req *rpcRequest, resp *interface{}) e
 
 	return client.Call(resp, req.method, req.params...)
 }
+
+// The enclave requires the `from` field to be set so that it can encrypt the response, but sources like MetaMask often
+// don't set it. So we check whether it's present; if absent, we walk through the arguments in the request's `data`
+// field, and if any of the arguments match our viewing key address, we set the `from` field to that address.
+// TODO - Move this method into multi_acc_helper.go.
+func setCallFromFieldIfMissing(args []interface{}, account common.Address) ([]interface{}, error) {
+	callParams, err := parseParams(args)
+	if err != nil {
+		return nil, fmt.Errorf("could not parse eth_call params. Cause: %w", err)
+	}
+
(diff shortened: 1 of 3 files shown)


tools/walletextension/multi_acc_helper.go
```diff
@@ -224,3 +224,22 @@ func executeCall(client *rpc.EncRPCClient, req *rpcRequest, resp *interface{}) e
 
 	return client.Call(resp, req.method, req.params...)
 }
+
+// The enclave requires the `from` field to be set so that it can encrypt the response, but sources like MetaMask often
+// don't set it. So we check whether it's present; if absent, we walk through the arguments in the request's `data`
+// field, and if any of the arguments match our viewing key address, we set the `from` field to that address.
+func setCallFromFieldIfMissing(args []interface{}, account common.Address) ([]interface{}, error) {
```
joeldudley, line 231: This was previously in the main `wallet_extension.go` file, but was only used here.

tools/walletextension/persistence/persistence.go
```diff
@@ -0,0 +1,164 @@
...
+
+	record := []string{
+		p.hostAddr,
+		viewingKey.Account.Hex(),
+		// We encode the bytes as hex to ensure there are no unintentional line breaks to make parsing the file harder.
+		hex.EncodeToString(viewingPrivateKeyBytes),
+		hex.EncodeToString(viewingKey.SignedKey),
+	}
+
+	persistenceFile, err := os.OpenFile(p.filePath, os.O_APPEND|os.O_WRONLY, 0o644)
```
moraygrieve, line 73: Does this lock the file so that another WE process would need to wait until the lock is released before it wrote to the file etc?

moraygrieve (approved):
LGTM - just the one question around file locking, but that will probably be addressed later I think?

tools/walletextension/persistence/persistence.go
joeldudley: AFAIK, based on my understanding of `O_APPEND`, each line written should be atomic. Should there's no risk of lines being overwritten or garbled.

PR #665, opened 2022-09-14
WE: Return subscribed logs over websockets

joeldudley:

We'd already plumbed in the return of logs from the enclave to the host to the WE. The final step was to return the logs from the WE to the user.

- Sends logs onwards from the WE to the user
- Adds integration test of logs being received

### DIFF
--- a/integration/walletextension/wallet_extension_test.go
+++ b/integration/walletextension/wallet_extension_test.go
@@ -53,15 +53,16 @@ const (
 	testLogs     = "../.build/wallet_extension/"
 	l2ChainIDHex = "0x309"
 
-	reqJSONKeyTo        = "to"
-	reqJSONKeyFrom      = "from"
-	reqJSONKeyData      = "data"
-	respJSONKeyStatus   = "status"
-	latestBlock         = "latest"
-	statusSuccess       = "0x1"
-	errInsecure         = "enclave could not respond securely to %s request"
-	errSubscribeFail    = "received an eth_subscribe request but the connection does not support subscriptions"
-	errInvalidRPCMethod = "rpc request failed: the method %s does not exist/is not available"
+	reqJSONKeyTo            = "to"
+	reqJSONKeyFrom          = "from"
+	reqJSONKeyData          = "data"
+	respJSONKeyStatus       = "status"
+	respJSONKeyContractAddr = "contractAddress"
+	latestBlock             = "latest"
+	statusSuccess           = "0x1"
+	errInsecure             = "enclave could not respond securely to %s request"
+	errSubscribeFail        = "received an eth_subscribe request but the connection does not support subscriptions"
+	errInvalidRPCMethod     = "rpc request failed: the method %s does not exist/is not available"
 
 	walletExtensionPort   = int(integration.StartPortWalletExtensionTest)
 	walletExtensionPortWS = int(integration.StartPortWalletExtensionTest + 1)
@@ -350,16 +351,6 @@ func TestCanDecryptSuccessfullyAfterRestartingWalletExtension(t *testing.T) {
 	}
 }
 
-func TestCannotSubscribeOverHTTP(t *testing.T) {
-	createWalletExtension(t)
-
-	respBody := makeHTTPEthJSONReq(rpc.RPCSubscribe, []interface{}{rpc.RPCSubscriptionTypeLogs, filters.FilterCriteria{}})
-
-	if !strings.Contains(string(respBody), errSubscribeFail) {
-		t.Fatalf("Expected error message \"%s\", got \"%s\"", errSubscribeFail, respBody)
-	}
-}
-
 func TestCanMakeRequestOverWS(t *testing.T) {
 	createWalletExtension(t)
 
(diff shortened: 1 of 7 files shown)


integration/walletextension/wallet_extension_test.go
```diff
@@ -350,20 +351,10 @@ func TestCanDecryptSuccessfullyAfterRestartingWalletExtension(t *testing.T) {
 	}
 }
 
-func TestCannotSubscribeOverHTTP(t *testing.T) {
```
joeldudley, line 353: I left this test unchanged, but moved it down, so that the two subscription tests were grouped together.

integration/walletextension/wallet_extension_test.go
```diff
@@ -456,36 +501,42 @@ func makeHTTPEthJSONReq(method string, params interface{}) []byte {
 }
 
 // Makes an Ethereum JSON RPC request over websockets and returns the response body as JSON.
-func makeWSEthJSONReqAsJSON(method string, params interface{}) map[string]interface{} {
-	respBody := makeWSEthJSONReq(method, params)
-	return convertRespBodyToJSON(respBody)
+func makeWSEthJSONReqAsJSON(method string, params interface{}) (map[string]interface{}, *websocket.Conn) {
```
joeldudley, line 504: This returns a connection object now, so that we can listen for events.

integration/walletextension/wallet_extension_test.go
```diff
@@ -456,36 +501,42 @@ func makeHTTPEthJSONReq(method string, params interface{}) []byte {
...
 
 // Makes an Ethereum JSON RPC request over websockets and returns the response body.
-func makeWSEthJSONReq(method string, params interface{}) []byte {
+func makeWSEthJSONReq(method string, params interface{}) ([]byte, *websocket.Conn) {
 	conn, resp, err := websocket.DefaultDialer.Dial(walletExtensionAddrWS, nil)
 	if resp != nil && resp.Body != nil {
 		defer resp.Body.Close()
 	}
-	if conn != nil {
-		defer conn.Close()
```
joeldudley, line 471: Annoyingly, we can no longer just `defer` closing the connection, as we need it to stay open if nothing goes wrong so that we can listen for logs.

tools/walletextension/wallet_extension.go
```diff
@@ -223,17 +224,17 @@ func (we *WalletExtension) handleEthJSON(readWriter readwriter.ReadWriter) {
...
 	respMap := make(map[string]interface{})
-	respMap[resJSONKeyID] = rpcReq.ID
-	respMap[resJSONKeyRPCVer] = jsonrpc.Version
+	respMap[respJSONKeyID] = rpcReq.ID
+	respMap[respJSONKeyRPCVer] = jsonrpc.Version
 	respMap[RespJSONKeyResult] = rpcResp
 
 	// https://github.com/ethereum/EIPs/blob/master/EIPS/eip-658.md
 	// TODO fix this upstream on the decode
-	if result, found := respMap["result"]; found { //nolint
```
joeldudley, line 232: I took the opportunity in this PR to replace these hardcoded strings with constants.

tools/walletextension/readwriter/read_writer.go
```diff
@@ -79,9 +81,16 @@ func (h *HTTPReadWriter) SupportsSubscriptions() bool {
...
 }
 
+func (h *HTTPReadWriter) IsClosed() bool {
+	return false
+}
+
 func (w *WSReadWriter) ReadRequest() ([]byte, error) {
 	_, msg, err := w.conn.ReadMessage()
 	if err != nil {
+		if websocket.IsCloseError(err) {
```
joeldudley, line 91: The only way to determine if a websocket is closed is that you receive an "is closed" error when you try and read or write to it. So we check here, so we can update the `ReadWriter`'s internal state and allow others to react accordingly (in this case, cancel existing subscriptions).

tools/walletextension/accountmanager/account_manager.go
```diff
@@ -44,21 +47,21 @@ func (m *AccountManager) AddClient(address common.Address, client *rpc.EncRPCCli
 
 // ProxyRequest tries to identify the correct EncRPCClient to proxy the request to the Obscuro node, or it will attempt
 // the request with all clients until it succeeds
-func (m *AccountManager) ProxyRequest(rpcReq *RPCRequest, rpcResp *interface{}) error {
+func (m *AccountManager) ProxyRequest(rpcReq *RPCRequest, rpcResp *interface{}, readWriter readwriter.ReadWriter) error {
```
BedrockSquirrel, line 50: I'm finding `readWriter` a  bit confusingly vague, is it a standard name for this sort of thing?

It represents a WS or HTTP connection from a WalExt user I guess, so maybe could be called something like `userConn`, `walExtConn`.

tools/walletextension/accountmanager/account_manager.go
```diff
@@ -211,18 +214,36 @@ func executeSubscribe(client *rpc.EncRPCClient, req *RPCRequest, _ *interface{})
...
 			}
 		}
 	}()
 
-	// TODO - #453 - Unsubscribe the subscription if the websocket is closed.
+	// We periodically check if the websocket is closed, and terminate the subscription.
+	go func() {
+		for {
+			if readWriter.IsClosed() {
+				subscription.Unsubscribe()
```
BedrockSquirrel, line 242: I think you need to break out of the for loop here or you might leak the goroutine.

BedrockSquirrel (commented):
Couple thoughts

tools/walletextension/accountmanager/account_manager.go
joeldudley: Renamed to `UserConn`.

tools/walletextension/accountmanager/account_manager.go
joeldudley: Thanks, fixed :)

BedrockSquirrel (approved):
LGTM

PR #668, opened 2022-09-14
EstimateGas to run estimation on the enclave + As an encrypted endpoint

otherview:

- https://github.com/obscuronet/obscuro-internal/issues/929
- First iteration 
   -  Hooks the estimation to the enclave
   - Ensures the endpoint is encrypted
   - Added enclave tests

- Endpoint runs on the enclave
- Endpoint is encrypted


go/enclave/enclave_test.go
```diff
@@ -0,0 +1,225 @@
...
+	encryptedViewingKeyBytes, err := ecies.Encrypt(rand.Reader, _enclavePubKey, key.PublicKey, nil, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// add the VK to the enclave
+	return key, enclave.AddViewingKey(encryptedViewingKeyBytes, key.SignedKey)
+}
+
+// createRandomEnclave returns a new instance of the enclave with random values
```
BedrockSquirrel, line 212: Maybe call this stuff 'arbitrary' rather than 'random'. I was getting ready for a rant about deterministic tests haha.

integration/walletextension/wallet_extension_test.go
```diff
@@ -427,6 +429,25 @@ func TestCannotSubscribeOverHTTP(t *testing.T) {
 	}
 }
 
+func TestSuccessGasEstimate(t *testing.T) {
```
BedrockSquirrel, line 432: I feel like we need a place for integration tests of specific methods/scenarios but maybe not as part of this PR, we should think about it. I can see why it ended up in here even though it's not really testing wallet extension, because it feels cleaner to put it here than shoehorn it into the simulation tests.

BedrockSquirrel (approved):
LGTM, couple of minor comments/thoughts. Need to add `EstimateGas()` to AuthClient too but no rush, there's plenty of other stuff not in there yet.

integration/walletextension/wallet_extension_test.go
```diff
@@ -427,6 +429,25 @@ func TestCannotSubscribeOverHTTP(t *testing.T) {
...
+
+	callMsgBytes, err := json.Marshal(callMsg)
+	if err != nil {
+		t.Fatalf("unable to marshal CallMsg - %s", err)
+	}
+
+	callMsgHex := hexutil.Encode(callMsgBytes)
+	getBalanceJSON := makeHTTPEthJSONReqAsJSON(rpc.RPCEstimateGas, []string{callMsgHex, latestBlock})
+
+	if getBalanceJSON[walletextension.RespJSONKeyResult].(string) != "0x12a05f200" {
```
joeldudley, line 446: @otherview Can we have a constant for this magic number? It would be great if it could even be calculated somehow, but I assume that's not possible?

go/enclave/enclave.go
```diff
@@ -22,24 +22,25 @@ import (
...
 	"github.com/obscuronet/go-obscuro/go/common"
-	obscurocore "github.com/obscuronet/go-obscuro/go/enclave/core"
+
 	"github.com/obscuronet/go-obscuro/go/enclave/db"
 	"github.com/obscuronet/go-obscuro/go/enclave/mempool"
 	"github.com/obscuronet/go-obscuro/go/ethadapter/erc20contractlib"
 	"github.com/obscuronet/go-obscuro/go/ethadapter/mgmtcontractlib"
+
+	gethcommon "github.com/ethereum/go-ethereum/common"
+	obscurocore "github.com/obscuronet/go-obscuro/go/enclave/core"
```
joeldudley, line 42: Shouldn't we use `gethcore` and `gethcrypto` instead of aliasing these, for consistency?

go/rpc/encrypted_client.go
```diff
@@ -33,6 +33,7 @@ var sensitiveMethods = []string{
 	RPCGetTxReceipt,
 	RPCSendRawTransaction,
 	RPCSubscribe,
+	RPCEstimateGas,
```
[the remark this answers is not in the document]
joeldudley: @otherview And the same here: https://github.com/obscuronet/go-obscuro/blob/main/tools/walletextension/accountmanager/account_manager.go#L256

go/enclave/enclave_test.go
```diff
@@ -0,0 +1,225 @@
...
+	"github.com/obscuronet/go-obscuro/integration/datagenerator"
+
+	gethcommon "github.com/ethereum/go-ethereum/common"
+)
+
+const _testEnclavePublicKeyHex = "034d3b7e63a8bcd532ee3d1d6ecad9d67fca7821981a044551f0f0cbec74d0bc5e"
+
+var _enclavePubKey *ecies.PublicKey
+
+func init() { //nolint:gochecknoinits
```
joeldudley, line 26: @otherview Is it more idiomatic to do this with a `TestMain`, like here? https://github.com/obscuronet/go-obscuro/blob/main/integration/walletextension/wallet_extension_test.go#L94

go/enclave/enclave.go
```diff
@@ -561,6 +562,37 @@ func (e *enclaveImpl) Stop() error {
...
+
+	// extract params from byte slice to array of strings
+	var paramString []string
+	err = json.Unmarshal(paramBytes, &paramString)
+	if err != nil {
+		return nil, fmt.Errorf("unable to decode EstimateGas params - %w", err)
+	}
+
+	// marshall the string into a callMsg
+	callMsg, err := rpc.ExtractCallMsg(gethcommon.Hex2Bytes(paramString[0][2:]))
```
joeldudley, line 582: @otherview Can you add a comment explaining the slicing here?

integration/walletextension/wallet_extension_test.go
```diff
@@ -427,6 +429,25 @@ func TestCannotSubscribeOverHTTP(t *testing.T) {
...
+		t.Fatalf("unexpected gas")
```
joeldudley, line 447: @otherview Can you print the expected vs actual gas here? e.g. it will be interesting to see if the actual gas is `0x0`.

integration/walletextension/wallet_extension_test.go
joeldudley, line 432: @otherview Can we have a reverse test, that you cannot estimate gas without submitting a viewing key?

joeldudley, line 432: @otherview Also, it would be nice if the test name could be aligned with those of all the other tests (e.g. `TestCanEstimateGasAfterSubmittingViewingKey`).

integration/walletextension/wallet_extension_test.go
BedrockSquirrel: If you use the `testify` library that is used in some places in our code (perhaps only by me, thought it was standard...) you get that output for free. I.e. `assert.Equals(result, expected, "unexpected gas")`

go/enclave/enclave_test.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+}
+
+func gasEstimateSuccess(t *testing.T, w wallet.Wallet, enclave common.Enclave, vk *rpc.ViewingKey) {
+	// create the callMsg
+	callMsg := datagenerator.CreateCallMsg()
+	callMsg.From = w.Address()
+	callMsgBytes, err := json.Marshal(callMsg)
+	if err != nil {
+		t.Error(err)
```
joeldudley, line 71: @otherview I don't think I agree with the mix of `t.Error`/`panic`/`t.Fatal` in these tests. I think in a few places, you use a `t.Error`, but the subsequent steps don't make sense if that step fails (e.g. if you can't encrypt the params, why try and estimate gas from them?).

go/enclave/enclave_test.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	// callMsg serialized as a Hex
+	callMsgHex := "0x" + gethcommon.Bytes2Hex(callMsgBytes)
```
joeldudley, line 75: @otherview I think it would be good to pull reusable bits of code into helper functions to help make the tests more readable. There's lots of hexifying/encrypt/submitting, such that it can hide the purpose of the test.

integration/walletextension/wallet_extension_test.go
otherview: Good point - This will change with the next PR when the gas estimations are dynamic. Just a place holder to make sure the tests are accurate atm

go/rpc/encrypted_client.go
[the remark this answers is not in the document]
otherview: Oh, yeah, didn't know about that. Nice catch 👍

go/enclave/enclave_test.go
otherview: I don't think TestMain would be required here as it's not trying to execute tests in the execution scope. The init is not running a test per se, just initializing a var.

go/enclave/enclave.go
otherview: I didn't create aliasing, but makes sense, I can fix it 👍

go/enclave/enclave.go
otherview: Also added some more testing 👍

integration/walletextension/wallet_extension_test.go
otherview: Yep, in the plans when the gas estimation is dynamic. I don't expect it to change atm.

integration/walletextension/wallet_extension_test.go
otherview: Can do 👍

go/enclave/enclave_test.go
otherview: Yep, that's a good point. Cleaned and reviewed those.

go/enclave/enclave_test.go
otherview: Agree. I don't want to create function to replace a one line action. But as there more endpoints being tested I think helpers will gain a better shape.

integration/walletextension/wallet_extension_test.go
otherview: Yeah that's a really good point.
I think as we expand on the WE and test coverage we'll be able to decouple testing WE wrapping and enclave logic ?
That's kinda where I was going with the `enclave_test.go` but as you refered we should brainstorm a bit more about it. This was just a first go at it I guess.

go/enclave/enclave_test.go
otherview: To be fair the values are not random at all ! I just noticed that, but let me change that name, you're right.

integration/walletextension/wallet_extension_test.go
otherview: I think you guys are in the right direction. Changed the errors to be more narrowed (not a catch all) and used the `testify` package too!

integration/walletextension/wallet_extension_test.go
joeldudley: Yeah - these have stopped being wallet extension tests, and more tests of single RPC calls. Maybe it would clear it up to just rename this to `rpc_test.go` or whatever, and add a readme (in another PR)?

go/enclave/enclave_test.go
joeldudley: I prefer something like the following, but maybe you think that's worse?

```
var _enclavePubKey = loadPubKey()

func loadPubKey() *ecies.PublicKey {
	// fetch the usable enclave pub key
	enclPubECDSA, err := crypto.DecompressPubkey(gethcommon.Hex2Bytes(_testEnclavePublicKeyHex))
	if err != nil {
		panic(err)
	}

	return ecies.ImportECDSAPublic(enclPubECDSA)
}
```

integration/walletextension/wallet_extension_test.go
joeldudley: The additional tests would test how the WE handles this (it should reject it with some meaningful error, rather than blowing up or whatever). It comes back to what we test via the new `enclave_test.go` vs here.

Tbh, I think I prefer having the tests here rather than in `enclave_test.go`, since it tests more (it also tests errors passing back to the host and through the WE). Think it maybe subjective.

@BedrockSquirrel Thoughts?

joeldudley (approved):

go/enclave/enclave_test.go
otherview: Discussed offline 👍

PR #672, opened 2022-09-15
Parse filter criteria in log subscription

joeldudley:

When subscribing for logs, a filter criteria object can be passed, to control which logs are received.

- Parses filter criteria object


go/rpc/encrypted_client.go
```diff
@@ -177,19 +179,38 @@ func (c *EncRPCClient) Subscribe(ctx context.Context, namespace string, ch inter
...
+	// If there are less than two arguments, it means no filter criteria was passed.
+	if len(args) < 2 {
+		logSubscription.Filter = &filters.FilterCriteria{}
+	} else {
+		filterCriteriaJSON, err := json.Marshal(args[1])
+		if err != nil {
+			return nil, fmt.Errorf("could not marshall filter criteria to JSON. Cause: %w", err)
+		}
+
+		filterCriteria := &filters.FilterCriteria{}
```
BedrockSquirrel, line 204: Think I see this more commonly as `var filterCriteria filters.FilterCriteria` and then the pointer ref on the next line should still work.

go/rpc/encrypted_client.go
```diff
@@ -177,19 +179,38 @@ func (c *EncRPCClient) Subscribe(ctx context.Context, namespace string, ch inter
...
+		err = json.Unmarshal(filterCriteriaJSON, &filterCriteria)
```
BedrockSquirrel, line 205: How comes we have to marshal and then unmarshal here? Guess casting doesn't work for some reason? Maybe worth a comment to avoid confusion/people trying to get rid of it without realising why it's needed.

go/rpc/encrypted_client.go
```diff
@@ -177,19 +179,38 @@ func (c *EncRPCClient) Subscribe(ctx context.Context, namespace string, ch inter
...
+		if err != nil {
+			return nil, fmt.Errorf("could not unmarshall filter criteria to JSON. Cause: %w", err)
```
BedrockSquirrel, line 207: My favourite typo haha, marshal is such a stupid word. Double-L here and on line 201.

BedrockSquirrel (approved):
LGTM - v minor comments

go/rpc/encrypted_client.go
joeldudley: Fixed.

go/rpc/encrypted_client.go
joeldudley: Fixed

go/rpc/encrypted_client.go
joeldudley: Yeah, I'll add one. Basically, we get the object as a map, and we need it as JSON to map it to a filter criteria

PR #673, opened 2022-09-15
Creates WE unit test file. Moves a test there from integration tests.

joeldudley:

We want to move away from having a second set of integration tests, and do more unit testing of the WE.

- Moves a first test from the WE integration tests to the WE unit tests.


BedrockSquirrel (approved):
Yeah seems reasonable to me, still feel like we haven't quite got got a good plan on the testing front but this seems like a worthwhile move in the right direction.

PR #675, opened 2022-09-15
WE: Move integration tests of sensitive methods without viewing key to unit tests

joeldudley:

Previously, we had individual integration tests for each sensitive method. This was a pain to update.

- Deleted the integration tests for calling each sensitive method without a viewing key
- Added a unit test to check that none of the sensitive methods can be called without a viewing key

### DIFF
--- a/tools/walletextension/test/wallet_extension_test.go
+++ b/tools/walletextension/test/wallet_extension_test.go
@@ -3,8 +3,10 @@ package test
 import (
 	"context"
 	"fmt"
+	"github.com/obscuronet/go-obscuro/tools/walletextension/accountmanager"
 	"net/http"
 	"os"
+	"strings"
 	"testing"
 
 	"github.com/ethereum/go-ethereum/eth/filters"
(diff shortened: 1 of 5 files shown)


tools/walletextension/test/wallet_extension_test.go
```diff
@@ -70,14 +92,15 @@ func createWalExt() (func(), error) {
 
 // Creates a dummy host that the wallet extension can connect to.
 func createDummyHost() (*http.Server, error) {
-	server := &http.Server{Addr: fmt.Sprintf("%s:%d", localhost, nodePortWS)} //nolint:gosec
-	http.HandleFunc("/ready", func(w http.ResponseWriter, r *http.Request) {})
-	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+	serveMux := http.NewServeMux()
```
joeldudley, line 95: This approach was required to allow us to run multiple tests from the same file. Otherwise, it complains about endpoint reregistration.

BedrockSquirrel (approved):
LGTM

PR #676, opened 2022-09-15
Replaces server with real Geth RPC layer.

joeldudley:

It will be easier to write tests going forward, as we can easily add API endpoints to hit.

- Replaces custom HTTP server with Geth RPC layer in WE unit tests
- Uses test cleanup instead of using defers and shutdown handles in WE unit tests


BedrockSquirrel (approved):
LGTM

PR #677, opened 2022-09-15
WE: Replace can-do integration tests with unit tests

joeldudley:

It's simpler to have a unit test that tests all sensitive methods, rather than an integration test for each one.

- Creates a unit test to test the result can be decrypted for all sensitive methods
- Removes integration tests that test the same


BedrockSquirrel (approved):
LGTM

PR #679, opened 2022-09-15
Revert to always encoding rpc params as a list

BedrockSquirrel:

- We made a change to encrypt param as a single element if there was just one of them but semantically it is a list really and we hit issues with methods that have one req param and one optional param (fiddly for code that handles it)

- Revert commit https://github.com/obscuronet/go-obscuro/commit/1f6f04ba81cf907ceade1ffa1a9bb40e98ec6531
- Make estimateGas tolerant to not always getting two params

### DIFF
--- a/go/enclave/enclave.go
+++ b/go/enclave/enclave.go
@@ -360,12 +360,15 @@ func (e *enclaveImpl) GetTransaction(encryptedParams common.EncryptedParamsGetTx
 	if err != nil {
 		return nil, fmt.Errorf("failed to decrypt encrypted RPC request params. Cause: %w", err)
 	}
-	var txHex string
-	err = json.Unmarshal(hashBytes, &txHex)
+	var paramList []string
+	err = json.Unmarshal(hashBytes, &paramList)
 	if err != nil {
-		return nil, fmt.Errorf("failed to unmarshal eth_getTransactionByHash params from JSON. Cause: %w", err)
+		return nil, fmt.Errorf("failed to unmarshal RPC request params from JSON. Cause: %w", err)
 	}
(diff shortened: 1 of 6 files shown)


go/enclave/enclave.go
```diff
@@ -567,19 +570,20 @@ func (e *enclaveImpl) EstimateGas(encryptedParams common.EncryptedParamsEstimate
...
 		return nil, fmt.Errorf("unable to decode EstimateGas params - %w", err)
 	}
 
-	// params must always be [callMsg, block number]
-	if len(paramString) != 2 {
-		return nil, fmt.Errorf("invalid number of params, expected 2 got %d", len(paramString))
+	// params are [callMsg, block number (optional)]
+	// todo: handle blocknum, for now we always assume "latest" as with other methods
+	if len(paramList) == 0 {
+		return nil, fmt.Errorf("required at least one param, but received zero")
```
otherview, line 582: Just had a quick look - I think it's fine to merge as it fixes things, but the API for eth_call assumes a callMsg and a block number always. https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_estimategas

Could the issue be that if something sends a blockNumber:nill the paramList len becomes 1 instead of 2 ?

tudor-malene (approved):
lgtm

go/enclave/enclave.go
BedrockSquirrel: Yeah I think that was it, although I think the problem is when it doesn't send a second param rather than sending nil. External tools don't always send block number when they don't have to which causes problems of we're treating exactly one argument as a special case.

PR #680, opened 2022-09-16
WE: Unit test API should decrypt params with enclave key to check validity

joeldudley:

Currently, in the unit tests, the API just sends back the message "success", encrypted with the viewing key. This means that we do not check that the WE is sending the right params to the API, or that they're encrypted with the enclave private key

- The API now decrypts the received params with the enclave private key, and sends them back encrypted with the viewing key
- Combines the two APIs - we didn't actually need the EthAPI and ObscuroAPI to be separate classes


tools/walletextension/test/apis.go
```diff
@@ -4,87 +4,134 @@ import (
...
-	encryptedBytes, err := ecies.Encrypt(rand.Reader, api.viewingKey, []byte(successMsg), nil, nil)
+func (api *DummyAPI) Logs(ctx context.Context, encryptedParams common.EncryptedParamsLogSubscription) (*rpc.Subscription, error) {
+	params, err := api.enclavePrivateKey.Decrypt(encryptedParams, nil, nil)
+	if err != nil {
+		return nil, fmt.Errorf("could not decrypt params with enclave private key")
+	}
+
+	notifier, _ := rpc.NotifierFromContext(ctx)
+	sub := notifier.CreateSubscription()
+	go func() {
```
BedrockSquirrel, line 106: I don't think this ever dies?

tools/walletextension/test/wallet_extension_test.go
```diff
@@ -119,56 +120,98 @@ func TestCannotInvokeSensitiveMethodsWithViewingKeyForAnotherAccount(t *testing.
...
+	_, viewingKeyBytes := RegisterPrivateKey(t, walExtAddr)
+	dummyAPI.setViewingKey(viewingKeyBytes)
 
-	respBody := MakeHTTPEthJSONReq(walExtAddr, rpc.RPCGetBalance, []interface{}{map[string]interface{}{}})
+	_, conn := MakeWSEthJSONReq(walExtAddrWS, rpc.RPCSubscribe, []interface{}{rpc.RPCSubscriptionTypeLogs, filterCriteriaJSON{Topics: []interface{}{dummyHash}}})
 
-	if !strings.Contains(string(respBody), successMsg) {
-		t.Fatalf("expected response containing '%s', got '%s'", successMsg, string(respBody))
+	// We set a timeout to kill the test, in case we never receive a log.
+	go func() {
```
BedrockSquirrel, line 188: I think you can write this more elegantly like:
```
timeout := time.AfterFunc(3 * time.Second, func() {
    panic("timed out waiting to receive a log via the subscription")
})
// then you can make sure it gets cancelled after success with:
defer timeout.Stop()
```

BedrockSquirrel (commented):
Couple comments but seems good generally

tools/walletextension/test/apis.go
joeldudley: It'll be reaped once the test ends. I think that's fine here.

BedrockSquirrel: The Go test engine cleans up go routines after every test?

BedrockSquirrel (approved):
LGTM

PR #683, opened 2022-09-16
Fix bug on network client call result pointer

BedrockSquirrel:

- client `Call` expects a pointer as the comment says, and it just passes it through to the next layer of client
- Tudor noticed difference between Call and CallContext, Call was taking the pointer of the result

- Change Call to pass the result through unchanged
- It's not used much, I've checked tests are passing and it should be identical to the context method.


tudor-malene (approved):
lgtm

PR #689, opened 2022-09-21
Logs: Route events correctly based on subscription IDs

joeldudley:

Different subscriptions should receive different logs.

- Route logs using the subscription ID as a topic
- Update checks in the sims
- Extends the sim validation logic for logs
- Fix bug in checking whether a topic could be a user address

### DIFF
--- a/go/enclave/events/subscription_manager.go
+++ b/go/enclave/events/subscription_manager.go
@@ -124,7 +124,7 @@ func getUserAddrs(log *types.Log, db *state.StateDB) []string {
 		// Since addresses are 20 bytes long, while hashes are 32, only topics with 12 leading zero bytes can
 		// (potentially) be user addresses.
 		topicHex := topic.Hex()
-		if topicHex[2:len(zeroBytesHex)/2] != zeroBytesHex {
+		if topicHex[2:len(zeroBytesHex)+2] != zeroBytesHex {
 			continue
 		}
 
--- a/go/enclave/events/subscription_manager.go
+++ b/go/enclave/events/subscription_manager.go
@@ -18,6 +18,7 @@ import (
 )
 
 const (
+	// The leading zero bytes in a hash indicating that it is possibly an address, since it only has 20 bytes of data.
 	zeroBytesHex = "000000000000000000000000"
 )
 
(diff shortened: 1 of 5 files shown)


go/enclave/events/subscription_manager.go
```diff
@@ -118,13 +119,13 @@ func (s *SubscriptionManager) EncryptLogs(logsBySubID map[uuid.UUID][]*types.Log
...
 func getUserAddrs(log *types.Log, db *state.StateDB) []string {
-	var nonContractAddrs []string //nolint:prealloc
+	var userAddrs []string //nolint:prealloc
 
 	for _, topic := range log.Topics {
 		// Since addresses are 20 bytes long, while hashes are 32, only topics with 12 leading zero bytes can
 		// (potentially) be user addresses.
 		topicHex := topic.Hex()
-		if topicHex[2:len(zeroBytesHex)/2] != zeroBytesHex {
+		if topicHex[2:len(zeroBytesHex)+2] != zeroBytesHex {
```
joeldudley, line 128: This was a bug, leading to everything being categorised as not-potentially-an-address. It wasn't picked up, because the fact that we weren't routing events based on subscription IDs meant we couldn't do relevancy checks in the sims. The changes to the sim validation will prevent us backsliding here.

integration/simulation/simulation.go
```diff
@@ -115,7 +116,14 @@ func (s *Simulation) trackLogs() {
 		return
 	}
 
+	subscriptionsCreated := 0
```
joeldudley, line 119: Previously, we only created a single subscription for a single wallet. I've bumped it to allowing two to be created here, since otherwise the relevancy checks in the validation are meaningless. Hopefully, once we've worked on the performance, we can give everyone a subscription.

integration/simulation/validate_chain.go
```diff
@@ -377,35 +379,64 @@ func checkLogsReceived(t *testing.T, s *Simulation) {
 		sub.Unsubscribe()
 	}
 
-	for _, channel := range s.LogChannels {
-		checkReceivedHOCAndPOCLogs(t, channel)
+	logsReceived := 0
```
joeldudley, line 382: In practice, not every subscription was getting a relevant log from both the HOC and POC contracts, due to the randomness in how the sim runs. So now, we check that at least one log is received (i.e. the subscriptions are actually delivering stuff), and then further down we check that for what we do receive, it's relevant and is from the HOC or POC contracts (the only event emitters in the sims).

integration/simulation/validate_chain.go
```diff
@@ -377,35 +379,64 @@ func checkLogsReceived(t *testing.T, s *Simulation) {
...
-				return
-			}
+		default:
+			return logsReceived
+		}
+	}
+}
+
+// Checks whether the log is relevant to the recipient (either a lifecycle event or a relevant user event).
+func isRelevant(owner string, receivedLog types.Log) bool {
```
joeldudley, line 417: We replicate a lot of the logic for deciding relevancy in the `SubscriptionManager` here.

BedrockSquirrel (approved):
LGTM

PR #691, opened 2022-09-21
Replaces panics in enclave client with errors

joeldudley:

Occasionally, during simulation shutdown, the shutdown sequence will cause one of these panics to be triggered, causing a successful simulation to fail. More generally, it's better to raise these issues and allow them to be handled.

- Replace panics in enclave_client with return of errors


BedrockSquirrel (approved):
LGTM

PR #697, opened 2022-09-22
Removes unnecessary AccountSubscription type.

joeldudley:

Previously, each subscription was going to have a list of authenticated accounts, so we made a struct to represent each one.

Since, we've decided to move towards a single account per subscription, so the struct isn't necessary.


BedrockSquirrel (approved):
LGTM

PR #700, opened 2022-09-23
Combine ingest and submit block methods

BedrockSquirrel:

- continuing efforts to simplify the host logic and improve its resilience
- this doesn't achieve much on its own but easier to read small PRs rather than a giant refactor/rework branch

- delete the ingest method
- add an `isLatest` flag to the submit block method so enclave knows if it should build a rollup


go/enclave/rollupchain/rollup_chain.go
```diff
@@ -469,21 +438,16 @@ func allReceipts(txReceipts []*types.Receipt, depositReceipts []*types.Receipt)
...
-		return common.BlockSubmissionResponse{IngestedBlock: false, BlockNotIngestedCause: "Block was genesis block."}
-	}
-
 	_, foundBlock := rc.storage.FetchBlock(block.Hash())
 	if foundBlock {
 		return common.BlockSubmissionResponse{IngestedBlock: false, BlockNotIngestedCause: "Block already ingested."}
 	}
 
-	if ingestionFailedResponse := rc.insertBlockIntoL1Chain(&block); ingestionFailedResponse != nil {
+	if ingestionFailedResponse := rc.insertBlockIntoL1Chain(&block); !rc.isGenesisBlock(&block) && ingestionFailedResponse != nil {
```
BedrockSquirrel, line 450: This check was being made by the `Ingest` method, inserting the genesis block fails but we still need to store it so we don't want to error out here.

This will go away when I make the change to start reading from the management contract deploy block so not too worried about it being ugly for now.

PR #703, opened 2022-09-26
Upgrade EdgelessDB to 4GB heap memory for testnets

BedrockSquirrel:

We have seen some testnet and dev-testnet failures recently where the enclave fails to write to its enclave (which is running in a different docker container on the same VM).

The EDB logs are not helpful, they just say something like "Aborted." Given this is a new issue and we've been running for longer with more state to persist recently I think we might be pushing up against some sort of limit.

- Edgeless provide EDB in two pre-canned flavours, 1GB heap and 4GB heap (customisable heap is on the roadmap)
- They say [here](https://docs.edgeless.systems/edgelessdb/#/getting-started/install?id=install-edgelessdb):
"Use edgelessdb-sgx-1gb primarily to test EdgelessDB on machines with limited resources. Prefer edgelessdb-sgx-4gb for production deployments."

And our testnets feel more like the latter description.

After this change we should continue to monitor and long term find some way to get stats on usage in non-production testnets (help us identify when it is problematic or if changes have dramatically increased it etc.).

- Upgrade testnet configs to use 4GB EDB image

### DIFF
--- a/docs/testnet/create-a-new-node.md
+++ b/docs/testnet/create-a-new-node.md
@@ -122,7 +122,7 @@ Docker Engine is installed and running. The `docker` group is created but no use
 ## Start Edgeless Database
 1. Run EdgelessDB on an SGX-capable system:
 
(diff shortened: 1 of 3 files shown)


tudor-malene (approved):
lgtm

PR #707, opened 2022-09-28
Use beefier SGX VMs for testnet to handle the 4GB EDB image

BedrockSquirrel:

Following discussion with Thomas at Edgeless we came to the conclusion that the DC2s SGX VM might not have the memory to handle the 4GB EDB image. He tested it successfully with the DC4s VM so this PR bumps our VMs up to DC4s.

If this PR is merged then I will *not* merge the PR that rolls back the 1GB -> 4GB EDB image upgrade so we can test if this fixes our issues with EDB falling over.


tudor-malene (approved):
lgtm

PR #708, opened 2022-09-28
Increase wait for enclave client connection

BedrockSquirrel:

- After recent change to use the production-ready 4GB Edgeless DB image we are seeing the enclave client fail to connect from the host
- I can see in the logs that the enclave finished its handshake with the edgeless DB just one second later (the previous issue of it aborting has resolved).
- There is no rush for this connection, the host should wait a reasonable amount of time at startup for the enclave RPC server to be responsive

- Increase enclave client connection timeout from 30 seconds to 60 seconds


tudor-malene (approved):
lgtm

PR #716, opened 2022-09-30
Include subscription ID in log response

joeldudley:

Per the docs here: https://ethereum.org/en/developers/tutorials/using-websockets/#subscription-types, we need to include the subscription ID when we return an event (a log, in our case).

Note that to do this, I had to return the ID from the host side each time. It seems like you should be given the ID when you create the subscription with the client, but you aren't.

An alternative would be to have two channels back, one to return the subscription ID once (which would then be stored by the client), and one to return the logs after that. But that felt heavy and confusing.


integration/simulation/simulation.go
```diff
@@ -66,14 +66,14 @@ func (s *Simulation) Start() {
 	s.prefundL1Accounts()      // Prefund every L1 wallet
 
 	timer := time.Now()
-	log.Info("Starting injection")
+	fmt.Printf("Starting injection\n")
```
joeldudley, line 69: It made more sense to print these so they can be seen when running the tests, than to stick them in the logs, IMO. It's quite nice seeing these appear, to show you that the full network sim is making progress.

go/common/types.go
```diff
@@ -197,3 +199,15 @@ type LogSubscription struct {
 	// A subscriber-defined filter to apply to the stream of logs.
 	Filter *filters.FilterCriteria
 }
+
+// IDAndEncLog pairs an encrypted log with the ID of the subscription that generated it.
+type IDAndEncLog struct {
+	ID     rpc.ID
```
BedrockSquirrel, line 205: Maybe call this field subID or something, casual reader would assume that's an ID for the log message

integration/simulation/simulation.go
BedrockSquirrel: Do both then imo, because it's nice when you have to investigate something in the log file to see when the injection started.

BedrockSquirrel (commented):
minor comments

BedrockSquirrel (approved):
LGTM

PR #718, opened 2022-09-30
WE: Return correct response when new subscription is created

joeldudley:

A subscription is supposed to return the subscription ID when the subscription is created (e.g. to allow unsubscription, and to establish which events correct to which subscription). See https://geth.ethereum.org/docs/rpc/pubsub for the format the response is expected is.

### DIFF
--- a/go/host/hostrunner/cli.go
+++ b/go/host/hostrunner/cli.go
@@ -90,8 +90,8 @@ func ParseConfig() config.HostConfig {
 	cfg.P2PPublicAddress = *p2pPublicAddress
 	cfg.L1NodeHost = *l1NodeHost
 	cfg.L1NodeWebsocketPort = uint(*l1NodePort)
-	cfg.ClientRPCTimeout = time.Duration(*enclaveRPCTimeoutSecs) * time.Second
-	cfg.EnclaveRPCTimeout = time.Duration(*clientRPCTimeoutSecs) * time.Second
+	cfg.ClientRPCTimeout = time.Duration(*clientRPCTimeoutSecs) * time.Second
+	cfg.EnclaveRPCTimeout = time.Duration(*enclaveRPCTimeoutSecs) * time.Second
 	cfg.L1RPCTimeout = time.Duration(*l1RPCTimeoutSecs) * time.Second
 	cfg.P2PConnectionTimeout = time.Duration(*p2pConnectionTimeoutSecs) * time.Second
 	cfg.RollupContractAddress = common.HexToAddress(*rollupContractAddress)
--- a/go/host/interfaces.go
+++ b/go/host/interfaces.go
@@ -26,7 +26,7 @@ type Host interface {
 	// Subscribe feeds logs matching the encrypted log subscription to the matchedLogs channel.
 	Subscribe(id rpc.ID, encryptedLogSubscription common.EncryptedParamsLogSubscription, matchedLogs chan []byte) error
 	// Unsubscribe terminates a log subscription between the host and the enclave.
-	Unsubscribe(id rpc.ID) error
+	Unsubscribe(id rpc.ID)
 	// Stop gracefully stops the host execution.
 	Stop()
 }
--- a/go/host/rpc/clientapi/client_api_filter.go
+++ b/go/host/rpc/clientapi/client_api_filter.go
@@ -39,7 +39,7 @@ func (api *FilterAPI) Logs(ctx context.Context, encryptedParams common.Encrypted
 	}
 
 	// We return the ID of the newly-created subscription, before returning any log events.
-	err = notifier.Notify(subscription.ID, common.IDAndEncLog{ //nolint:contextcheck
+	err = notifier.Notify(subscription.ID, common.IDAndEncLog{
 		SubID: subscription.ID,
 	})
 	if err != nil {
--- a/go/host/rpc/clientapi/client_api_filter.go
+++ b/go/host/rpc/clientapi/client_api_filter.go
@@ -38,6 +38,9 @@ func (api *FilterAPI) Logs(ctx context.Context, encryptedParams common.Encrypted
 		return nil, fmt.Errorf("could not subscribe for logs. Cause: %w", err)
 	}
 
+	// todo - joel - send back ID immediately
+	// We return the ID of the newly-created subscription.
+
 	go func() {
 		for {
 			select {
--- a/go/host/rpc/clientapi/client_api_filter.go
+++ b/go/host/rpc/clientapi/client_api_filter.go
@@ -39,7 +39,7 @@ func (api *FilterAPI) Logs(ctx context.Context, encryptedParams common.Encrypted
 	}
 
 	// We return the ID of the newly-created subscription, before returning any log events.
-	err = notifier.Notify(subscription.ID, common.IDAndEncLog{
+	err = notifier.Notify(subscription.ID, common.IDAndEncLog{ //nolint:contextcheck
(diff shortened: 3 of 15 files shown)


go/host/hostrunner/cli.go
```diff
@@ -90,8 +90,8 @@ func ParseConfig() config.HostConfig {
 	cfg.P2PPublicAddress = *p2pPublicAddress
 	cfg.L1NodeHost = *l1NodeHost
 	cfg.L1NodeWebsocketPort = uint(*l1NodePort)
-	cfg.ClientRPCTimeout = time.Duration(*enclaveRPCTimeoutSecs) * time.Second
```
joeldudley, line 93: Came across this mix-up, thought I'd fix it here.

go/host/interfaces.go
```diff
@@ -26,7 +26,7 @@ type Host interface {
 	// Subscribe feeds logs matching the encrypted log subscription to the matchedLogs channel.
 	Subscribe(id rpc.ID, encryptedLogSubscription common.EncryptedParamsLogSubscription, matchedLogs chan []byte) error
 	// Unsubscribe terminates a log subscription between the host and the enclave.
-	Unsubscribe(id rpc.ID) error
```
joeldudley, line 29: I felt that logging the error in the method was better than returning an error. There's not much the consumer can do if the unsubscribe fails.

go/host/rpc/clientapi/client_api_filter.go
```diff
@@ -38,6 +38,15 @@ func (api *FilterAPI) Logs(ctx context.Context, encryptedParams common.Encrypted
 		return nil, fmt.Errorf("could not subscribe for logs. Cause: %w", err)
 	}
 
+	// We return the ID of the newly-created subscription, before returning any log events.
```
joeldudley, line 41: Since there's no way for us to get the subscription ID based on the returned subscription object, we send a first packet that's just the subscription ID (and no log) instead.

go/rpc/client.go
```diff
@@ -45,7 +45,7 @@ type Client interface {
 	// CallContext If the context is canceled before the call has successfully returned, CallContext returns immediately.
 	CallContext(ctx context.Context, result interface{}, method string, args ...interface{}) error
 	// Subscribe creates a subscription to the Obscuro host.
-	Subscribe(ctx context.Context, namespace string, channel interface{}, args ...interface{}) (*rpc.ClientSubscription, error)
+	Subscribe(ctx context.Context, result interface{}, namespace string, channel interface{}, args ...interface{}) (*rpc.ClientSubscription, error)
```
joeldudley, line 48: We now need a `result` object, which we can set the subscription ID to.

tools/walletextension/test/wallet_extension_test.go
```diff
@@ -216,26 +218,35 @@ func TestCanSubscribeForLogsOverWebsockets(t *testing.T) {
...
+	logMap := logResp[common.JSONKeyParams].(map[string]interface{})[common.JSONKeyResult].(map[string]interface{})
 	logTopic := logMap[jsonKeyTopics].([]interface{})[0].(string)
 
 	if !strings.Contains(logTopic, dummyHash.Hex()) {
 		t.Fatalf("expected response containing '%s', got '%s'", dummyHash.Hex(), logTopic)
 	}
 }
 
+// Checks that the response to a subscription request is correctly-formatted.
+func validateSubscriptionResponse(t *testing.T, resp []byte) {
```
joeldudley, line 242: I use a regex to ensure the response is well-formed. I should probably do this for the other tests too. I've opened a ticket.

go/rpc/encrypted_client.go
```diff
@@ -147,45 +147,53 @@ func (c *EncRPCClient) Subscribe(ctx context.Context, namespace string, ch inter
 		return nil, fmt.Errorf("expected a channel of type `chan types.Log`, got %T", ch)
 	}
 	clientChannel := make(chan common.IDAndEncLog)
-	subscription, err := c.obscuroClient.Subscribe(ctx, namespace, clientChannel, subscriptionType, encryptedParams)
+	subscription, err := c.obscuroClient.Subscribe(ctx, nil, namespace, clientChannel, subscriptionType, encryptedParams)
 	if err != nil {
 		return nil, err
 	}
 
-	go func() {
```
joeldudley, line 155: The goroutine has been pulled out into a new `forwardLogs` method. I wouldn't have done the refactoring in this PR, but the linter was complaining about cyclomatic complexity, and you know I hate cyclomatic complexity.

go/rpc/encrypted_client.go
```diff
@@ -147,45 +147,53 @@ func (c *EncRPCClient) Subscribe(ctx context.Context, namespace string, ch inter
...
-					}
-					logCh <- idAndLog
-				}
+	return subscription, nil
+}
 
-			case <-subscription.Err(): // This channel's sole purpose is to be closed when the subscription is unsubscribed.
-				log.Error("subscription closed")
-				break
+func (c *EncRPCClient) forwardLogs(clientChannel chan common.IDAndEncLog, logCh chan common.IDAndLog, subscription *rpc.ClientSubscription) {
```
joeldudley, line 166: All the logic in `forwardsLogs` is unchanged, the diff just comes up a bit ugly.

go/rpc/encrypted_client.go
```diff
@@ -122,7 +122,7 @@ func (c *EncRPCClient) CallContext(ctx context.Context, result interface{}, meth
 	return nil
 }
 
-func (c *EncRPCClient) Subscribe(ctx context.Context, namespace string, ch interface{}, args ...interface{}) (*rpc.ClientSubscription, error) {
+func (c *EncRPCClient) Subscribe(ctx context.Context, result interface{}, namespace string, ch interface{}, args ...interface{}) (*rpc.ClientSubscription, error) {
```
BedrockSquirrel, line 125: If result is always a subscription ID could we have it be a return value with the Subscription? Or does this match a geth interface or something?

go/rpc/encrypted_client.go
BedrockSquirrel: It's my third least-favourite kind of complexity

tools/walletextension/test/wallet_extension_test.go
BedrockSquirrel: It might be cleaner/less fragile to just use the json parser. Formatting and ordering shouldn't matter and it can fish out the values as part of that.

BedrockSquirrel (commented):
minor comments

go/rpc/encrypted_client.go
joeldudley: I tried, but it doesn't come out very clean. The issue is that every client implements the same interface, so changing this forces `networkClient` to just return a nil value, which would then be overridden by enc rpc client based on the subscription contents.

tools/walletextension/common/constants.go
```diff
@@ -15,4 +15,6 @@ const (
 	JSONKeyRPCVersion   = "jsonrpc"
 	JSONKeySignature    = "signature"
 	JSONKeySubscription = "subscription"
+
+	JSONID = "1"
```
BedrockSquirrel, line 19: If this is just a dummy ID for testing then it probably shouldn't be in here

BedrockSquirrel (commented):
tiny thing, that JSONID constant seems more important than I think it is

BedrockSquirrel (approved):
LGTM

PR #720, opened 2022-10-03
Fix broken subscription filter criteria. Expand integration tests of subscriptions.

joeldudley:

The filter criteria were broken before. This is something I'd meant to investigate for a while, as I was seeing some failures in the CI gates. But because we were testing so few subscriptions, it usually passed.

- Fixes encoding and decoding of filter criteria when adding a subscription. For some reason, JSON marshaling of the subscription was causing the filter criteria to decode as an empty object, so the filters were lost. This is fixed by switching to RLP
- Expands subscriptions, creating one for every client on every wallet, rather than one client each on only two wallets. This was causing performance issues before, but these have been addressed

### DIFF
--- a/integration/simulation/validate_chain.go
+++ b/integration/simulation/validate_chain.go
@@ -342,9 +342,7 @@ func checkTransactionReceipts(ctx context.Context, t *testing.T, nodeIdx int, rp
 
 		// We check that the logs are relevant to the sender.
 		for _, retrievedLog := range receipt.Logs {
-			if !isRelevant(sender.Hex(), *retrievedLog) {
-				t.Errorf("receipt contained log that was not relevant (neither a lifecycle event nor relevant to the sender)")
-			}
+			assertIsRelevant(t, sender.Hex(), *retrievedLog)
 		}
 
 		if receipt.Status == types.ReceiptStatusFailed {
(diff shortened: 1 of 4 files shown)


integration/simulation/validate_chain.go
```diff
@@ -393,38 +393,41 @@ func checkLogsReceived(t *testing.T, s *Simulation) {
...
 		t.Errorf("no logs received during simulation")
 	}
 }
 
 // Checks that the owner has only received relevant logs, and only from the HOC or POC contracts.
 func checkReceivedHOCAndPOCLogs(t *testing.T, owner string, channel chan common.IDAndLog) int {
-	logsReceived := 0
+	var logsReceived []*types.Log
 
+out:
```
BedrockSquirrel, line 405: Would this work to drain the queue without the break label thingy?
```
for len(channel) > 0 {
   idAndLog <-channel
   // ...
}
```

BedrockSquirrel (approved):
LGTM, minor suggestion

integration/simulation/validate_chain.go
joeldudley: It doesn't seem to work, no. Not sure why, maybe one to revisit in the future.

PR #721, opened 2022-10-03
Avoids returning empty log lists to reduce traffic.

joeldudley:

Previously, we were returning a log list for every active subscription in every block submission response - even if no new logs were created for that subscription. To minimise traffic, we should only return a log list for a given subscription if it has any items.


BedrockSquirrel (approved):
lgtm

PR #723, opened 2022-10-03
Return logs keyed by rollup number by subscription ID, rather than just by subscription ID

joeldudley:

In each block submission response, the enclave returns all logs matching all subscriptions, even historical ones. This is needed to support retrieving log snapshots in the future, and to handle dropped connections.

But to match the Geth API, we need to only return logs via subscriptions once. As a first step to do this, we need to return the rollup number each log is for, so that we'll be able to ensure on the host side that we are returning the logs for a given rollup only once.


go/host/node/node.go
```diff
@@ -609,8 +609,10 @@ func (a *Node) storeBlockProcessingResult(result common.BlockSubmissionResponse)
 
 // Distributes logs to subscribed clients.
 func (a *Node) sendLogsToSubscribers(result common.BlockSubmissionResponse) {
-	for subscriptionID, encryptedLogs := range result.SubscribedLogs {
-		a.logsChs[subscriptionID] <- encryptedLogs
+	for subscriptionID, encLogsByRollup := range result.SubscribedLogs {
+		for _, encryptedLogs := range encLogsByRollup {
+			a.logsChs[subscriptionID] <- encryptedLogs
```
joeldudley, line 614: We aren't implementing the only-once delivery logic yet, to keep the PR small.

BedrockSquirrel (approved):
LGTM

PR #725, opened 2022-10-04
Each subscription receives each log exactly once

joeldudley:

Previously, each subscription would receive all logs matching the subscription every time there was a new block, even if they'd been sent them before. This did not match Geth's semantics, which was to send each log only once.

- Changes logic on the host, so that each log is only distributed once
- Adds tests of no duplicate logs to the integration tests

### DIFF
--- a/go/host/node/node.go
+++ b/go/host/node/node.go
@@ -67,15 +67,12 @@ type Node struct {
 	forkRPCCh    chan []common.EncodedBlock // The channel that new forks from the L1 node are sent to
 	rollupsP2PCh chan common.EncodedRollup  // The channel that new rollups from peers are sent to
 	txP2PCh      chan common.EncryptedTx    // The channel that new transactions from peers are sent to
-	logsChs      map[rpc.ID]chan []byte     // The channels that logs are sent to, one per subscription
 
 	nodeDB *db.DB // Stores the node's publicly-available data
 
-	// library to handle Management Contract lib operations
-	mgmtContractLib mgmtcontractlib.MgmtContractLib
-
-	// Wallet used to issue ethereum transactions
-	ethWallet wallet.Wallet
+	mgmtContractLib  mgmtcontractlib.MgmtContractLib // Library to handle Management Contract lib operations
+	ethWallet        wallet.Wallet                   // Wallet used to issue ethereum transactions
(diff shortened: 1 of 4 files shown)


integration/simulation/simulation.go
```diff
@@ -44,8 +44,8 @@ type Simulation struct {
 	SimulationTime   time.Duration
 	Stats            *stats.Stats
 	Params           *params.SimParams
-	LogChannels      map[string]chan common.IDAndLog // Maps an owner to the channel on which they receive logs for all their wallets.
-	Subscriptions    []ethereum.Subscription         // A slice of all created event subscriptions.
+	LogChannels      map[string][]chan common.IDAndLog // Maps an owner to the channels on which they receive logs for each client.
```
joeldudley, line 47: We need a slice of channels now, one for each client. Otherwise, with every client subscribing, we'd see dupes (one per client), which is what we're testing for.

go/host/node/node.go
```diff
@@ -607,11 +606,37 @@ func (a *Node) storeBlockProcessingResult(result common.BlockSubmissionResponse)
...
-			a.logsChs[subscriptionID] <- encryptedLogs
+		logSub, found := a.logSubscriptions[subscriptionID]
+		if !found {
+			log.Error("received a log for subscription with ID %s, but no such subscription exists", subscriptionID)
+			continue
+		}
+
+		for rollupNumber, encryptedLogs := range encLogsByRollup {
+			// We have received a log from a rollup this subscription hasn't seen before.
+			if rollupNumber > logSub.latestSeenRollup {
```
BedrockSquirrel, line 622: Bit confused by this latestSeenRollup thing, why do we send logs from the enclave for a subscription when that log had already been sent to the host? Or is that maybe an optimisation to come later?

BedrockSquirrel (approved):
LGTM, just need to clarify my understanding on the sifting through already sent logs bit.

Also I wonder if we should consider splitting out this subscription logic at some point, `node` is a bit of a dumping ground which I'm trying to slim down at the moment. But can leave that for now and wait til we've got a clearer picture on how it's going to be structured perhaps.

go/host/node/node.go
joeldudley: I think it could be optimised away in the future. Don't know if there's a concern about the host dropping logs or missing a certain block submission response, and needing them again.

PR #726, opened 2022-10-04
Pulls log event management out of node.go.

joeldudley:

node.go has become a dumping ground.


BedrockSquirrel (approved):
LVGTM, thanks - gives us something to unit test against if we find cases we want to check for too

PR #736, opened 2022-10-07
Move filtering out of already-seen logs from host to enclave

joeldudley:

Currently, we send all logs over RPC, only to drop the already-seen ones on the host. It's much more efficient to discard the already-seen ones on the enclave.

### DIFF
--- a/go/common/logs.go
+++ b/go/common/logs.go
@@ -17,6 +17,8 @@ type LogSubscription struct {
 	Signature *[]byte
 	// A subscriber-defined filter to apply to the stream of logs.
 	Filter *filters.FilterCriteria
+	// Used to track which rollups the subscription has already been sent logs for.
+	LastSeenRollup uint64
 }
 
 // IDAndEncLog pairs an encrypted log with the ID of the subscription that generated it.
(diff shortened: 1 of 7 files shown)


go/common/logs.go
```diff
@@ -17,6 +17,8 @@ type LogSubscription struct {
 	Signature *[]byte
 	// A subscriber-defined filter to apply to the stream of logs.
 	Filter *filters.FilterCriteria
+	// Used to track which rollups the subscription has already been sent logs for.
+	LastSeenRollup uint64
```
BedrockSquirrel, line 21: I wonder if this also needs to include a hash for the last seen rollup to make sure logs are updated if things change. Might be worth adding a todo, not sure if it's an issue or not but if it is I'll come across it with this block ordering work hopefully.

BedrockSquirrel (approved):
LGTM - optional todo comment to remind us to check it handles re-orgs etc. sensibly

go/common/logs.go
joeldudley: Cool, added a general todo to think about chain reorgs. Thanks.

PR #745, opened 2022-10-19
Don't store last seen rollup per subscription

joeldudley:

When I first implemented subscriptions, I thought subscriptions could also be used to return historical logs (e.g. subscribing from index `head - 100`). This meant we had to track which logs a subscription had been sent, to ensure it had "caught up" to the head without redelivering logs.

In practice, the `eth_subscribe` API is only for new subscriptions, so we can drop this mechanism.


BedrockSquirrel (approved):
LGTM

PR #752, opened 2022-10-21
Addition of change log and faucet server to readme

moraygrieve:

- In preparation for the 0.5 release

- The top level readme, and the change log


tudor-malene (approved):
lgtm

PR #770, opened 2022-10-27
Update testnet-kpis.md

GavT:

- Update to KPI source for "dApps in development"

- Text change

### DIFF
--- a/planning/testnet-kpis.md
+++ b/planning/testnet-kpis.md
@@ -7,8 +7,8 @@ Key performance indicators (KPIs) will be used to determine the amount of Testne
 |--|--|--|--|--|
 | Testnet documentation page views (https://docs.obscu.ro/testnet/) | Gauges interest from developers in building on testnet. A good early indicator. Straightforward to capture | Google Analytics | Number of unique page views in the last 4 weeks | Targets will be determined once initial baseline data is captured |
 | Testnet uptime | Captures how robust and ready for mainnet Obscuro is | DataDog avg:system.uptime{*} | Average Testnet uptime over the last 4 weeks||
-| Number of end user accounts| Captures users using testnet in general and testing new dApps | TBD| Number of new end user account addresses in the last 4 weeks ||
-|DApps in development| This is difficult to measure as it’s not something we can look to the network to accurately measure. We can use our business development CRM data as an indicator. | Airtable "Partnerships" tab| Number of "Soft Commits" and "LOI/MOU signed" in the last 4 weeks||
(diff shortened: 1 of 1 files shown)


tooward (commented):
- "LOI/MOU signed" - wasn't aware this is a part of our process. I'd like to know if this is necessary / typical for L2s before we put it as a metric that we measure ourselves against.
- Feels like we need to list anything that is a functional Dapp on Testnet on a webpage so others can use them when needed (ex: DEX). Maybe that is how we can measure it? As James noted, figuring out deployments is going to become an engineering task if we go the address route.
- Why is end user accounts needed if we are measuring wallet extension downloads? Measures same thing.

CaisManai:
> * Why is end user accounts needed if we are measuring wallet extension downloads? Measures same thing.

It measures conversion i.e. if a user downloads the wallet, will they run it? If we see a disparity between the two we'll know users don't trust the binary

GavT:
> * "LOI/MOU signed" - wasn't aware this is a part of our process. I'd like to know if this is necessary / typical for L2s before we put it as a metric that we measure ourselves against.

Yeah, this would just add friction.

GavT:
> * Feels like we need to list anything that is a functional Dapp on Testnet on a webpage so others can use them when needed (ex: DEX). Maybe that is how we can measure it? As James noted, figuring out deployments is going to become an engineering task if we go the address route.

There is something in the backlog to do this, but its non trivial to get it right. I expect this will become the data source for the KPI  of "new dApps deployed on testnet".

tooward:
> > * Why is end user accounts needed if we are measuring wallet extension downloads? Measures same thing.
> 
> It measures conversion i.e. if a user downloads the wallet, will they run it? If we see a disparity between the two we'll know users don't trust the binary

Yes, but my point is we need fewer metrics we actually review and change our behavior based on. This is a nuance. As a 
product person I agree you should look at this. For reporting in general two measures that capture a nuance won't be understood. KISS

tooward:
> > * Feels like we need to list anything that is a functional Dapp on Testnet on a webpage so others can use them when needed (ex: DEX). Maybe that is how we can measure it? As James noted, figuring out deployments is going to become an engineering task if we go the address route.
> 
> There is something in the backlog to do this, but its non trivial to get it right. I expect this will become the data source for the KPI of "new dApps deployed on testnet".

What is in the backlog? Website (simple) or figuring out what are non-owned accounts? Website pays in two ways - your metric and a better visual for anyone checking us out. I'd just use the website and move on.

CaisManai (approved):

PR #772, opened 2022-10-28
Use error response for failed enclave ingestion

BedrockSquirrel:

- Cleaning up the node/enclave structure for recovering from persistence
- Want to start returning the hash of the head L1 block that enclave has seen (can't atm as enclave doesn't store L1)

- Remove IngestedBlock flag
- Return error when failed to ingest
- Add ability to include L1 head hash in error


go/host/node/node.go
```diff
@@ -468,43 +475,33 @@ type blockAndParent struct {
...
 }
 
-func (a *Node) processBlocks(blocks []common.EncodedBlock, interrupt *int32) error {
-	var result common.BlockSubmissionResponse
-	for _, block := range blocks {
-		// For the genesis block the parent is nil
-		if block == nil {
-			continue
-		}
+func (a *Node) processBlock(block common.EncodedBlock, latest bool, interrupt *int32) error {
```
BedrockSquirrel, line 478: This method used to take a list of blocks (either a parent and child pair or a list of blocks from a fork).

It treated the last block in that list differently (error was more important and it bothered to publish the rollup). This change makes that explicit and hopefully easier to follow.

go/common/enclave.go
```diff
@@ -130,3 +132,48 @@ type ProducedSecretResponse struct {
 	RequesterID gethcommon.Address
 	HostAddress string
 }
+
+const (
+	bsePrefix = "blockSubmitError: "
+	l1HeadKey = "l1Head"
+)
+
+var l1HeadErrorSeparator = fmt.Sprintf(", %s=", l1HeadKey)
```
joeldudley, line 141: I think this is pretty overengineered. I'm having trouble working out what it's doing based on reading the code.

Maybe you just need a const for the separator, then you can either hardcode the equals or use a const for that?

go/common/enclave.go
```diff
@@ -130,3 +132,48 @@ type ProducedSecretResponse struct {
...
+	return r.Wrapped
+}
+
+// RehydrateBlockSubmitError is used to rebuild a BlockSubmitError type from a string (useful after its type has been lost in RPC communication)
+// the format of the error is expected to be: "BlockSubmitError: {errorMsg}, l1Head={l1HeadHash}".
+func RehydrateBlockSubmitError(err error) (*BlockSubmitError, bool) {
+	str := err.Error()
+	split := strings.Split(str, l1HeadErrorSeparator)
+	if len(split) != 2 {
+		return nil, false
```
joeldudley, line 168: Wouldn't it be better to return an informative error, rather than false?

go/common/enclave.go
```diff
@@ -130,3 +132,48 @@ type ProducedSecretResponse struct {
...
+	}
+	var hash *gethcommon.Hash
+	if len(split[1]) > 0 {
+		h := gethcommon.HexToHash(split[1])
+		hash = &h
```
joeldudley, line 173: Wonder if it's cleaner to declare `hash` as a non-pointer type, then take the address on line 176, below?

go/enclave/enclave.go
```diff
@@ -217,18 +217,21 @@ func (e *enclaveImpl) Start(block types.Block) error {
...
-		IngestedBlock:  true,
 	}, nil
 }
 
 // SubmitBlock is used to update the enclave with an additional L1 block.
-func (e *enclaveImpl) SubmitBlock(block types.Block, isLatest bool) (common.BlockSubmissionResponse, error) {
-	bsr := e.chain.SubmitBlock(block, isLatest)
+func (e *enclaveImpl) SubmitBlock(block types.Block, isLatest bool) (*common.BlockSubmissionResponse, error) {
+	bsr, err := e.chain.SubmitBlock(block, isLatest)
+	e.logger.Trace("Processed block submission", "success", err == nil, "blk", block.Number(), "blkHash", block.Hash())
```
joeldudley, line 231: Shouldn't you check the error before declaring this a success?

go/enclave/rollupchain/rollup_chain.go
```diff
@@ -104,26 +110,24 @@ func (rc *RollupChain) ProduceGenesis(blkHash gethcommon.Hash) (*obscurocore.Rol
 
 // Inserts the block into the L1 chain if it exists and the block is not the genesis block. Returns a non-nil
 // BlockSubmissionResponse if the insertion failed.
```
joeldudley, line 112: Outdated comment.

go/common/enclave.go
```diff
@@ -130,3 +132,48 @@ type ProducedSecretResponse struct {
...
+	bsePrefix = "blockSubmitError: "
+	l1HeadKey = "l1Head"
+)
+
+var l1HeadErrorSeparator = fmt.Sprintf(", %s=", l1HeadKey)
+
+// BlockSubmitError is used as a standard format for error response from enclave for block submission errors
+// The L1 Head hash tells the host what the enclave knows as the canonical chain head, so it can feed it the appropriate block.
+type BlockSubmitError struct {
+	L1Head  *gethcommon.Hash
```
joeldudley, line 146: Why this approach, rather than a separate RPC call to get the canonical chain head? Less efficient, but easier to reason about perhaps?

joeldudley (changes requested):
I'm not sure about this. It feels like it would be much cleaner without trying to return the L1 head hash as part of the error, and just use a separate RPC call. Easier to reason about, and wouldn't require the creation of `BlockSubmitError` and all the attendant casting/unwrapping/etc.

go/common/enclave.go
BedrockSquirrel: I think it was right idea, wrong execution. I think it's important that enclave says, "no, this is where I am" when given a block it's not expecting, multiple RPC calls is clunky.

But this hacky approach at using all the obscure features that Go's error system has to offer was wrong. I've just found the work @tudor-malene did around serializable errors using geth's RPC error types in `evm_facade.go`. `gethrpc.DataError` seems like exactly what we need here.

go/enclave/enclave.go
BedrockSquirrel: It prints `success=true` or `success=false` based on whether there was an error. It's the key=value magic thing that these loggers use instead of string formatter verbs.

But valid point, I'll just use the if statement that's there and add an else to give clear wording instead. It's too important to the meaning to hide it in the metadata.

go/common/enclave.go
```diff
@@ -130,3 +132,22 @@ type ProducedSecretResponse struct {
 	RequesterID gethcommon.Address
 	HostAddress string
 }
+
+// BlockRejectError is used as a standard format for error response from enclave for block submission errors
+// The L1 Head hash tells the host what the enclave knows as the canonical chain head, so it can feed it the appropriate block.
+type BlockRejectError struct {
+	L1Head  gethcommon.Hash
+	Wrapped error
```
joeldudley, line 140: Since you're forced to have the `Unwrap` method, maybe make this field private and force the use of that method?

go/enclave/rpc_server.go
```diff
@@ -125,6 +126,16 @@ func (s *server) SubmitBlock(_ context.Context, request *generated.SubmitBlockRe
 	bl := s.decodeBlock(request.EncodedBlock)
 	blockSubmissionResponse, err := s.enclave.SubmitBlock(bl, request.IsLatest)
 	if err != nil {
+		var rejErr common.BlockRejectError
+		isReject := errors.As(err, &rejErr)
+		if isReject {
+			// todo: we should avoid errors in response messages and use the gRPC error objects for this stuff (standardized across all enclave responses)
+			msg, err := rpc.ToBlockSubmissionRejectionMsg(&rejErr)
```
joeldudley, line 133: Is this logic correct? If we fail to convert to a block submission rejection message, we ignore the resulting error, and fall back to reporting the old error?

At a minimum, we should log the failed conversion to a block submission rejection message, I think.

joeldudley (approved):
Just a small change or two.

go/enclave/rpc_server.go
BedrockSquirrel: Yeah fair point, I'll log it clearly. It's very weird scenario because we already confirmed that it was a block rejection error.

go/common/enclave.go
BedrockSquirrel: Yeah kinda agree but would need a method to populate it as well so will probably just leave it. I hope this stuff doesn't live too long tbh.

PR #777, opened 2022-10-30
Watch for receipts to confirm L1 transactions

BedrockSquirrel:

- We sometimes have failures where secrets haven't exchanged or rollups get lost because L1 transaction never got mined
- These can be hard to diagnose and also we need to be provide visibility for this issue for node operators

- Add a watch for receipt part to the L1 transaction submission in the host
- Move the backing-off retry strategy to the retry package
- Fix minor bug in the retry logic where we checked if Done before recording the latest retry attempt (didn't matter in the timeout strategy)


tudor-malene (approved):
lgtm

PR #788, opened 2022-11-03
Renames node package to host. Pulls host interface out into common package.

joeldudley:

- Host was defined in a `node.go` file in a `node` package, which didn't make sense (a host is only half a node)
- Moved host interface definition to the common package (this was required to avoid an import cycle, and makes sense logically)


BedrockSquirrel (approved):
LGTM

PR #791, opened 2022-11-03
Confusing description of `DB.writeRollupNumber`. Minor clean-up

joeldudley:


go/host/db/hostdb.go
```diff
@@ -285,20 +285,20 @@ func (db *DB) writeHeadRollup(w ethdb.KeyValueWriter, val gethcommon.Hash) {
...
 func (db *DB) writeRollupHash(w ethdb.KeyValueWriter, header *common.Header) {
 	key := rollupHashKey(header.Number)
 	if err := w.Put(key, header.Hash().Bytes()); err != nil {
 		db.logger.Crit("could not put header in DB.", log.ErrKey, err)
 	}
 }
 
-// Stores the hash of a rollup into the database, keyed by the hashes of the transactions in the rollup
-func (db *DB) writeRollupNumber(w ethdb.KeyValueWriter, txHash gethcommon.Hash, rollupNumber *big.Int) {
+// Stores a rollup's number in the database, keyed by the hash of a transaction in that rollup.
```
joeldudley, line 296: This description change is the gist of the PR. Was confusing me for a bit what was meant here.

BedrockSquirrel (approved):

go/common/enclave.go
```diff
@@ -109,9 +109,8 @@ type Enclave interface {
 type BlockSubmissionResponse struct {
 	BlockHeader *types.Header // the header of the consumed block. Todo - only the hash required
 
-	ProducedRollup ExtRollup // The new Rollup when ingesting the block produces a new Rollup
```
joeldudley, line 112: I've renamed these fields to clarify what they mean, and dropped the `FoundNewHead` method (you can see if you've found a new head because the `NewRollupHeader` will be non-nil.

go/host/db/hostdb.go
```diff
@@ -52,16 +52,16 @@ func NewLevelDBBackedDB(logger gethlog.Logger) *DB {
 	}
 }
 
-// GetCurrentBlockHead returns the current block header (head) of the Node
-func (db *DB) GetCurrentBlockHead() *types.Header {
```
joeldudley, line 56: I renamed this to be clearer. You're getting the header of the head block.

joeldudley, line 56: I renamed this to be clearer. You're getting the header of the head block.

go/host/host.go
```diff
@@ -522,9 +522,9 @@ func (h *host) publishRollup(producedRollup common.ExtRollup) {
 
 func (h *host) storeBlockProcessingResult(result *common.BlockSubmissionResponse) {
 	// only update the host rollup headers if the enclave has found a new rollup head
-	if result.FoundNewHead {
+	if result.NewRollupHeader != nil {
 		// adding a header will update the head if it has a higher height
-		headerWithHashes := common.HeaderWithTxHashes{Header: result.RollupHead, TxHashes: result.ProducedRollup.TxHashes}
+		headerWithHashes := common.HeaderWithTxHashes{Header: result.NewRollupHeader, TxHashes: result.HeadRollup.TxHashes}
```
BedrockSquirrel, line 527: I think this is a (pre-existing but fairly recent) bug. ProducedRollup is the rollup *after* NewRollupHeader, I just put a log message in and verified that result.RollupHead was a different height to result.ProducedRollup

There's an important distinction between the produced rollup and the newly ingested rollup (the former is a candidate rollup that's just been produced to publish, the latter is a rollup that was just discovered in an L1 block). I don't think we can removed the field from the Response object.

BedrockSquirrel (changes requested):
Think this needs a bit of a rethink. There is a bug in the behaviour which is what made it confusing.

BedrockSquirrel (commented):
The build is reliably failing with a test failure which I think is relevant but not sure why. I'll try to get that bug fixed where we mix rollups from different heights and then hopefully there'll be less confusion to revisit this.

joeldudley:
@BedrockSquirrel I renamed an RPC method, but didn't update the string constant.

BedrockSquirrel (approved):
LGTM

PR #801, opened 2022-11-07
Remove the client side wait loop implementation as it is no longer needed

moraygrieve:

- Looks like it is no longer need to implement a wait loop on the client side in Obscuro (most likely due to returning nil rather than an error) so we can remove this from the documentation. 

- Docs around deploying a contract programatically 
- Change log


tudor-malene (approved):
lgtm.

Q: This is now in line with the behaviour of geth?

moraygrieve:
@tudor-malene yes this is now consistent with Geth and other networks

PR #804, opened 2022-11-07
Allow for balance to be called from other methods

otherview:

- Related with https://github.com/obscuronet/obscuro-internal/issues/929
- Minor refactor so that the dynamic gas estimation can check the balance of the account of the call

- Minor refactor of funcionality so it can be called in other methods


BedrockSquirrel (approved):
LGTM

PR #809, opened 2022-11-09
Have sims test the eth_blockNumber endpoint

joeldudley:

### DIFF
--- a/go/obsclient/obsclient.go
+++ b/go/obsclient/obsclient.go
@@ -1,9 +1,6 @@
 package obsclient
 
 import (
-	"math/big"
-
-	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/obscuronet/go-obscuro/go/rpc"
 )
 
@@ -28,16 +25,4 @@ func NewObsClient(c rpc.Client) *ObsClient {
 
 func (oc *ObsClient) Close() {
 	oc.rpcClient.Stop()
-}
-
-// Blockchain Access
-
-// ChainID retrieves the current chain ID for transaction replay protection.
-func (oc *ObsClient) ChainID() (*big.Int, error) {
-	var result hexutil.Big
-	err := oc.rpcClient.Call(&result, rpc.ChainID)
-	if err != nil {
-		return nil, err
-	}
-	return (*big.Int)(&result), err
 }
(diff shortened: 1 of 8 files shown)


go/obsclient/obsclient.go
```diff
@@ -29,15 +26,3 @@ func NewObsClient(c rpc.Client) *ObsClient {
 func (oc *ObsClient) Close() {
 	oc.rpcClient.Stop()
 }
-
-// Blockchain Access
-
-// ChainID retrieves the current chain ID for transaction replay protection.
-func (oc *ObsClient) ChainID() (*big.Int, error) {
```
BedrockSquirrel, line 36: Why get rid of this?

integration/simulation/utils.go
```diff
@@ -88,6 +90,23 @@ func getRollupHeader(client rpc.Client, hash gethcommon.Hash) *common.Header {
 	return result
 }
 
+// Uses the client to retrieve the current block number.
+func getBlockNumber(client rpc.Client) uint64 {
```
BedrockSquirrel, line 94: Can we add the method to Obsclient and remove this?

The analogous gethclient method looks like this:
```
// BlockNumber returns the most recent block number
func (ec *Client) BlockNumber(ctx context.Context) (uint64, error) {
	var result hexutil.Uint64
	err := ec.c.CallContext(ctx, &result, "eth_blockNumber")
	return uint64(result), err
}
```

BedrockSquirrel: Some of the other methods in here should be on the client too, could maybe leave a todo for them

BedrockSquirrel (changes requested):
Generally LGTM but can we move the call to the unauthenticated client

go/obsclient/obsclient.go
joeldudley: Unused. Seemed arbitrary to just have that one?

BedrockSquirrel: It's the client for unauthorised rpc requests., we just haven't added the rest yet.

joeldudley: Ok, I'll readd.

BedrockSquirrel (approved):
LGTM

PR #825, opened 2022-11-14
Adding health check endpoint

otherview:

- https://github.com/obscuronet/obscuro-internal/issues/1170
- As per title, adds a simple iteration of an healthpoint

- Obscuro endpoint added

### DIFF
--- a/testnet/testnet-local-build_images.sh
+++ b/testnet/testnet-local-build_images.sh
@@ -13,11 +13,11 @@ start_path="$(cd "$(dirname "${0}")" && pwd)"
 testnet_path="${start_path}"
 root_path="${testnet_path}/.."
 
-# run the builds in parallel
+# run the builds in parallel - echo the full command to output
 command() {
-    echo $4 start
+    echo $@ started
      $( "$@" )
-    echo $4 complete
+    echo $@ completed
 }
 
 command docker build -t testnetobscuronet.azurecr.io/obscuronet/obscuro_gethnetwork:latest -f "${testnet_path}/gethnetwork.Dockerfile" "${root_path}" &


go/common/enclave.go
```diff
@@ -103,6 +103,9 @@ type Enclave interface {
 
 	// GetLogs returns all the logs matching the filter.
 	GetLogs(encryptedParams EncryptedParamsGetLogs) (EncryptedResponseGetLogs, error)
+
+	// HealthCheck returns whether the enclave is in a healthy state
+	HealthCheck() (bool, error)
```
joeldudley, line 108: Should we just extend the `Status` method for this?

go/enclave/enclave.go
```diff
@@ -723,6 +723,15 @@ func (e *enclaveImpl) DoEstimateGas(args *gethapi.TransactionArgs, blockNr gethr
 	return hexutil.Uint64(hi), nil
 }
 
+// HealthCheck returns whether the enclave is deemed healthy
+func (e *enclaveImpl) HealthCheck() (bool, error) {
+	if e.storage.FetchHeadState() == nil {
```
joeldudley, line 728: I think it would be cleaner to add a new method. You don't want this to break if the API of this method ever breaks.

joeldudley:
@otherview To avoid rot, can you check that all the nodes are healthy (based on this endpoint) in the sims?

joeldudley (changes requested):
Per comments.

go/common/enclave.go
otherview: Good q. I guess the status can be a subset of the health check. Like the HealthCheck can provide that info. 
But extending feels like it would be mixing the logic of whats healthy and at what stage are things.

go/enclave/enclave.go
otherview: You mean add a method in the storage ? Yeah that makes sense. What should the method do ? Fetch the latest rollup ?

go/common/enclave.go
joeldudley: Ok

go/enclave/enclave.go
joeldudley: I'd say try and read something. Even if the read returns an empty result (e.g. reading a dummy hash), you know it's successfully able to speak to the DB.

go/enclave/db/storage.go
```diff
@@ -199,6 +199,11 @@ func (s *storageImpl) IsBlockAncestor(block *types.Block, maybeAncestor common.L
 	return s.IsBlockAncestor(p, maybeAncestor)
 }
 
+func (s *storageImpl) HealthCheck() (bool, error) {
+	headRollup := s.FetchHeadRollup()
```
joeldudley, line 203: @otherview I've checked and `FetchHeadRollup` calls `ReadHeadRollupHash`, which actually swallows any errors. Can you change that method to actually return an error? Then I think he check here should be whether an error is returned, not whether the head rollup is nil (otherwise, there's the edge case of the healthcheck being performed too early).

go/obsclient/obsclient.go
```diff
@@ -80,3 +80,10 @@ func (oc *ObsClient) RollupHeaderByHash(hash gethcommon.Hash) (*common.Header, e
 	}
 	return rollupHeader, err
 }
+
+// Health returns the health of the node.
+func (oc *ObsClient) Health() (bool, error) {
```
joeldudley, line 85: Should this be `IsHealthy`? Your call.

go/enclave/db/storage.go
otherview: That's *one* reason why I left the healthcheck simple. The Storage is swallowing a bunch of errors ( same thing in the StoreRollup ). I'm not sure why we're not surfacing errors, maybe we were panicking before and it was all in mem  ? 

I can do that in a new PR  though. Maybe changing the error handling on the storage package ?

go/obsclient/obsclient.go
otherview: I would like to be returning a health json object in the future, depending on how we see it used on datadog.

go/enclave/db/storage.go
joeldudley: I guess so. It just seems dumb to have a healthcheck that pings the DB, except that it swallows any errors (I get that it works here, assuming a first rollup has been produced). Why don't we add a todo for now?

joeldudley (approved):
I think we need that todo, but otherwise approved.

go/enclave/db/storage.go
otherview: Agreed. Let me add a todo 👍

PR #829, opened 2022-11-15
Fixes grabbing a rollup via ObscuroScan

joeldudley:

Previously, there was a dedicated ObscuroScan API to get a rollup by number. I removed this and replaced it with a call to get the rollup hash, followed by a call to get the rollup itself, to shrink the size of the API. But this surfaced an error in the calculation of hashes in our headers once returned to the client.

### DIFF
--- a/go/common/headers.go
+++ b/go/common/headers.go
@@ -23,7 +23,7 @@ type Header struct {
 	ParentHash  L2RootHash
 	Agg         common.Address
 	Nonce       types.BlockNonce // Nonce ensure compatibility with ethereum
-	L1Proof     L1RootHash       // the L1 block where the Parent was published
+	L1Proof     L1RootHash       // the L1 block where the parent was published
 	Root        StateRoot
 	TxHash      common.Hash // todo - include the synthetic deposits
 	Number      *big.Int    // the rollup height
@@ -32,6 +32,7 @@ type Header struct {
 	Extra       []byte
 	R, S        *big.Int // signature values
 	Withdrawals []Withdrawal
+	Hash        common.Hash
 
 	// Specification fields - not used for now but are expected to be available
 	UncleHash  common.Hash    `json:"sha3Uncles"`
@@ -41,8 +42,7 @@ type Header struct {
 	GasUsed    uint64         `json:"gasUsed"    `
 	Time       uint64         `json:"timestamp"   `
 	MixDigest  common.Hash    `json:"mixHash"`
-	// BaseFee was added by EIP-1559 and is ignored in legacy headers.
-	BaseFee *big.Int `json:"baseFeePerGas"`
+	BaseFee    *big.Int       `json:"baseFeePerGas"` // BaseFee was added by EIP-1559 and is ignored in legacy headers.
 }
 
 // Withdrawal - this is the withdrawal instruction that is included in the rollup header.
@@ -59,12 +59,13 @@ type HeaderWithTxHashes struct {
 	TxHashes []TxHash
 }
 
-// Hash returns the block hash of the header, which is simply the keccak256 hash of its
+// CalcHash returns the block hash of the header, which is simply the keccak256 hash of its
 // RLP encoding excluding the signature.
-func (h *Header) Hash() L2RootHash {
+func (h *Header) CalcHash() L2RootHash {
 	cp := *h
 	cp.R = nil
 	cp.S = nil
+	cp.Hash = common.Hash{}
 	hash, err := rlpHash(cp)
 	if err != nil {
 		panic("err hashing a rollup header")
--- a/go/common/headers.go
+++ b/go/common/headers.go
@@ -20,29 +20,33 @@ var hasherPool = sync.Pool{
 // Header is a public / plaintext struct that holds common properties of rollups batches.
 // Making changes to this struct will require GRPC + GRPC Converters regen
 type Header struct {
+	// The fields present in Geth's `types/Header` struct.
 	ParentHash  L2RootHash
-	Agg         common.Address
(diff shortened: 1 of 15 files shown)


go/common/headers.go
```diff
@@ -20,29 +20,33 @@ var hasherPool = sync.Pool{
 // Header is a public / plaintext struct that holds common properties of rollups batches.
 // Making changes to this struct will require GRPC + GRPC Converters regen
 type Header struct {
```
joeldudley, line 22: I reordered the fields here, to first have the ones in Geth's `types/Header`, then the custom fields. This makes it easier to understand what we're adding to the Geth type.

go/common/headers.go
```diff
@@ -20,29 +20,33 @@ var hasherPool = sync.Pool{
...
-	Difficulty *big.Int       `json:"difficulty" `
-	GasLimit   uint64         `json:"gasLimit"  `
-	GasUsed    uint64         `json:"gasUsed"    `
-	Time       uint64         `json:"timestamp"   `
-	MixDigest  common.Hash    `json:"mixHash"`
-	// BaseFee was added by EIP-1559 and is ignored in legacy headers.
-	BaseFee *big.Int `json:"baseFeePerGas"`
+	// A precalculated hash to use instead of `CalcHash` on the client side.
+	// TODO - See if we can forgo this and calculate the hash on the client side
+	Hash common.Hash
```
joeldudley, line 49: As discussed - we store a hash to be used on the client side. Note that Geth does the same, though implicitly, as part of the JSON serialisation.

go/host/rpc/clientapi/client_api_eth.go
```diff
@@ -184,13 +183,15 @@ func headerWithHashesToMap(headerWithHashes *common.HeaderWithTxHashes) map[stri
...
 		"number":           header.Number.Uint64(),
-		"hash":             header.Hash(),
+		"hash":             header.CalcHash(),
 		"parentHash":       header.ParentHash,
 		"nonce":            header.Nonce,
 		"logsBloom":        header.Bloom,
 		"stateRoot":        header.Root,
 		"receiptsRoot":     header.ReceiptHash,
 		"miner":            header.Agg,
+		"coinbase":         header.Agg,
```
joeldudley, line 193: Added a few missing fields.

go/host/rpc/clientapi/client_api_eth.go
```diff
@@ -183,27 +182,29 @@ func (api *EthereumAPI) FeeHistory(context.Context, rpc.DecimalOrHex, rpc.BlockN
 func headerWithHashesToMap(headerWithHashes *common.HeaderWithTxHashes) map[string]interface{} {
 	header := headerWithHashes.Header
 	return map[string]interface{}{
```
joeldudley, line 184: Reorders these fields to match their order in the `common.Header` struct, and adds a few missing fields.

go/common/headers.go
```diff
@@ -59,12 +63,13 @@ type HeaderWithTxHashes struct {
...
 
-// Hash returns the block hash of the header, which is simply the keccak256 hash of its
+// CalcHash returns the block hash of the header, which is simply the keccak256 hash of its
 // RLP encoding excluding the signature.
-func (h *Header) Hash() L2RootHash {
+func (h *Header) CalcHash() L2RootHash {
 	cp := *h
 	cp.R = nil
 	cp.S = nil
+	cp.Hash = common.Hash{}
```
BedrockSquirrel, line 72: Worth a comment here to say you always use a copy with a zeroed hash to avoid circular calculation.

joeldudley: Fixed

go/common/headers.go
```diff
@@ -20,29 +20,29 @@ var hasherPool = sync.Pool{
...
-	L1Proof     L1RootHash       // the L1 block used by the enclave to generate the current rollup
-	Root        StateRoot
-	TxHash      common.Hash // todo - include the synthetic deposits
-	Number      *big.Int    // the rollup height
-	Bloom       types.Bloom
-	ReceiptHash common.Hash
-	Extra       []byte
-	R, S        *big.Int // signature values
-	Withdrawals []Withdrawal
+	UncleHash   common.Hash    `json:"sha3Uncles"`
```
joeldudley, line 25: We have to be careful to have these JSON annotations wherever the field name diverges from what's in the JSON object returned by the Ethereum API. Otherwise, fields will end up nulled out when deserialised.

go/rpc/client.go
```diff
@@ -24,7 +24,6 @@ const (
 	AddViewingKey         = "obscuro_addViewingKey"
 	Health                = "obscuro_health"
 	GetBlockHeaderByHash  = "obscuroscan_getBlockHeaderByHash"
-	GetHeadRollupHeader   = "obscuroscan_getHeadRollupHeader"
```
joeldudley, line 27: Was no longer used.

integration/simulation/validate_chain.go
```diff
@@ -313,6 +313,15 @@ func checkBlockchainOfObscuroNode(t *testing.T, rpcHandles *network.RPCHandles,
 	// (execute deposits and transactions and compare to the state in the rollup)
 
 	heights[nodeIdx] = l2Height.Uint64()
+
+	// check that the headers are serialised and deserialised correctly, by recomputing a header's hash
+	parentHeader, err := obscuroClient.RollupHeaderByHash(headRollupHeader.ParentHash)
```
joeldudley, line 318: Added a check in the sims to ensure we don't backslide.

BedrockSquirrel (approved):
LGTM

PR #832, opened 2022-11-15
Fixing node start out of sync

otherview:

- https://github.com/obscuronet/obscuro-internal/issues/1182
- Adding more logs to figure out what's happening
- Non-sequencer nodes start is now delayed by 1 minute in testnet / dev testnet deployments


moraygrieve (approved):
Is there any way to wait on a signal to know it is ready (rather than a default sleep of 60 seconds)? It would either have to be waiting on a log statement (which is probably hard when it goes to stdout within the container) or some form of health endpoint that returns true when it is ready. One for later maybe ... 


LGTM

otherview:
> Is there any way to wait on a signal to know it is ready (rather than a default sleep of 60 seconds)? It would either have to be waiting on a log statement (which is probably hard when it goes to stdout within the container) or some form of health endpoint that returns true when it is ready. One for later maybe ...
> 
> LGTM

Yes, potentially the health endpoint can be checked beforehand ? Maybe one for later

PR #836, opened 2022-11-16
HeaderWithHashes is in the `common` module, but it's specific to the host DB

joeldudley:

HeaderWithHashes is in the `common` module, but it's specific to the host DB. I've moved it across for clarity.

### DIFF
--- a/go/host/db/hostdb.go
+++ b/go/host/db/hostdb.go
@@ -108,6 +108,7 @@ func (db *DB) AddRollupHeader(header *common.Header, txHashes []common.TxHash) {
 	b := db.kvStore.NewBatch()
 	headerWithHashes := &HeaderWithTxHashes{Header: header, TxHashes: txHashes}
 	db.writeRollupHeader(b, headerWithHashes)
+	db.writeRollupTxHashes(b, header.Hash(), txHashes) // Required by ObscuroScan, to display a list of recent transactions.
 	db.writeRollupHash(b, headerWithHashes.Header)
 	for _, txHash := range headerWithHashes.TxHashes {
 		db.writeRollupNumber(b, headerWithHashes.Header, txHash)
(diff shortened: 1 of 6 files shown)


go/host/db/hostdb.go
```diff
@@ -83,22 +89,24 @@ func (db *DB) AddBlockHeader(header *types.Header) {
 }
 
 // GetHeadRollupHeader returns the header of the current rollup (head) of the Node, or (nil, false) if no such header is found.
-func (db *DB) GetHeadRollupHeader() (*common.HeaderWithTxHashes, bool) {
+func (db *DB) GetHeadRollupHeader() (*common.Header, bool) {
```
joeldudley, line 92: No need to return the tx hashes here.

go/host/host.go
```diff
@@ -531,8 +531,7 @@ func (h *host) storeBlockProcessingResult(result *common.BlockSubmissionResponse
 	// only update the host rollup headers if the enclave has found a new rollup head
 	if result.FoundNewHead {
 		// adding a header will update the head if it has a higher height
-		headerWithHashes := common.HeaderWithTxHashes{Header: result.IngestedRollupHeader, TxHashes: result.ProducedRollup.TxHashes}
-		h.db.AddRollupHeader(&headerWithHashes)
+		h.db.AddRollupHeader(result.IngestedRollupHeader, result.ProducedRollup.TxHashes)
```
BedrockSquirrel, line 534: While you're here could you add a todo saying there's a bug here and that the tx hashes are being stored against the header for a different rollup?

There's a task for it but I think it's worth noting in the code to help avoid confusion for readers.

BedrockSquirrel (approved):
LGTM, one request that's barely related to the PR

BedrockSquirrel:
LGTM still - would be a good idea to add that todo though imo

PR #837, opened 2022-11-16
Adds methods to host DB for batches

joeldudley:

We need methods for reading and writing batches. These are very similar to the rollup r/w methods.

I also split the `hostdb.go` file between block, rollup and batch files, as it was getting too large. I'd like to do a follow-up PR, creating three separate interfaces for these.

The block and rollup methods are pretty much unchanged, I'd say only the batch stuff needs to be reviewed.


go/host/db/batches.go
```diff
@@ -0,0 +1,112 @@
...
+// DB methods relating to batches.
+
+// headerKey = batchHeaderPrefix  + hash
+func batchHeaderKey(hash gethcommon.Hash) []byte {
+	return append(batchHeaderPrefix, hash.Bytes()...)
+}
+
+// GetHeadBatchHeader returns the header of the node's current head batch, or (nil, false) if no such header is found.
+func (db *DB) GetHeadBatchHeader() (*common.Header, bool) {
+	headBatchHash, found := db.readHeadBatchHash(db.kvStore)
```
otherview, line 23: No need to have a method of type and pass a property of type.
```suggestion
	headBatchHash, found := readHeadBatchHash(db.kvStore)
```
OR
```suggestion
	headBatchHash, found := db.readHeadBatchHash()
```

go/host/db/batches.go
```diff
@@ -0,0 +1,112 @@
...
+
+// DB methods relating to batches.
+
+// headerKey = batchHeaderPrefix  + hash
+func batchHeaderKey(hash gethcommon.Hash) []byte {
+	return append(batchHeaderPrefix, hash.Bytes()...)
+}
+
+// GetHeadBatchHeader returns the header of the node's current head batch, or (nil, false) if no such header is found.
+func (db *DB) GetHeadBatchHeader() (*common.Header, bool) {
```
otherview, line 22: Optional, but it's likely better to return errors instead of booleans.

go/host/db/batches.go
```diff
@@ -0,0 +1,112 @@
...
+
+// AddBatchHeader adds a batch's header to the known headers
+func (db *DB) AddBatchHeader(header *common.Header, txHashes []common.TxHash) {
+	b := db.kvStore.NewBatch()
+	db.writeBatchHeader(b, header)
+
+	// TODO - #718 - Store the batch txs, batch hash, and batch number per transaction hash, if needed (see `AddRollupHeader`).
+
+	// There's a potential race here, but absolute accuracy of the number of transactions is not required.
+	currentTotal := db.readTotalTransactions(db.kvStore)
```
otherview, line 38: Same coment as above

go/host/db/batches.go
```diff
@@ -0,0 +1,112 @@
...
+	db.writeTotalTransactions(b, newTotal)
+
+	// update the head if the new height is greater than the existing one
+	headBatchHeader, found := db.GetHeadBatchHeader()
+	if !found || headBatchHeader.Number.Int64() <= header.Number.Int64() {
+		db.writeHeadBatchHash(b, header.Hash())
+	}
+
+	if err := b.Write(); err != nil {
+		db.logger.Crit("Could not write batch.", log.ErrKey, err)
```
otherview, line 49: Should the error be returned here ?

go/host/db/batches.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+}
+
+// Retrieves the batch header corresponding to the hash, or (nil, false) if no such header is found.
+func (db *DB) readBatchHeader(r ethdb.KeyValueReader, hash gethcommon.Hash) (*common.Header, bool) {
+	f, err := r.Has(batchHeaderKey(hash))
+	if err != nil {
+		db.logger.Crit("could not retrieve batch header.", log.ErrKey, err)
```
otherview, line 57: Error should be returned

go/host/db/batches.go
```diff
@@ -0,0 +1,112 @@
...
+		db.writeHeadBatchHash(b, header.Hash())
+	}
+
+	if err := b.Write(); err != nil {
+		db.logger.Crit("Could not write batch.", log.ErrKey, err)
+	}
+}
+
+// Retrieves the batch header corresponding to the hash, or (nil, false) if no such header is found.
+func (db *DB) readBatchHeader(r ethdb.KeyValueReader, hash gethcommon.Hash) (*common.Header, bool) {
```
otherview, line 54: Optionally, `errutil.ErrNotFound` can be returned for when the resource is not found

go/host/db/batches.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	header := new(common.Header)
+	if err := rlp.Decode(bytes.NewReader(data), header); err != nil {
+		db.logger.Crit("could not decode batch header.", log.ErrKey, err)
+	}
+	return header, true
+}
+
+// Retrieves the hash of the head batch.
+func (db *DB) readHeadBatchHash(r ethdb.KeyValueReader) (*gethcommon.Hash, bool) {
```
otherview, line 77: Error swallowing

go/host/db/batches.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/ethereum/go-ethereum/ethdb"
+	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/obscuronet/go-obscuro/go/common"
+	"github.com/obscuronet/go-obscuro/go/common/log"
+)
+
+// DB methods relating to batches.
+
+// headerKey = batchHeaderPrefix  + hash
+func batchHeaderKey(hash gethcommon.Hash) []byte {
```
otherview, line 17: Style pref: typically private methods go after the public methods.

go/host/db/batches.go
```diff
@@ -0,0 +1,112 @@
+package db
```
otherview, line 1: This should have a unit test file to accompany it. It's nicely written and decoupled should be easy to inject mock objects and test the batches logic

go/host/db/blocks.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/ethdb"
+	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/obscuronet/go-obscuro/go/common/log"
+)
+
+// DB methods relating to batches.
+
+// headerKey = blockHeaderPrefix  + hash
+func blockHeaderKey(hash gethcommon.Hash) []byte {
```
otherview, line 16: typically private methods are after the public ones

go/host/db/blocks.go
```diff
@@ -0,0 +1,107 @@
...
+	db.writeBlockHeader(b, header)
+
+	// update the head if the new height is greater than the existing one
+	headBlockHeader, found := db.GetHeadBlockHeader()
+	if !found || headBlockHeader.Number.Int64() <= header.Number.Int64() {
+		db.writeHeadBlockHash(b, header.Hash())
+	}
+
+	if err := b.Write(); err != nil {
+		db.logger.Crit("Could not write rollup.", log.ErrKey, err)
```
otherview, line 46: I don't think we should swallow errors.

go/host/db/blocks.go
```diff
@@ -0,0 +1,107 @@
...
+		db.logger.Crit("could not retrieve head block.", log.ErrKey, err)
+	}
+	if !f {
+		return nil
+	}
+	value, err := r.Get(headBlock)
+	if err != nil {
+		db.logger.Crit("could not retrieve head block.", log.ErrKey, err)
+	}
+	h := gethcommon.BytesToHash(value)
```
otherview, line 105: Can `r.Get(headBlock)` return nil ?

go/host/db/rollups.go
```diff
@@ -0,0 +1,278 @@
...
+
+// GetRollupHeader returns the rollup header given the Hash, or (nil, false) if no such header is found.
+func (db *DB) GetRollupHeader(hash gethcommon.Hash) (*common.Header, bool) {
+	return db.readRollupHeader(db.kvStore, hash)
+}
+
+// AddRollupHeader adds a rollup's header to the known headers
+func (db *DB) AddRollupHeader(header *common.Header, txHashes []common.TxHash) {
+	b := db.kvStore.NewBatch()
+	db.writeRollupHeader(b, header)
```
otherview, line 55: If one write fails should the other rollback ?

go/host/db/blocks.go
joeldudley: Not swallowed - Crit() causes the program to exit.

go/host/db/batches.go
joeldudley: Not swallowed - Crit() causes the program to exit.

go/host/db/batches.go
joeldudley: Not swallowed - Crit() causes the program to exit.

go/host/db/batches.go
joeldudley: Fixed

go/host/db/batches.go
joeldudley: Thanks, fixing.

go/host/db/blocks.go
joeldudley: Fixing.

go/host/db/batches.go
otherview: Sorry, I mean that the error isn't being returned, swallowed was the wrong word, but it was in the sense that the caller doesn't have to handle it.

joeldudley: Yeah, the approach we’ve chosen here is for the process to exit if the db is unreachable or has a critical issue. Maybe I can add a todo to handle critical db errors gracefully?

go/host/db/blocks.go
joeldudley: Only if the err is not nil, which we check.

go/host/db/rollups.go
joeldudley: I think so. Do you think we shouldn't?

go/host/db/batches.go
joeldudley: Raised https://github.com/obscuronet/obscuro-internal/issues/1195. Will tackle after.

go/host/db/batches.go
joeldudley: Raised https://github.com/obscuronet/obscuro-internal/issues/1195. Will tackle after.

go/host/db/batches.go
joeldudley: Raised https://github.com/obscuronet/obscuro-internal/issues/1195. Will tackle after.

go/host/db/batches.go
joeldudley: Raised https://github.com/obscuronet/obscuro-internal/issues/1195. Will tackle after.

go/host/db/blocks.go
joeldudley: Raised https://github.com/obscuronet/obscuro-internal/issues/1195. Will tackle after.

go/host/db/rollups.go
otherview: I think we should. But it seems like at the moment if one write fails the approach is to panic. Is that intended yeah ?

otherview (approved):
I think there are a few todos on unit testing and error handling, other than that lgtm 👍

go/host/db/rollups.go
joeldudley: Yeah - you want all the writes to succeed or fail together.

otherview: I understand that's implicit, but what if there was a TODO mentioning that ? At the moment if one write fails the whole node stops, it's not an explicit logic of what its expected to happen.

joeldudley: That'll be picked up by the follow-up ticket, if that's ok

go/host/db/blocks_test.go
```diff
@@ -0,0 +1,89 @@
+package db
+
+import (
+	"math/big"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+const magicNumber = 777
```
otherview, line 10: Gonna guess the magic number is the minimum required for making a header ? Can a comment be added to it ?

otherview (approved):
Would add a couple of comments, lgtm

go/host/db/blocks_test.go
joeldudley: No, it's just a field to check is still there when the object comes back from the DB. Will rename.

PR #838, opened 2022-11-16
Fix issues with submit block errors

BedrockSquirrel:

- The block feeding changes I'm making rely on the enclave reporting back what it knows as the L1 head with failures
- There were too many spammy logs from the parent block failures

- Fix the rpc conversion to produce the correct error type from the BlockSubmissionResponse object
- Change the log level to debug for those failures when they happen for 'parent' blocks in the main path (blocks that have usually already been processed and so will always fail as a duplicate)

I verified in the logs that I'm now seeing the hashes for the enclave's view of the L1 head and no longer seeing the spam of "could not submit block" "block already processed"

### DIFF
--- a/go/enclave/rpc_server.go
+++ b/go/enclave/rpc_server.go
@@ -124,11 +124,11 @@ func (s *server) SubmitBlock(_ context.Context, request *generated.SubmitBlockRe
 	bl := s.decodeBlock(request.EncodedBlock)
 	blockSubmissionResponse, err := s.enclave.SubmitBlock(bl, request.IsLatest)
 	if err != nil {
-		var rejErr common.BlockRejectError
+		var rejErr *common.BlockRejectError
 		isReject := errors.As(err, &rejErr)
 		if isReject {
 			// todo: we should avoid errors in response messages and use the gRPC error objects for this stuff (standardized across all enclave responses)
-			msg, err := rpc.ToBlockSubmissionRejectionMsg(&rejErr)
+			msg, err := rpc.ToBlockSubmissionRejectionMsg(rejErr)
 			if err == nil {
 				// send back reject err response
 				return &generated.SubmitBlockResponse{BlockSubmissionResponse: &msg}, nil
--- a/go/host/host.go
+++ b/go/host/host.go
@@ -404,11 +404,22 @@ func (h *host) startProcessing() {
 			roundInterrupt = triggerInterrupt(roundInterrupt)
 			err := h.processBlock(b.p, false)
 			if err != nil {
-				h.logger.Info("Could not process parent block. ", log.ErrKey, err)
+				var rejErr *common.BlockRejectError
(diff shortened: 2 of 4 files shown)


go/host/host.go
```diff
@@ -385,8 +385,8 @@ func (h *host) waitForEnclave() {
 }
 
 // starts the host main processing loop
-func (h *host) startProcessing() {
-	h.logger.Info("§§§ Start processing...")
+// todo: matt to fix this complexity `nolint` in next PR with block provider
+func (h *host) startProcessing() { //nolint:gocognit
```
BedrockSquirrel, line 389: This method gets a lot simpler with my block provider wiring PR that's in flight so I'm just ignoring the complexity warning for now.

go/host/host.go
```diff
@@ -889,7 +900,7 @@ func (h *host) bootstrapHost() types.Block {
 		result, err := h.enclaveClient.SubmitBlock(cb, false)
 		if err != nil {
 			var bsErr *common.BlockRejectError
-			isBSE := errors.As(err, bsErr)
+			isBSE := errors.As(err, &bsErr)
```
tudor-malene, line 903: isn't this the address of the pointer?

go/enclave/rpc_server.go
```diff
@@ -124,11 +124,11 @@ func (s *server) SubmitBlock(_ context.Context, request *generated.SubmitBlockRe
 	bl := s.decodeBlock(request.EncodedBlock)
 	blockSubmissionResponse, err := s.enclave.SubmitBlock(bl, request.IsLatest)
 	if err != nil {
-		var rejErr common.BlockRejectError
+		var rejErr *common.BlockRejectError
 		isReject := errors.As(err, &rejErr)
```
tudor-malene, line 128: is this right? if rejErr is already a pointer?

go/host/host.go
```diff
@@ -404,11 +404,22 @@ func (h *host) startProcessing() {
 			roundInterrupt = triggerInterrupt(roundInterrupt)
 			err := h.processBlock(b.p, false)
 			if err != nil {
-				h.logger.Info("Could not process parent block. ", log.ErrKey, err)
+				var rejErr *common.BlockRejectError
+				if errors.As(err, &rejErr) {
```
tudor-malene, line 408: ditto

tudor-malene (commented):
looks good in principle. 
Had a few questions about pointer stuff

go/host/host.go
BedrockSquirrel: Yeah I found it weird too but actually this was a bug here before and if you look at the usage in online examples it always seems to be done this way, I sent some links on discord including this example from official docs:

Screenshot 2022-11-17 at 10 37 50

tudor-malene (approved):
lgtm

PR #851, opened 2022-11-21
Allow obcuroscan to be run against a local testnet

moraygrieve:

- Adds the build of the obscuroscan image into the local testnet build, provides a script to start up obscuroscan locally, and updates the readme accordingly. 

- The top level readme, and build and runs script in the testnet folder


tudor-malene (commented):
my docker-fu is limited. Looks reasonable.
Does this also include the datadog agent?

moraygrieve:
This is all just for a local testnet run so no integration to datadog for that. At the moment the only version of obscuroscan we support is the main testnet, which means I cannot test prior to releasing. This just allows me to build and run the container locally

tudor-malene (approved):
Thanks Moray.
LGTM

PR #861, opened 2022-11-22
Set correct timeout for giving up on L1 reconnection

BedrockSquirrel:

- We were using the context timeouts as the max time before giving up so gave up after 10-15s instead of a longer period
- This should help us reconnect if there are L1 "blips" but won't help with recovery for more serious issues

- Fix timeout


tudor-malene (approved):
lgtm

PR #870, opened 2022-11-23
Adds integration test of getTotalTxs.

joeldudley:

We need integration tests of the Obscuroscan APIs to stop Obscuroscan from breaking silently.

### DIFF
--- a/integration/simulation/validate_chain.go
+++ b/integration/simulation/validate_chain.go
@@ -46,7 +46,15 @@ const (
 // For example, all injected transactions were processed correctly, the height of the rollup chain is a function of the total
 // time of the simulation and the average block duration, that all Obscuro nodes are roughly in sync, etc
 func checkNetworkValidity(t *testing.T, s *Simulation) {
-	// ensure L1 and L2 txs were issued
+	checkTransactionsInjected(t, s)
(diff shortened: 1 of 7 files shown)


integration/simulation/validate_chain.go
```diff
@@ -561,3 +565,17 @@ func assertNoDupeLogs(t *testing.T, logs []*types.Log) {
 		}
 	}
 }
+
+// Checks that the various APIs powering Obscuroscan are working correctly.
+func checkObscuroscan(t *testing.T, s *Simulation) {
+	for idx, client := range s.RPCHandles.RPCClients {
+		var totalTxs *big.Int
+		err := client.Call(&totalTxs, rpc.GetTotalTxs)
```
BedrockSquirrel, line 573: Can we add this method to the ObscuroClient rather than using the raw RPC client? When we're calling from Go we should never have to work directly with the RPC calls imo.

Even if we're not sure how we want this method to look eventually, this RPC endpoint is publicly available so adding it to the client means we don't have to duplicate the calling logic and it's discoverable.

BedrockSquirrel (changes requested):
LGTM but can we avoid using the RPC client directly, make the method available on the actual client.

BedrockSquirrel (approved):
LGTM

PR #881, opened 2022-11-24
Fixes NPE when receipt takes longer than expected time

otherview:

- As per title, a return was needed otherwise `receipt` would create a NPE at the trace operation

### DIFF
--- a/go/host/host.go
+++ b/go/host/host.go
@@ -650,15 +650,20 @@ func (h *host) signAndBroadcastL1Tx(tx types.TxData, tries uint64) error {
 func (h *host) watchForReceipt(txHash common.TxHash) {
 	var receipt *types.Receipt
 	var err error
-	err = retry.Do(func() error {
-		receipt, err = h.ethClient.TransactionReceipt(txHash)
-		return err
-	}, retry.NewTimeoutStrategy(maxWaitForL1Receipt, retryIntervalForL1Receipt))
+	err = retry.Do(
+		func() error {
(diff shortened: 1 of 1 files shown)


go/host/host.go
```diff
@@ -650,15 +650,20 @@ func (h *host) signAndBroadcastL1Tx(tx types.TxData, tries uint64) error {
 func (h *host) watchForReceipt(txHash common.TxHash) {
 	var receipt *types.Receipt
 	var err error
-	err = retry.Do(func() error {
-		receipt, err = h.ethClient.TransactionReceipt(txHash)
-		return err
-	}, retry.NewTimeoutStrategy(maxWaitForL1Receipt, retryIntervalForL1Receipt))
+	err = retry.Do(
```
BedrockSquirrel, line 653: Minor (subjective) thing but I preferred the formatting as it was before tbh.

When I see functions as parameters in Go code it usually seems to be like that, with the anonymous func started inline to reduce the indentation, like javascript-style (e.g. the `sub.Receive()` example in the [pubsub callback Go docs](https://pkg.go.dev/cloud.google.com/go/pubsub)).

BedrockSquirrel (approved):
LGTM, thanks for fixing! 😳

go/host/host.go
BedrockSquirrel: Eh, looking around a bit more I see both styles haha. It's probably just what I'm used to from JS.

otherview: I typically put one thing per line.. I think this is a job for linter :)

PR #882, opened 2022-11-24
Surface errors from accessors_receipts and accessors_metadata

joeldudley:

### DIFF
--- a/go/enclave/db/rawdb/accessors_indexes.go
+++ b/go/enclave/db/rawdb/accessors_indexes.go
@@ -110,7 +110,10 @@ func ReadReceipt(db ethdb.Reader, hash common.Hash, config *params.ChainConfig, 
 		return nil, common.Hash{}, 0, 0
 	}
 	// Read all the receipts from the block and return the one with the matching hash
-	receipts := ReadReceipts(db, blockHash, *blockNumber, config, logger)
+	receipts, err := ReadReceipts(db, blockHash, *blockNumber, config, logger)
+	if err != nil {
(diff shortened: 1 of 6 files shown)


go/enclave/db/rawdb/accessors_indexes.go
```diff
@@ -110,7 +110,10 @@ func ReadReceipt(db ethdb.Reader, hash common.Hash, config *params.ChainConfig,
 		return nil, common.Hash{}, 0, 0
 	}
 	// Read all the receipts from the block and return the one with the matching hash
-	receipts := ReadReceipts(db, blockHash, *blockNumber, config, logger)
+	receipts, err := ReadReceipts(db, blockHash, *blockNumber, config, logger)
+	if err != nil {
+		logger.Error(fmt.Sprintf("Receipt could not be retrieved %s = %d; %s = %s; %s = %s;", "number", *blockNumber, "hash", blockHash, "txhash", hash))
```
BedrockSquirrel, line 115: The key=value thing is what the gethlogger does by default isn't it? So you can just have like:
`logger.Error("Receipt could not be retrieved", "number", *blockNumber, "hash", blockHash, "txhash", hash)`
 I just noticed there's others like this around this file, maybe you could do them too if it's not too annoying?

Also, I think datadog provides useful filtering ability on the key=value format metadata, which we probably break by introducing spaces in between.

BedrockSquirrel (commented):
LGTM but I think we should use the standard log formatting mechanism for the messages in there ideally

BedrockSquirrel (approved):
LGTM thanks

PR #885, opened 2022-11-25
Make our ErrNotFound match Ethereum's

joeldudley:

Currently, we use a lot of Ethereum components. Some of these may be throwing out their own `ethereum.NotFound` errors which we are failing to convert to `errutil.ErrNotFound`. This is a potential source of bugs, as we may have an `errutil.ErrNotFound` check that misses an unconverted `ethereum.NotFound` error.

### DIFF
--- a/go/common/errutil/errors_util.go
+++ b/go/common/errutil/errors_util.go
@@ -1,8 +1,12 @@
 package errutil
 
-import "errors"
+import (
+	"errors"
+
+	"github.com/ethereum/go-ethereum"
+)
 
 var (
-	ErrNotFound = errors.New("not found")
+	ErrNotFound = ethereum.NotFound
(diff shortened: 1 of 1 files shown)


go/common/errutil/errors_util.go
```diff
@@ -1,8 +1,12 @@
...
-import "errors"
+import (
+	"errors"
+
+	"github.com/ethereum/go-ethereum"
+)
 
 var (
-	ErrNotFound = errors.New("not found")
+	ErrNotFound = ethereum.NotFound
```
BedrockSquirrel, line 10: Can you put a comment above this saying why it's important that it's an alias just on the off-chance someone does try and change this.

otherview (approved):
Can you add a comment stating why we're doing this aliases please ? Future Pedro thanks you

PR #887, opened 2022-11-25
Allow hosts to catch up on missed batches

joeldudley:

A node can easily miss batches, e.g. due to a network issue, or because they started up late. This PR introduces a catch-up mechanism.

The idea is that when you receive a batch, you check whether you have it's full history. If you are missing any historical batches, you discard the batch and ask the sequencer to send over all the batches up to and including this new batch.

### DIFF
--- a/go/common/host/host.go
+++ b/go/common/host/host.go
@@ -55,7 +55,7 @@ type P2P interface {
 	UpdatePeerList([]string)
 	BroadcastTx(tx common.EncryptedTx) error
 	BroadcastBatch(batch *common.ExtBatch) error
-	RequestBatchesSince(batchNumber *big.Int) ([]*common.ExtBatch, error)
+	RequestBatches(from *big.Int, to *big.Int) ([]*common.ExtBatch, error)
 }
 
 // ReconnectingBlockProvider interface allows host to monitor and await L1 blocks.
--- a/go/common/host/host.go
+++ b/go/common/host/host.go
@@ -23,8 +23,8 @@ type Host interface {
 	SubmitAndBroadcastTx(encryptedParams common.EncryptedParamsSendRawTx) (common.EncryptedResponseSendRawTx, error)
 	// ReceiveTx processes a transaction received from a peer host.
 	ReceiveTx(tx common.EncryptedTx)
-	// ReceiveBatch processes a batch received from a peer host.
-	ReceiveBatch(batch common.EncodedBatch)
+	// ReceiveBatches receives a set of batches from a peer host.
+	ReceiveBatches(batches common.EncodedBatches)
 	// Subscribe feeds logs matching the encrypted log subscription to the matchedLogs channel.
 	Subscribe(id rpc.ID, encryptedLogSubscription common.EncryptedParamsLogSubscription, matchedLogs chan []byte) error
 	// Unsubscribe terminates a log subscription between the host and the enclave.
--- a/go/common/host/host.go
+++ b/go/common/host/host.go
@@ -57,7 +57,7 @@ type P2P interface {
 	UpdatePeerList([]string)
 	BroadcastTx(tx common.EncryptedTx) error
 	BroadcastBatch(batch *common.ExtBatch) error
-	RequestBatches(from *big.Int, to *big.Int) error
+	RequestBatches(batchRequest *common.BatchRequest) error
 }
 
 // ReconnectingBlockProvider interface allows host to monitor and await L1 blocks.
--- a/go/common/host/host.go
+++ b/go/common/host/host.go
@@ -61,8 +61,8 @@ type P2P interface {
 	BroadcastBatch(batch *common.ExtBatch) error
 	// RequestBatches requests batches from the sequencer.
 	RequestBatches(batchRequest *common.BatchRequest) error
-	// SendBatch sends batches to a specific node, in response to a batch request.
-	SendBatch(batches []*common.ExtBatch, to *gethcommon.Address) error
+	// SendBatches sends batches to a specific node, in response to a batch request.
+	SendBatches(batches []*common.ExtBatch, to string) error
 }
 
 // ReconnectingBlockProvider interface allows host to monitor and await L1 blocks.
--- a/go/common/host/host.go
+++ b/go/common/host/host.go
@@ -25,6 +25,8 @@ type Host interface {
 	ReceiveTx(tx common.EncryptedTx)
 	// ReceiveBatches receives a set of batches from a peer host.
 	ReceiveBatches(batches common.EncodedBatches)
+	// SendBatches sends a set of batches to a peer host. Used during catch-up.
+	SendBatches(batchRequest common.EncodedBatchRequest)
 	// Subscribe feeds logs matching the encrypted log subscription to the matchedLogs channel.
 	Subscribe(id rpc.ID, encryptedLogSubscription common.EncryptedParamsLogSubscription, matchedLogs chan []byte) error
 	// Unsubscribe terminates a log subscription between the host and the enclave.
--- a/go/common/host/host.go
+++ b/go/common/host/host.go
@@ -55,9 +55,14 @@ type P2P interface {
 	StartListening(callback Host)
 	StopListening() error
 	UpdatePeerList([]string)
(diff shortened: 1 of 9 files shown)


go/common/host/host.go
```diff
@@ -23,8 +23,8 @@ type Host interface {
 	SubmitAndBroadcastTx(encryptedParams common.EncryptedParamsSendRawTx) (common.EncryptedResponseSendRawTx, error)
 	// ReceiveTx processes a transaction received from a peer host.
 	ReceiveTx(tx common.EncryptedTx)
-	// ReceiveBatch processes a batch received from a peer host.
-	ReceiveBatch(batch common.EncodedBatch)
+	// ReceiveBatches receives a set of batches from a peer host.
+	ReceiveBatches(batches common.EncodedBatches)
```
joeldudley, line 27: This is a set of batches now, not just one, to accommodate the case where you're requesting a historical chain (and not just the latest).

go/host/p2p/p2p.go
```diff
@@ -162,9 +184,22 @@ func (p *p2pImpl) broadcast(msg Message) error {
 	return nil
 }
 
+// Sends a message to the sequencer.
+func (p *p2pImpl) sendToSequencer(msg message) error {
+	msgEncoded, err := rlp.EncodeToBytes(msg)
+	if err != nil {
+		return fmt.Errorf("could not encode message to send to sequencer. Cause: %w", err)
+	}
+	// TODO - #718 - Use better method to identify sequencer?
```
joeldudley, line 193: Thoughts?

BedrockSquirrel: I think it's fine for now, maybe set it up as like `p.currSequencer` as a slight abstraction to make it clearer what's going on. Longer term when we're decentralized I wonder if it would be better to use any peers to do the batches catch-up rather than all relying the current leader.

BedrockSquirrel (commented):
Seems good as a high-level approach. These batches operate like a block-chain though I guess? In that each batch knows the hash of its parent batch? You're ignoring batches that don't arrive in consecutive canonical order.

So I'd suggest mimicking the approach I'm taking with the blockprovider rather than 'looking for gaps': there are gaps if, and only if, you don't know the parent batch of the batch that just arrived. Then you ask for all blocks from the latest batch you have got

I think it'll be a bit more resilient and code might be a little simpler. You basically tell the sender, "this is where I'm at" and the sender decides where it should be sending from (makes it resilient to forks, and allows for batching of batches if you're very far behind).

Don't think I explained that v well, can chat about it if it helps.

joeldudley:
@BedrockSquirrel I think I'm doing what you suggest? You receive a batch. If you don't have its parent, you tell the sequencer which batch you need blocks from, and it sends them over. So if you receive batch 9 and you only have up to batch 4, you ask the sequencer to send over batches 4-9.

BedrockSquirrel:
Yeah maybe it is what you're doing, then I think you can probably simplify the code a bit. Like what is the `requestMissingBatches` method doing scanning backwards and stuff. You could just record the last stored batch as state, then if you get something you don't recognise then you request from that, there shouldn't be any calculation required. And the talk about 'looking for gaps' is a bit confusing, made me think you'd be scanning backwards and making a list of ones you hadn't seen or something.

go/host/host.go
```diff
@@ -1025,26 +1039,57 @@ func (h *host) checkBlockForSecretResponse(block *types.Block) bool {
...
 	}
 
-	// TODO - #718 - Have the enclave process batch, so that it's up to date.
-
-	// TODO - #718 - Implement a catch-up mechanism for historical batches.
+	// TODO - #718 - Have the enclave process the batch, so that it's up to date.
 
-	err = h.db.AddBatchHeader(&batch)
+	// We request any batches we've missed.
+	batchRequest, err := h.batchManager.CreateBatchRequest(batches)
```
BedrockSquirrel, line 1056: I find this whole flow very confusing. We've just received some batches I think, and then we call createBatchRequest with those batches. If you get rid of the sanity checks on incoming batches for now does it all still work? I think you're creating confusion by trying to fend off attacks that might not really exist.

go/common/host/host.go
```diff
@@ -23,8 +23,10 @@ type Host interface {
...
+	// SendBatches sends a set of batches to a peer host. Used during catch-up.
+	SendBatches(batchRequest common.EncodedBatchRequest)
```
BedrockSquirrel, line 29: This API feels a little odd, one of these methods is driven from internally, one from externally so they don't feel like they belong on the same interface. I guess we should have P2PReceiver and P2PRequester interfaces or something but that can be a future refactor.

BedrockSquirrel (commented):
LGTM high-level but it still feels a bit hard to follow, as if it expects multiple back-and-forths to deliver the batches. Requesting batches as a response to receiving batches feels odd, are we actually seeing gaps in the lists being sent over? Maybe better to just fix the cause of those gaps instead rather than worrying about defense at the moment.

go/host/host.go
joeldudley: I've renamed it for clarity - it's now `IsMissingBatches`, which returns a bool indicating if any historical batches are missing, plus the batch request needed to retrieve them.

All this would be cleaner if `BatchManager` had a reference to the P2P component, but it felt better overall to make all P2P calls go via the host. Thoughts?

joeldudley:
@BedrockSquirrel I think the flow can't really be simplified. It's not specifically about attacks. Rather, it's about what happens when the node is just starting up or has missed batches for a while (network partition, crash...). For example, in the sims as it stands (without these changes), the node will never receive the first x batches, because they happened before the node started.

For example, node A is currently on batch 25. It receives batch 30 from the sequencer. It needs the missing batches. There are three options, I think:

1. Store batch 30, and request batches 26-29
2. Create a session with the sequencer, requesting and receiving batches until it's up to date
3. Ignore batch 30, and request batches 26-30, and store the entire list when it receives it

(1) is messy, because you have a period where you have stored light batches that don't have ancestors. Not sure what bugs that would cause, but it scares me
(2) forces us to have a stateful messaging protocol, whereas all the P2P so far has been stateless, and it's so much simpler to keep it that way IMO
(3) is what I've gone for. It's stateless - you don't try and store some fragment of the chain, you just fire-and-forget a request to receive a "filled-in" chain

The stuff about checking for gaps or dupes could be dropped for now, but it's pretty well encapsulated now and would only need to be readded later.

BedrockSquirrel:
I don't agree that the dupe/gap thing would need to be checked for, it seems to me that the code is optimising for a situation like:

"send me 4, 5, 6 please"
"sure, here's 4 and 6"
"no, send me 4, 5, 6"

I would expect it to follow a similar pattern as the block ingestion on the enclave. That is:

1. have a method called something like `processBatch()` which takes a single batch and is used both for new batches and for batches that have come from p2p history requests

2. if processBatch can't process a batch because it's not got its parent or something it sends a request for historical batches that says like "I need batches 4 onwards"

3. Responder to the request says ok sure, sends back all the batches it has from 4 onwards, say 4-7

4. we call process batch on each of those elements in turn, if they were (bizarrely) in the wrong order or had a gap in the list then it would go to step 2. when it fails to process the next one.

I think this algorithm is stateless and resilient for a few scenarios such as initial bootstrap, falling behind, and forks. It prioritises a single code path for all batches to keep the flow simple.

joeldudley:
@BedrockSquirrel That's exactly the algorithm I've written I think?

go/host/batchmanager/batch_manager.go
```diff
@@ -0,0 +1,115 @@
...
+
+		batches = append(batches, batch)
+		currentBatch = big.NewInt(0).Add(currentBatch, big.NewInt(1))
+	}
+
+	return batches, nil
+}
+
+// Starting from the provided number, we walk the chain batch until we find a stored batch.
+func (b *BatchManager) findEarliestMissingBatch(startBatchNumber *big.Int) (*big.Int, error) {
```
BedrockSquirrel, line 92: Worth storing the headBatch, and updating it each time we store a newer batch so we can do the comparison without having to walk backwards?

BedrockSquirrel (approved):
LGTM, clearer now, thanks

PR #889, opened 2022-11-28
simplify blockprovider process control to use context

BedrockSquirrel:

- Blockprovider manages a goroutine that is responsible for streaming out the blocks to its client
- When client requests a new stream the old stream dies and the process is restarted with a clean stream channel
- There were bugs with the goroutine controlling mechanism in block provider
- And it was confusing to understand

- Replace the atomic status changes to use a context-based approach to stopping the streaming goroutine

### DIFF
--- a/go/ethadapter/blockprovider.go
+++ b/go/ethadapter/blockprovider.go
@@ -5,20 +5,12 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
-	"sync/atomic"
 	"time"
 
 	gethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	gethlog "github.com/ethereum/go-ethereum/log"
 	"github.com/obscuronet/go-obscuro/go/common/log"
-)
-
-type blockProviderStatus int32
-
-const (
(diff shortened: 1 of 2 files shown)


go/ethadapter/blockprovider.go
```diff
@@ -89,19 +84,24 @@ func (e *EthBlockProvider) IsLive(h gethcommon.Hash) bool {
...
-			time.Sleep(time.Second)
-			continue
+func (e *EthBlockProvider) streamBlocks(ctx context.Context, fromHeight *big.Int, streamCh chan *types.Block) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		default:
+			// this will block if we're up-to-date with live blocks
+			block, err := e.fetchNextCanonicalBlock(fromHeight)
```
BedrockSquirrel, line 94: This is the main loop for the block provider:
- get next block to send (this might be blocking, waiting for next block if client is up-to-date)
- send block to consumer (this might be blocking if consumer is slow to pull from the channel)

This PR adds the context check to break us out of this loop if cancelled.

tudor-malene (approved):
lgtm.
Definitely better now

PR #900, opened 2022-11-29
Log error if no peers list yet to send batch request to

BedrockSquirrel:

-If batches are getting processed before peers list has been updated (or peers list becomes empty for any reason) we don't want to see panics like this:
```
panic: runtime error: index out of range [0] with length 0

goroutine 313 [running]:
github.com/obscuronet/go-obscuro/go/host/p2p.(*p2pImpl).RequestBatches(0xc00048c780, 0xc0010a64a0?)
    /Users/joeldudley/Desktop/go-obscuro/go/host/p2p/p2p.go:116 +0xb0
```

The nature of the batch catch-up (batchup, if you will) is that it will recover later on if and when peers are available.

The reason this becomes an issue after my PR is that we jump straight into the main loop and we want to be resilient for the inputs for whatever state the host is in.

### DIFF
--- a/go/host/p2p/p2p.go
+++ b/go/host/p2p/p2p.go
@@ -1,6 +1,7 @@
 package p2p
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"net"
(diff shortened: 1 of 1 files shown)


joeldudley (approved):
Nice

PR #906, opened 2022-11-30
Fix blockprovider integration bugs (pre-genesis and returning to prev fork)

BedrockSquirrel:
The first commit of this PR was reviewed and merged yesterday. Then reverted because I noticed an increase in local test failures. Only the second commit is new here, it fixes the issues I was seeing.

the intermittent failures from that blockprovider PR yesterday turned out to be from a 'fun' bug when L1 returns to a fork it had previously abandoned.

When enclave finds a block it doesn't like it returns an error with its view of the current L1 head, then the host uses that information to start streaming from the correct place to continue.
```
   / a2 - a3 - a4
a1
   \ b2 
```
In this situation if the L1 was forking backwards and forward so that the enclave was fed: a1, a2, then b2, then back to a2, a3 it would get stuck.

Then this looping exchange takes place:

Enclave: a2 is a duplicate, I've seen it before. My current head is b2.
Host: ok your latest canonical ancestor is a1, I'll send you the block after that (a2) to get you back on track
Enclave: a2 is a duplicate, I've seen it before. My current head is b2.
{ and repeat forever}

My solution for now is that I will update the head pointer when we reject a duplicate. Will mention it in the logs so any side-effects of forks can be noticed as we keep improving this stuff.

- separate the L1 head storage from the L2 head storage:
  - set L1 head as soon as L1 block ingested
  - set L1 head when duplicate block submitted

### DIFF
--- a/go/enclave/db/interfaces.go
+++ b/go/enclave/db/interfaces.go
@@ -16,8 +16,10 @@ import (
 type BlockResolver interface {
 	// FetchBlock returns the L1 Block with the given hash.
 	FetchBlock(blockHash common.L1RootHash) (*types.Block, error)
-	// StoreBlock persists the L1 Block
-	StoreBlock(block *types.Block)
+	// StoreL1HeadBlock stores L1 blocks (which should only be stored when they are the head of the chain that is known to the enclave)
+	StoreL1HeadBlock(block *types.Block) error
+	// SetL1Head sets the L1 canonical head to the given block (used if block has already been ingested but head has rewound/moved)
+	SetL1Head(block gethcommon.Hash) error
 	// ParentBlock returns the L1 Block's parent.
 	ParentBlock(block *types.Block) (*types.Block, error)
 	// IsAncestor returns true if maybeAncestor is an ancestor of the L1 Block, and false otherwise
--- a/go/enclave/db/storage.go
+++ b/go/enclave/db/storage.go
@@ -116,9 +116,29 @@ func (s *storageImpl) FetchRollups(height uint64) ([]*core.Rollup, error) {
 	return obscurorawdb.ReadRollupsForHeight(s.db, height)
 }
 
-func (s *storageImpl) StoreBlock(b *types.Block) {
+// SetL1Head L1 blocks should only be stored when they are the head of the chain that is known to the enclave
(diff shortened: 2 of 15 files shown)


go/enclave/rollupchain/rollup_chain.go
```diff
@@ -306,10 +311,13 @@ func (rc *RollupChain) insertAndStoreL1Block(block types.Block, isLatest bool) e
...
 	}
-	rc.logger.Trace("block inserted successfully",
-		"height", block.NumberU64(), "hash", block.Hash(), "ingestionType", ingestionType)
 
-	rc.storage.StoreBlock(&block)
+	err = rc.storage.StoreL1HeadBlock(&block)
+	if err != nil {
+		return fmt.Errorf("failed to store L1 block, processing cannot continue. Cause: %w", err)
+	}
+	rc.logger.Error("§§§ block inserted successfully",
```
joeldudley, line 319: Is the `§§§` in there on purpose?

go/enclave/db/storage.go
```diff
@@ -116,9 +116,29 @@ func (s *storageImpl) FetchRollups(height uint64) ([]*core.Rollup, error) {
 	return obscurorawdb.ReadRollupsForHeight(s.db, height)
 }
 
-func (s *storageImpl) StoreBlock(b *types.Block) {
+// SetL1Head L1 blocks should only be stored when they are the head of the chain that is known to the enclave
```
joeldudley, line 119: I think this comment should only be on the interface, not the impl.

go/enclave/db/interfaces.go
```diff
@@ -16,8 +16,10 @@ import (
 type BlockResolver interface {
 	// FetchBlock returns the L1 Block with the given hash.
 	FetchBlock(blockHash common.L1RootHash) (*types.Block, error)
-	// StoreBlock persists the L1 Block
-	StoreBlock(block *types.Block)
+	// StoreL1HeadBlock stores L1 blocks (which should only be stored when they are the head of the chain that is known to the enclave)
```
joeldudley, line 19: Can you make it clear here that it stores the L1 block *and* updates the L1 head?

BedrockSquirrel:
This PR is replaced with a lighter touch fix: https://github.com/obscuronet/go-obscuro/pull/916

PR #907, opened 2022-11-30
Remove `HeadsAfterL1Block` type; some clean-up

joeldudley:

Currently, we have a `HeadsAfterL1Block` type, with three fields:

* the L1 head
* the L2 head
* whether a given block created a new L2 head

Having all three bundled together was causing issue in the move to batches, since there is no longer a 1-1 relationship between blocks and rollups/batches.

Unbundling the `HeadsAfterL1Block` type helps there. But it also made the code much clearer. Often, the L1 and L2 heads were being passed around together, when only one was needed.


go/common/enclave.go
```diff
@@ -110,7 +110,6 @@ type Enclave interface {
 
 // BlockSubmissionResponse is the response sent from the enclave back to the node after ingesting a block
 type BlockSubmissionResponse struct {
-	UpdatedHeadRollup       bool                      // Whether the block contained a new head rollup for the canonical rollup chain.
```
joeldudley, line 113: I got rid of this field. It made removing `HeadsAfterL1Blocks` much easier, and it's simple enough on the host side to check that `IngestedRollupHeader` isn't nil.

go/enclave/rollupchain/rollup_chain.go
```diff
@@ -106,17 +106,13 @@ func (rc *RollupChain) ProduceNewRollup(blockHash *common.L1RootHash) (*common.E
 	if err != nil {
 		return nil, fmt.Errorf("could not retrieve block to produce rollup. Cause: %w", err)
 	}
-	chainHeads, err := rc.storage.FetchHeadsAfterL1Block(*blockHash)
+	l2Head, err := rc.storage.FetchL2Head(*blockHash)
 	if err != nil {
-		if errors.Is(err, errutil.ErrNotFound) {
```
joeldudley, line 111: With the unbundling of L1 block ingestion and rollup production, this check didn't make sense.

go/enclave/rollupchain/rollup_chain.go
```diff
@@ -370,83 +366,64 @@ func (rc *RollupChain) insertBlockIntoL1Chain(block *types.Block, isLatest bool)
...
-			// We can return the genesis state immediately.
-			return headsAfterGenesisRollup, nil
+		if errors.Is(err, errIsPreGenesis) || errors.Is(err, errIsGenesisRollupInBlock) {
+			// Either we're still waiting for the genesis rollup, or it's already stored and we can return it immediately.
+			return l2Head, nil
 		}
 		return nil, fmt.Errorf("could not handle genesis rollup. Cause: %w", err)
 	}
 
-	// To calculate the state after the current block, we need the state after the parent.
```
joeldudley, line 426: This block was moved down into `calculateAndStoreNewHeads`.

BedrockSquirrel (approved):
LGTM - much clearer again, now we can see when we're just working with L2 state and its clearer that the L1 block -> L2 head is more of a lookup/pinning mechanism.

PR #913, opened 2022-11-30
Removes buggy `FetchHeadRollup` method on storage

joeldudley:

### DIFF
--- a/go/enclave/db/storage.go
+++ b/go/enclave/db/storage.go
@@ -61,18 +61,6 @@ func (s *storageImpl) FetchGenesisRollup() (*core.Rollup, error) {
 	return rollup, nil
 }
 
-func (s *storageImpl) FetchHeadRollup() (*core.Rollup, error) {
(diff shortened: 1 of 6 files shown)


go/enclave/db/storage.go
```diff
@@ -210,11 +198,18 @@ func (s *storageImpl) IsBlockAncestor(block *types.Block, maybeAncestor common.L
 }
 
 func (s *storageImpl) HealthCheck() (bool, error) {
-	headRollup, err := s.FetchHeadRollup()
+	_, l2Head, err := s.FetchCurrentHeads()
```
BedrockSquirrel, line 201: Is it worth keeping the FetchHeadRollup method and using this code as its body? Looks like it's used twice and I'm sure it'll be used more in future

BedrockSquirrel (commented):
LGTM but I'd be tempted to just leave the method but with the new implementation instead of storage

PR #916, opened 2022-12-01
Fix fork block loop and re-merge blockprovider PR

BedrockSquirrel:
Note: again, the first commit here is the commit from the other day which integrates block provider into the host for all block feeding (@tudor-malene reviewed). Only the small second commit here is a completely new change.

- Fix the block error loops in a lighter touch way than that PR from yesterday
- We're still figuring out the nuance of enclave state re: blocks/rollups/batches, I don't want to muddy the water there
- This change means it won't get stuck in a loop, the stream is allowed to continue if the enclave complains about a duplicate block with the view that it's probably just one or two duplicate blocks to get it back on track, respecting the L1Head info it provided already

- move standard submit block errors to common as they are part of the interface
- check for duplicate error when deciding whether to reset the block provider stream

### DIFF
--- a/go/common/enclave.go
+++ b/go/common/enclave.go
@@ -1,6 +1,7 @@
 package common
 
 import (
+	"errors"
 	"fmt"
 
 	gethcommon "github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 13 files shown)


tudor-malene (approved):
lgtm

PR #925, opened 2022-12-01
Have shorter timeout for awaiting receipt in in-mem sim

joeldudley:

Quite often, a change will break a node receiving transactions properly. In the in-mem sim, we currently have to wait 30 secs to find out this is the case, extending the feedback loop. I've changed it to be only 5 secs for the in-mem.

### DIFF
--- a/integration/simulation/params/params.go
+++ b/integration/simulation/params/params.go
@@ -39,6 +39,8 @@ type SimParams struct {
 
 	StartPort int  // The port from which to start allocating ports. Must be unique across all simulations.
 	IsInMem   bool // Denotes that the sim does not have a full RPC layer.
+
+	Timeout int // How long to wait for transactions to be confirmed.
 }
 
 type L1SetupData struct {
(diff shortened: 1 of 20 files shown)


integration/simulation/params/params.go
```diff
@@ -39,6 +39,8 @@ type SimParams struct {
 
 	StartPort int  // The port from which to start allocating ports. Must be unique across all simulations.
 	IsInMem   bool // Denotes that the sim does not have a full RPC layer.
+
+	Timeout int // How long to wait for transactions to be confirmed.
```
BedrockSquirrel, line 43: Don't want to be pernickety but could we call it `ReceiptTimeout` or `TxTimeout` because it's a bit vague at simulation config level. Also probably should put `Millis` on the name or use a `Duration`.

BedrockSquirrel (commented):
LGTM, just some fussing about the variable name.

Also, if you fancied upgrading AwaitReceipt to use the `retry.Do` mechanism while you're there it'd be much appreciated, it's been on my hitlist for a while, but that might not be a drop-in replacement so no worries if it's a faff.

integration/simulation/params/params.go
joeldudley: Fixed

BedrockSquirrel (approved):
LGTM, thanks for `retry.Do`ifying. Also, gofumpt yourself.

PR #927, opened 2022-12-02
Fixes batch catchup, and fixes bug in sequencer rollup production

joeldudley:

The existing batch catchup mechanism wasn't able to handle forks (e.g. if we reached rollup 50, then rolled back to rollup 45, it was too dumb to know that the last five rollups were invalid, and it should catch them up).

Fixing this revealed bugs in how the sequencer produced rollups, with occasional L2 chain forks. These are now fixed.


go/common/batches.go
```diff
@@ -10,12 +12,22 @@ type ExtBatch struct {
...
+	r.hash.Store(v)
+	return v
+}
 
 // BatchRequest is used when requesting a range of batches from a peer.
 type BatchRequest struct {
-	Requester            string
-	EarliestMissingBatch *big.Int
+	Requester        string
+	CurrentHeadBatch *gethcommon.Hash // The requester's view of the current head batch, or nil if they haven't stored any batches.
```
joeldudley, line 32: We moved from a model where the requester says which batches they're missing, to one where the sequencer works it out. The former had a fatal flaw, in that it failed when there had been a fork and it had rollups at the correct height, but on the wrong fork. Thanks Matt :)

go/common/enclave.go
```diff
@@ -45,7 +45,7 @@ type Enclave interface {
 	// It is the responsibility of the host to gossip the returned rollup
 	// For good functioning the caller should always submit blocks ordered by height
 	// submitting a block before receiving ancestors of it, will result in it being ignored
-	SubmitL1Block(block types.Block, isLatest bool) (*BlockSubmissionResponse, error)
+	SubmitL1Block(block types.Block, isLatest bool, isSequencer bool) (*BlockSubmissionResponse, error)
```
joeldudley, line 48: We're coupling rollup production to L1 block submission, to avoid a race condition where two rollups are produced at the same height if two L1 blocks come in at the same time. But we have to add this parameter, as only the sequencer should be producing blocks.

go/common/enclave.go
```diff
@@ -114,10 +114,11 @@ type Enclave interface {
 
 // BlockSubmissionResponse is the response sent from the enclave back to the node after ingesting a block
 type BlockSubmissionResponse struct {
-	IngestedRollupHeader    *Header                   // The header of the winning rollup contained in the ingested block, if any.
-	ProducedSecretResponses []*ProducedSecretResponse // if L1 block contained secret requests then there may be responses to publish
-	SubscribedLogs          map[rpc.ID][]byte         // The logs produced by the block and all its ancestors for each subscription ID.
-	RejectError             *BlockRejectError         // this is set if block was rejected, contains information about what block to submit next
+	ProducedRollup          *ExtRollup                // If the node is the sequencer, the rollup produced if the ingested L1 block contained a new rollup.
```
joeldudley, line 117: We send back a produced rollup now, instead of generating it using a separate call.

go/enclave/core/rollup.go
```diff
@@ -68,7 +68,7 @@ func EmptyRollup(agg gethcommon.Address, parent *common.Header, blkHash gethcomm
 		Agg:        agg,
 		ParentHash: parent.Hash(),
 		L1Proof:    blkHash,
-		Number:     big.NewInt(int64(parent.Number.Uint64() + 1)),
+		Number:     big.NewInt(0).Add(parent.Number, big.NewInt(1)),
```
joeldudley, line 71: Safer.

go/enclave/rollupchain/pobi.go
```diff
@@ -9,28 +9,28 @@ import (
 
 // Returns the new head rollup, and a boolean indicating whether this is a new rollup or the existing head rollup.
 // todo - add statistics to determine why there are conflicts.
-func selectNextRollup(parentRollup *core.Rollup, rollups []*core.Rollup, blockResolver db.BlockResolver) (*core.Rollup, bool) {
```
joeldudley, line 12: No actual changes here, just some cleanup.

go/enclave/rollupchain/rollup_chain.go
```diff
@@ -98,33 +98,6 @@ func New(
 	}
 }
 
-// ProduceNewRollup creates a new rollup, building on the latest chain heads.
-func (rc *RollupChain) ProduceNewRollup() (*common.ExtRollup, error) {
```
joeldudley, line 102: This function was moved down, since it is now internal (called as part of SubmitL1Block).

go/enclave/rollupchain/rollup_chain.go
```diff
@@ -157,21 +130,32 @@ func (rc *RollupChain) ProduceGenesisRollup(blkHash common.L1RootHash) (*core.Ro
...
+	// We update the L1 and L2 chain heads.
+	l2Head, isUpdatedRollupHead, err := rc.updateHeads(&block)
 	if err != nil {
 		return nil, rc.rejectBlockErr(err)
 	}
 
-	return rc.produceBlockSubmissionResponse(&block, l2Head)
+	// If we're the sequencer and we've ingested a rollup, we produce a new one.
+	var rollup *common.ExtRollup
+	if isSequencer && isUpdatedRollupHead {
```
joeldudley, line 151: This was crucial - we should only produce a rollup once a new one has been ingested.

Obviously, this will have to change in the future, once rollups in L1s are just used to validate the batches we have created. But we're not there yet.

go/enclave/rollupchain/rollup_chain.go
```diff
@@ -359,51 +343,81 @@ func (rc *RollupChain) insertBlockIntoL1Chain(block *types.Block, isLatest bool)
...
+	if isUpdatedRollupHead {
+		headRollup, err := rc.storage.FetchRollup(*l2Head)
+		if err != nil {
+			rc.logger.Crit("Could not fetch rollup", log.ErrKey, err)
+		}
 		ingestedRollupHeader = headRollup.Header
 	}
 
 	return &common.BlockSubmissionResponse{
+		ProducedRollup:       producedRollup,
```
joeldudley, line 387: We reintroduce returning the rollup with the block.

go/host/db/batches.go
```diff
@@ -35,13 +35,18 @@ func (db *DB) GetBatchHeader(hash gethcommon.Hash) (*common.Header, error) {
 func (db *DB) AddBatchHeader(batch *common.ExtBatch) error {
 	b := db.kvStore.NewBatch()
 
+	// Skip storing the batch if we've already stored it.
+	if _, err := db.GetBatchHeader(batch.Hash()); err == nil {
```
joeldudley, line 39: An optimisation.

go/host/host.go
```diff
@@ -504,29 +504,23 @@ func (h *host) processL1Block(block *types.Block, isLatestBlock bool) error {
...
 		return nil // nothing further to process since network had no genesis
 	}
 
-	rollup, err := h.enclaveClient.ProduceRollup()
-	if err != nil {
-		return fmt.Errorf("could not produce rollup. Cause: %w", err)
-	}
-
-	if rollup.Header != nil {
+	if result.ProducedRollup != nil && result.ProducedRollup.Header != nil {
```
joeldudley, line 519: The first check works for the in-mem sims, and the second for the full-network sims. Insane.

integration/simulation/network/network_utils.go
```diff
@@ -79,7 +80,7 @@ func createInMemObscuroNode(
 		NodeType:            nodeType,
 		GossipRoundDuration: avgGossipPeriod,
 		HasClientRPCHTTP:    false,
-		P2PPublicAddress:    "dummy_address", // Required because the node sanity-checks that this field is not empty at start-up.
+		P2PPublicAddress:    fmt.Sprintf("%d", id),
```
joeldudley, line 83: If we don't have this, it borks batch catch-up.

integration/simulation/network/obscuro_node_utils.go
```diff
@@ -215,8 +215,8 @@ func StopObscuroNodes(clients []rpc.Client) {
 		}(client)
 	}
 
-	if waitTimeout(&wg, 10*time.Second) {
-		panic("Timed out waiting for the Obscuro nodes to stop")
+	if waitTimeout(&wg, 5*time.Second) {
+		testlog.Logger().Warn("Timed out waiting for the Obscuro nodes to stop")
```
joeldudley, line 219: Wasn't worth failing the sims over IMO.

BedrockSquirrel (commented):
Seems generally reasonable, just a few small comments.

Also, just for my clarity, do we still send a batch after every L1 block? And do the transactions from (potentially) multiple batches all get included in their next rollup?

joeldudley:
> Also, just for my clarity, do we still send a batch after every L1 block? And do the transactions from (potentially) multiple batches all get included in their next rollup?

Yes, there's a batch after every block. For now, it's just one batch per block. In the future, yes, rollups will roll up multiple batches.

go/enclave/enclave.go
```diff
@@ -268,7 +268,7 @@ func (e *enclaveImpl) SubmitL1Block(block types.Block, isLatest bool) (*common.B
 }
 
 func (e *enclaveImpl) ProduceRollup() (*common.ExtRollup, error) {
-	return e.chain.ProduceNewRollup()
+	return nil, errutil.ErrNoImpl
```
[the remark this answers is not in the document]
joeldudley: I'd rather not delete. It's quite heavy to delete (lots of protobuf stuff), and I know that at some point I'll need to create batches outside of the cadence of L1 blocks.

joeldudley:
@BedrockSquirrel Reverted the node shutdown panic, and added the enclave config key to tell it whether it's the sequencer's enclave.

BedrockSquirrel (approved):
LGTM

PR #933, opened 2022-12-05
Host ID is now generated from PK

otherview:

- Part of https://github.com/obscuronet/obscuro-internal/issues/338
- Host Config split between Host Parsed Config ( the configs supplied externally) and Host Config ( the configs that are used in the Host)
- Host ID is now derived from the Private Key (as the Public Key Address)
- Enclave still takes the Host ID from console
- Added a TODO for the Host to check if the Enclave was started with the correct Host ID
- Configs as a pointer instead of value


go/config/host_config.go
```diff
@@ -15,10 +15,8 @@ const (
 	defaultP2PTimeoutSecs   = 10
 )
 
-// HostConfig contains the full configuration for an Obscuro host.
-type HostConfig struct {
-	// The host's identity
-	ID gethcommon.Address
+// HostParsedConfig contains the configuration that was parsed from a config file / command line to start the Obscuro host.
+type HostParsedConfig struct {
```
BedrockSquirrel, line 19: I don't really understand the need for this struct and I'm worried it's getting confusing because with that Toml one we've not got 3 different structs representing Host config. Could the ID just be calculated in the ParseConfig() method and kept as another field on the existing HostConfig or is there an issue with that?

BedrockSquirrel (commented):
Liking the ID setup but I'm a bit confused by the new Config type

go/config/host_config.go
otherview: It's a good q.

That was my first approach. 

Having the ID as a field where it is, in an 1-to-1 object that's used to parse the incoming flags AND as the execution config is misleading.
You'd have to implicitly know where the ID gets populated and that it won't directly come from the flags.
I didn't what to burden the ParseFlags because it's a host-specific operation and that would impact all tests that use the host runner.

The ID field shouldn't be written into as it must derive from the Private Key. I think it's important to make this distinction in the code, that's why I created a superset.

So now there's a HostParsedConfig which is tested with the toml tests and handles the flags inputs. And there's the HostConfig which is a sanitized/derived object that's used in the host execution.

What if I copy (plus modify) the HostParsedConfig into a HostConfig that only has flags that will be used in the execution ?
So instead of HostConfig super-seeding HostParsedConfig, it would be a new object with the required flags only ?

PR #947, opened 2022-12-07
Submit each batch to the enclave if it's successfully stored

joeldudley:

### DIFF
--- a/go/host/batchmanager/batch_manager.go
+++ b/go/host/batchmanager/batch_manager.go
@@ -25,29 +25,26 @@ func NewBatchManager(db *db.DB) *BatchManager {
 	}
 }
 
-// StoreBatches stores the provided batches. If there are missing batches in the chain, it returns a
-// `ErrBatchesMissing`.
-func (b *BatchManager) StoreBatches(batches []*common.ExtBatch) error {
-	for _, batch := range batches {
-		// If we have stored the batch's parent, or this batch is the genesis batch, we store the batch.
-		_, err := b.db.GetBatch(batch.Header.ParentHash)
-		if err == nil || batch.Header.Number.Uint64() == common.L2GenesisHeight {
-			err = b.db.AddBatchHeader(batch)
-			if err != nil {
-				return fmt.Errorf("could not store batch header. Cause: %w", err)
-			}
-			continue
-		}
+// StoreBatch stores the provided batch. If we cannot find the batch's parent, we return an `ErrBatchesMissing`.
+func (b *BatchManager) StoreBatch(batch *common.ExtBatch) error {
+	isGenesisBatch := batch.Header.Number.Uint64() == common.L2GenesisHeight
 
-		// If we could not find the parent, we return an `ErrBatchesMissing`.
-		if errors.Is(err, errutil.ErrNotFound) {
-			return ErrBatchesMissing
+	// If we have stored the batch's parent, or this batch is the genesis batch, we store the batch.
+	_, err := b.db.GetBatch(batch.Header.ParentHash)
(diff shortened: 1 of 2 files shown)


go/host/host.go
```diff
@@ -847,29 +847,27 @@ func (h *host) handleBatches(encodedBatches *common.EncodedBatches) error {
...
+			if err = h.p2p.RequestBatches(batchRequest); err != nil {
+				return fmt.Errorf("could not request historical batches. Cause: %w", err)
+			}
+			return nil
+		}
 
-	// TODO - #718 - We should probably submit each batch after storing it, and not submitting each one only if *every*
-	//  batch was stored correctly.
-	for _, batch := range batches {
+		// TODO - #718 - What happens if this op fails? We'll think we have the batch, so we'll never catch it up and
```
BedrockSquirrel, line 869: So shouldn't we store the batch only *after* successful submission to Enclave?

go/host/batchmanager/batch_manager.go
```diff
@@ -25,29 +25,26 @@ func NewBatchManager(db *db.DB) *BatchManager {
...
-		// If we could not find the parent, we return an `ErrBatchesMissing`.
-		if errors.Is(err, errutil.ErrNotFound) {
-			return ErrBatchesMissing
+// StoreBatch stores the provided batch. If we cannot find the batch's parent, we return an `ErrBatchesMissing`.
+func (b *BatchManager) StoreBatch(batch *common.ExtBatch) error {
+	isGenesisBatch := batch.Header.Number.Uint64() == common.L2GenesisHeight
+
+	// If we have stored the batch's parent, or this batch is the genesis batch, we store the batch.
+	_, err := b.db.GetBatch(batch.Header.ParentHash)
+	if isGenesisBatch || err == nil {
```
BedrockSquirrel, line 34: Subject but maybe we could flip this round so happy path is main path in typical go fashion?

```
if err:
    if err != notFound: return err
    if err == Not found && not genesis: return errNotFound
AddBatchHeader(batch)
return nil    
```

go/host/host.go
joeldudley: Yeah, will reorder.

go/host/batchmanager/batch_manager.go
joeldudley: It actually ends up being uglier, due to the special case of it possibly being the genesis - you have to use a triple nested block or repeat some conditions.

go/host/host.go
```diff
@@ -847,31 +847,27 @@ func (h *host) handleBatches(encodedBatches *common.EncodedBatches) error {
...
-			return fmt.Errorf("could not create batch request. Cause: %w", err)
-		}
-		if err = h.p2p.RequestBatches(batchRequest); err != nil {
-			return fmt.Errorf("could not request historical batches. Cause: %w", err)
-		}
-
-		// If we requested any batches, we return early and wait for the missing batches to arrive.
-		return nil
-	}
+			if !errors.Is(err, batchmanager.ErrBatchesMissing) {
```
BedrockSquirrel, line 858: Is there a risk of issues if the host storage gets out of sync with the enclave storage? 

Do the batches need to be stored on the host at all? Maybe just as a caching layer that falls back to the enclave.

Can maybe just have a `todo` for that for now but I think if the issue ever does arise it'll be a thorny one (e.g. the host requesting a batch, failing the submission to the enclave because the enclave already has it but the original persistence failed, and then requesting it again in an infinite loop)

BedrockSquirrel (approved):
LGTM - concern about the batch storage getting out of sync between host and enclave

PR #948, opened 2022-12-07
Adding stop and status scripts

otherview:

- https://github.com/obscuronet/obscuro-internal/issues/1270

### DIFF
--- a/testnet/obscuro-node-status.sh
+++ b/testnet/obscuro-node-status.sh
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+
+#
+# This script displays the Obscuro node health status
+#
+
+help_and_exit() {
+    echo ""
+    echo "Usage: "
+    echo "      -  $(basename "${0}") "
+    echo ""
+    exit 1  # Exit with error explicitly
+}
+# Ensure any fail is loud and explicit
+set -euo pipefail
+
+# Define local usage vars
+start_path="$(cd "$(dirname "${0}")" && pwd)"
+testnet_path="${start_path}"
+
+# Fetch options
+for argument in "$@"
+do
+    key=$(echo $argument | cut -f1 -d=)
+    value=$(echo $argument | cut -f2 -d=)
+
+    case "$key" in
+
+            --help)                     help_and_exit ;;
+            *)
+    esac
+done
+
+
+net_status=$(curl -s --request POST 'http://127.0.0.1:13000' \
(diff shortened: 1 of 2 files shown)


testnet/obscuro-node-status.sh
```diff
@@ -0,0 +1,42 @@
...
+
+    case "$key" in
+
+            --help)                     help_and_exit ;;
+            *)
+    esac
+done
+
+
+net_status=$(curl -s --request POST 'http://127.0.0.1:13000' \
```
moraygrieve, line 35: Does this assume you always run the node on port 13000? I thought in the start script the port could be set differently

testnet/stop-obscuro-node.sh
```diff
@@ -0,0 +1,35 @@
...
+
+    case "$key" in
+
+            --help)                     help_and_exit ;;
+            *)
+    esac
+done
+
+
+docker-compose stop
```
moraygrieve, line 35: Does this stop all running containers gracefully? i.e. what would happen if the user was running other containers on the same host?

testnet/obscuro-node-status.sh
otherview: Good point, I'll hook these up to the flags 👌

otherview: Actually, just checked and we hardcode those ports in the startup script. I guess there's no point in changing them until it's strictly necessary ? WDYT ?

moraygrieve: When you start the node you give the p2p_address i.e. "51.189.31.34:1000" ... is 1000 not the port the node listens on?

testnet/stop-obscuro-node.sh
moraygrieve: So you have to be in the testnet folder, and then it will automatically resolve to using the docker-compose.yml file?

moraygrieve: Running locally I don't think that then removes the old containers does it, so restarting would require you to do a docker container prune?

otherview: Yes, that's my understanding.

testnet/obscuro-node-status.sh
moraygrieve: aha got it

moraygrieve (approved):
LGTM

PR #957, opened 2022-12-11
Separate DI constructor for container in testing

BedrockSquirrel:

- For test use-cases we need to be able to start the containers with our own logger and potentially mocked out dependencies for the host
- Need this for the testing I'm doing on enclave/host restarts
- Note: this isn't super elegant but they're only used by tests

- Rename `NewXXXContainer` to `NewXXXContainerFromConfig` for the methods used by `main`s to initialise the servers
- Create a separate constructor that allows dependencies to be passed in (e.g. pass in the logger instead of constructing it based on config

### DIFF
--- a/go/enclave/container/enclave_container.go
+++ b/go/enclave/container/enclave_container.go
@@ -23,28 +23,6 @@ type EnclaveContainer struct {
 	Logger    gethlog.Logger
 }
 
-// NewEnclaveContainer wires up the components of the Enclave and its RPC server. Manages their lifecycle/monitors their status
-func NewEnclaveContainer(config config.EnclaveConfig) *EnclaveContainer {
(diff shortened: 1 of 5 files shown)


go/enclave/container/enclave_container.go
```diff
@@ -23,11 +23,35 @@ type EnclaveContainer struct {
 	Logger    gethlog.Logger
 }
 
-// NewEnclaveContainer wires up the components of the Enclave and its RPC server. Manages their lifecycle/monitors their status
-func NewEnclaveContainer(config config.EnclaveConfig) *EnclaveContainer {
+func (e *EnclaveContainer) Start() error {
```
BedrockSquirrel, line 26: This was just moved up from below, I put the container methods with the struct and constructors at the bottom.

joeldudley (approved):
LGTM

PR #966, opened 2022-12-13
Bridge - WIP

StefanIliev545:

This PR brings in the new bridge smart contracts along with hardhat tests and deployments.

Added new bridge smart contracts, test for them and in progress deployment revamp to use hardhat.

Important area would be the whole of the contracts/ folder, but specifically "deploy_l1" and "deploy_l2". These are used by a hardhat plugin (hardhat deploy) which is the official-unofficial way to manage deployments across different networks with support for L1 & L2 (their docs mention optimism uses it).

On the solidity side, without modifying the current definition of the message bus a couple of contracts are added - 
CrossChainEnabledObscuro - Inheriting this allows you to receive and queue cross chain messages. It is a presetup for becoming OZ cross chain compliant (right now it works pretty much the same way), but needs us to bump up the version which is causing issues with older smart contracts that need to be fixed separately from this PR.
CrossChainMessenger - verifies existance of messages and allows relaying messages whilst maintaining the cross chain context. 

L1 and L2 bridges - need more work in regards to ETH to wETH and OBX_native to OBX_L1 mappings; Whitelist is done using roles where the plan is that we'd add more roles for different token types that require different processing logic. The whitelist effectively signifies what processing logic handles the asset type along with it being supported. When whitelisting an asset a cross chain message is sent to the subordinate L2 bridge to create a token wrapper. 

The big generated files can be ignored, might remove them from git. 

The deployment scripts need to be wired up across the number of environments (docker, sim tests, etc) but the plan is that everything will use the same deployment scripts and the only thing different would be the network configs.   


tudor-malene (commented):
I'll wait until the hardhat pr is merged into master and this is rebased on that.
it's too much noise atm.

PR #972, opened 2022-12-13
Proper handling of genesis batch on the host side

joeldudley:

Previously, the genesis batch was submitted to the enclave by the validators, which didn't make sense (since the genesis is retrieved from the L1 blocks). But the genesis batch still needs to be distributed, since some API calls rely on retrieving all batches, including the genesis batch, from the host.


BedrockSquirrel (approved):
lgtm

PR #988, opened 2022-12-15
Add fast-fail mechanism to break out of retry

BedrockSquirrel:

- A couple of situations have come up where it would be useful to be able to exit from a retry mechanism
- One is implemented in this PR, when awaiting receipt we want to retry for "not found" but bail out for other errors (to avoid sitting there for a minute if we can't connect to the RPC node)

- add `retry.FailFast` function that wraps an error to break out of the retry loop
- change AwaitReceipt to use it

### Definition of done

- [x] Unit tests added to cover new or changed functionality 
- [x] Docs pages updated to cover new or changed functionality
- [x] [Changelog.md](https://github.com/obscuronet/go-obscuro/blob/main/docs/testnet/changelog.md) updated 
- [x] End-to-end tests run if required (see below)

### End-to-end tests

- Run the workflow

### DIFF
--- a/go/common/retry/retry.go
+++ b/go/common/retry/retry.go
@@ -1,6 +1,7 @@
 package retry
 
 import (
+	"errors"
 	"fmt"
 	"time"
 )
@@ -17,6 +18,12 @@ func Do(fn func() error, retryStrat Strategy) error {
 			return nil
 		}
 
+		var ffErr *failFastError
+		if errors.As(err, &ffErr) {
+			// if error has been wrapped as fail fast then we don't continue to retry
+			return ffErr.wrapped
+		}
+
 		// calling NextRetryInterval() marks the end of an attempt for the retry strategy, so we call it before checking Done()
 		nextInterval := retryStrat.NextRetryInterval()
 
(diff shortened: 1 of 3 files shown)


otherview (approved):
Instead of adding complexity to the retry method, I would create a new method.

Maybe `DoWithFastFail(fn func() error, fastFailError error, retryStrat Strategy)`.

The point being to decouple the func from the fail clause, the complexity of the fast failing is the responsibility of the retry mechanism and not of the `fn func() error`. The `fn` is simpler and only returns an error.

Everything looks good, and this particular would be the pattern I'd make but I don't think it's a blocker for the PR.

BedrockSquirrel:
@otherview I don't understand your suggestion really, how would that fastFailError be used as a parameter?

I think this approach is fairly clean tbh but maybe a version where the function has access to the retry strategy and can tell it to stop or something would work. I like that the failFastError is still just an error, if this function is used anywhere else that won't affect behaviour*.

*I just realised I didn't implement wraps so it breaks .Is()/.As() functionality, I'm adding that now.

PR #989, opened 2022-12-15
Reduce sleep time after unexpected block provider error

BedrockSquirrel:

When blockprovider hits an unexpected error it had a hardcoded 1 second sleep. For some reason it's hitting a bunch of these errors in the in-mem (which has a mock L1).

But 10 blocks go by in that 1 second in the in-mem sim, this is what's causing the inefficiency failures I think. This PR which reduces the hardcoded sleep seems to improve stability (we can look at the cause of the in-mem failures separately).

The unexpected error is rarely hit in normal usage so I don't think having a shorter sleep will have an impact. We should consider adding a back-off perhaps.

- Reduce hard-coded sleep after failure for in-mem sim stability
- Increase sleep at startup to allow mock L1 to produce first block before host starts (one source of unexpected errors)

### Definition of done

### End-to-end tests

- Run the workflow

### DIFF
--- a/integration/simulation/network/inmemory.go
+++ b/integration/simulation/network/inmemory.go
@@ -94,7 +94,7 @@ func (n *basicNetworkOfInMemoryNodes) Create(params *params.SimParams, stats *st
 	for _, m := range n.ethNodes {
 		t := m
 		go t.Start()
-		time.Sleep(params.AvgBlockDuration / 20)
+		time.Sleep(params.AvgBlockDuration)
 	}
 
 	for _, m := range obscuroNodes {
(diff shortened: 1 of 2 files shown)


integration/simulation/network/inmemory.go
```diff
@@ -94,7 +94,7 @@ func (n *basicNetworkOfInMemoryNodes) Create(params *params.SimParams, stats *st
 	for _, m := range n.ethNodes {
 		t := m
 		go t.Start()
-		time.Sleep(params.AvgBlockDuration / 20)
+		time.Sleep(params.AvgBlockDuration)
```
BedrockSquirrel, line 97: Before this change, every node would log this error: `err="could not fetch block at requested height=1 - not found"`

otherview (approved):

PR #997, opened 2022-12-15
Switching genesis to be a batch, not a rollup

joeldudley:

Previously, the workflow was:

* Sequencer produces genesis rollup
* Sequencer publishes genesis rollup to L1
* Sequencer and validators read genesis rollup from block

And then onto batches.

This PR changes it so that the sequencer produces the genesis rollup when it notices that one doesn't exist yet, and distributes it to other nodes via the standard mechanisms. This is much simpler, because it means there is not some special path to handle the genesis.

### Definition of done

### End-to-end tests

- Run the workflow


go/common/enclave.go
```diff
@@ -33,9 +33,6 @@ type Enclave interface {
 	// InitEnclave - initialise an enclave with a seed received by another enclave
 	InitEnclave(secret EncryptedSharedEnclaveSecret) error
 
-	// ProduceGenesis - produce the genesis batch
-	ProduceGenesis(blkHash gethcommon.Hash) (*ExtBatch, error)
```
joeldudley, line 37: We now produce the genesis as part of processing a block as the sequencer, rather than producing it ahead of time, submitting it to the L1, and reading it back as a rollup.

go/host/host.go
```diff
@@ -514,14 +508,6 @@ func (h *host) processL1Block(block *types.Block, isLatestBlock bool) error {
 		return nil
 	}
 
-	if h.genesisRequired {
```
joeldudley, line 517: The sequencer now produces the genesis automatically upon receiving an L1 block and noticing no head batch exists.

go/enclave/rollupchain/rollup_chain.go
```diff
@@ -102,48 +92,6 @@ func New(
 	}
 }
 
-// ProduceGenesisBatch creates a genesis batch linked to the provided L1 block and signs it.
-func (rc *RollupChain) ProduceGenesisBatch(blkHash common.L1RootHash) (*core.Batch, error) {
```
joeldudley, line 106: This was moved down, as it's no longer a public method.

go/enclave/rollupchain/rollup_chain.go
```diff
@@ -168,22 +116,26 @@ func (rc *RollupChain) UpdateL2Chain(batch *core.Batch) (*common.BatchHeader, er
 	rc.blockProcessingMutex.Lock()
 	defer rc.blockProcessingMutex.Unlock()
 
-	// We retrieve the genesis batch from the L1 chain instead.
-	if batch.Header.Number.Cmp(big.NewInt(int64(common.L2GenesisHeight))) == 0 {
```
joeldudley, line 172: We no longer expect to retrieve the genesis from a rollup as a validator. We expect to receive it as a batch instead.

go/enclave/rollupchain/rollup_chain.go
```diff
@@ -387,78 +339,119 @@ func (rc *RollupChain) insertBlockIntoL1Chain(block *types.Block, isLatest bool)
...
-			return -1, fmt.Errorf("could not retrieve genesis rollup. Cause: %w", err)
+			return nil, nil, fmt.Errorf("could not retrieve current head batch. Cause: %w", err)
+		}
+		isGenesis = true
+	}
+
+	var l2HeadTxReceipts types.Receipts
+	// If we're the sequencer, we produce a new L2 head to replace the old one.
+	if rc.nodeType == common.Sequencer {
+		l2Head, l2HeadTxReceipts, err = rc.produceAndStoreBatch(block, isGenesis)
```
BedrockSquirrel, line 360: I'm a bit worried that `isLatest` doesn't seem to be a factor in whether we produce a batch, is that correct?

As far as I can tell, it won't error during the `updateL1State` and so it'll fall all the way through to here, and produce one. So if we're behind the L1 we could be potentially be spamming out batches with transactions based on old state.

I'm a bit fuzzy on how this should work tbh but there's got to be some sort of alignment going on.

BedrockSquirrel (commented):
In general looks great, good job clearing away some of that crap. I am concerned that we're not paying attention to `isLatest` though so would like some clarity on that.

I think it wasn't an issue before because the batches wouldn't get produced until that first rollup was published to the L1 but with this change it probably gets a bit weird.

go/enclave/rollupchain/rollup_chain.go
```diff
@@ -387,78 +339,119 @@ func (rc *RollupChain) insertBlockIntoL1Chain(block *types.Block, isLatest bool)
...
-	case PostGenesis:
-		return rc.handlePostGenesisBlock(block, rollupsInBlock)
-	}
-	return nil, nil, fmt.Errorf("unrecognised block stage")
-}
-
-// Determines if this is a pre-genesis L2 block, the genesis L2 block, or a post-genesis L2 block.
-func (rc *RollupChain) getBlockStage(rollupsInBlock []*core.Rollup) (BlockStage, error) {
-	_, err := rc.storage.FetchBatchByHeight(0)
+	// We determine whether this is the genesis block.
```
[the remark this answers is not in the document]
BedrockSquirrel: What I said might not even be right actually. But yeah it could be a bit clearer, genesis block makes it sound like it's related to the L1 genesis.

go/enclave/rollupchain/rollup_chain.go
```diff
@@ -386,79 +338,117 @@ func (rc *RollupChain) insertBlockIntoL1Chain(block *types.Block, isLatest bool)
...
 		}
+		isGenesis = true
+	}
 
-		// If we haven't stored the genesis rollup and there are no rollups in this block, it cannot be the L2
-		// genesis block.
-		if len(rollupsInBlock) == 0 {
-			return PreGenesis, nil
+	var l2HeadTxReceipts types.Receipts
+	// If this isn't the genesis block, we retrieve the stored receipts.
```
BedrockSquirrel, line 358: This comment still pretty confusing. I don't know how to word it, but think it's more like "if L2 chain already exists" or "if L2 genesis batch exists" to avoid confusion (genesis block sounds like L1 genesis)

BedrockSquirrel (approved):
LGTM

PR #1001, opened 2022-12-16
Add Is() support to blockRejectErr for standard errors

BedrockSquirrel:

- There is a bug in the L1 catch-up mechanism, we get stuck in a loop where the enclave is complaining of a block it's already seen and the host sends it the same one again
- This piece of code in `host.go` was supposed to prevent that bug by continuing to stream when it complains of duplicate block:
Screenshot 2022-12-16 at 18 00 23

This piece of code didn't work because the `errors.Is()` was misreporting the error as *not* being an ErrBlockAlreadyProcessed. Because it lost its type when it was sent over the wire on RPC response

So we extend the blockRejErr.As() implementation to check for matching strings (which works for our standard errors).

- Have err.As() check string message as well as err type for the block reject error.

I have checked and this fixed the issue. After restarting the enclave and one case of the dup error it started processing blocks successfully.

### Definition of done

### End-to-end tests

- Run the workflow

### DIFF
--- a/go/common/enclave.go
+++ b/go/common/enclave.go
@@ -3,6 +3,7 @@ package common
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	gethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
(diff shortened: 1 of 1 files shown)


tudor-malene (approved):
lgtm

PR #1007, opened 2022-12-22
Use a curl command to request OBX

moraygrieve:

- In the temp absence of the Discord faucet bot we need to provide information on users on how to obtain OBX - this can be achieved by issuing a curl command directly to the faucet server. 

- Section on using the testnet token faucet.


BedrockSquirrel (approved):
LGTM, maybe worth giving an example of the address format they should use? (e.g. `0x75Ad715443e1E2EBdaFA33ABB3B08443966019A6` (don't know if capitalisation thing matters)

PR #1008, opened 2022-12-26
Updates PCCS URL for edgelessdb + ego

otherview:

- https://github.com/obscuronet/obscuro-internal/issues/1273
- In Alibabaclound the PCCS must be specified as the default values don't work.
- Fixed docker-compose env var ingestion ( it was failing to load variables that were not immediately used in the entry command of the container)
- Enclave now starts with an` entry.sh` so it can handle the PCCS address ( needed for ali baba and if you have a custom remote attestation url)
- Enclave is no longer running under docker privileged mode ( changed to used devices like sugested by the edgeless systems )
- Remove the entry point in the docker files as it's misleading that the image may run without any flags

### Definition of done

### End-to-end tests

- Run the workflow

### DIFF
--- a/go/enclave/main/entry.sh
+++ b/go/enclave/main/entry.sh
@@ -1,20 +1,30 @@
 #!/bin/sh
 set -e
+#
+# This script is the entry point for starting the enclave under a Docker container.
+# It allows running SGX sdk using different parameters.
+#
 
+# It's expected to be a link between the /dev/sgx_enclave Docker device and the container /dev/sgx/enclave
 mkdir -p /dev/sgx
 if [ ! -L /dev/sgx/enclave ]; then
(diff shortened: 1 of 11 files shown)


go/enclave/main/entry.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+set -e
```
StefanIliev545, line 2: This one is confusing me, let's add a couple of comments here on what is happening and why for future references.

StefanIliev545 (changes requested):
Looks great, just needs a couple of small comments.

go/enclave/main/entry.sh
otherview: Good point ! Adding them 👍

StefanIliev545 (approved):
LGTM

PR #1013, opened 2023-01-02
Overriden how transaction receipt is marshalled to json.

StefanIliev545:

Fix for #970 

Changed how receipt is being marshalled. Using a cloned function instead of the standard one in the geth library which is not compliant with the library. **It is a bit of a hack, which makes me uneasy, but there is no other way to encode null**

### Definition of done

### End-to-end tests

- Run the workflow

### DIFF
--- a/go/common/json_types.go
+++ b/go/common/json_types.go
@@ -1,0 +1,49 @@
+package common
+
+import (
+	"bytes"
+	"encoding/json"
+
+	gethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+type ReceiptJSON struct {
+	Type              hexutil.Uint64      `json:"type,omitempty"`
+	PostState         hexutil.Bytes       `json:"root"`
+	Status            hexutil.Uint64      `json:"status"`
+	CumulativeGasUsed hexutil.Uint64      `json:"cumulativeGasUsed" gencodec:"required"`
+	Bloom             types.Bloom         `json:"logsBloom"         gencodec:"required"`
+	Logs              []*types.Log        `json:"logs"              gencodec:"required"`
+	TxHash            gethcommon.Hash     `json:"transactionHash" gencodec:"required"`
+	ContractAddress   *gethcommon.Address `json:"contractAddress"`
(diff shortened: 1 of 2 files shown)


StefanIliev545:
Alternatively we can use something that converts the struct to a map and removes the contract address if it matches zero address, however this still overrides how json is generated and I feel it makes it more obfuscated.

go/enclave/enclave.go
```diff
@@ -510,11 +510,13 @@ func (e *enclaveImpl) GetTransactionReceipt(encryptedParams common.EncryptedPara
 	}
 
 	// We marshal the receipt to JSON.
-	txReceiptBytes, err := txReceipt.MarshalJSON()
+	txReceiptBytes, err := common.MarshalReceipt(txReceipt)
 	if err != nil {
 		return nil, fmt.Errorf("could not marshal transaction receipt to JSON in eth_getTransactionReceipt request. Cause: %w", err)
 	}
 
+	fmt.Println(string(txReceiptBytes))
```
otherview, line 518: Lingering print statement.

```suggestion
```

go/common/json_types.go
```diff
@@ -0,0 +1,49 @@
...
+	enc.Status = hexutil.Uint64(r.Status)
+	enc.CumulativeGasUsed = hexutil.Uint64(r.CumulativeGasUsed)
+	enc.Bloom = r.Bloom
+	enc.Logs = r.Logs
+	enc.TxHash = r.TxHash
+	// Contract address should only be set for receipts of transactions that initialize a contract.
+	// Otherwise return nil to match receipts returned by other testnets.
+	if !bytes.Equal(r.ContractAddress.Bytes(), gethcommon.BigToAddress(gethcommon.Big0).Bytes()) {
+		enc.ContractAddress = &r.ContractAddress
+	}
```
otherview, line 43: geth to always return `0x0` contract address when contract address is nil. Is there a reason to change this condition ?

otherview (commented):
I think our guideline is to replicate geth. I wouldn't necessarily change this.. unless we're missing something.

I've updated https://github.com/obscuronet/obscuro-internal/issues/970 with the comment.
It might be that we're drawing incorrect conclusions of bug. ie. We need to replicate the issue described in the test in a geth network, ropstein is run by infura which has a whole myriad of servers running.

PR #1014, opened 2023-01-02
Decouples Host and RPC server

otherview:

- Decouples the Host and the RPC server into the Container Layer
- Starting a network of Socket type container hosts now checks the Health Endpoint before returning

### Definition of done

### End-to-end tests

- Run the workflow

### DIFF
--- a/integration/simulation/network/obscuro_node_utils.go
+++ b/integration/simulation/network/obscuro_node_utils.go
@@ -125,10 +125,15 @@ func startStandaloneObscuroNodes(params *params.SimParams, gethClients []ethadap
 		var client rpc.Client
 		var err error
 
-		// create a connection to the created nodes
+		// create a connection to the newly created nodes - panic if no connection is made after some time
+		startTime := time.Now()
 		for connected := false; !connected; time.Sleep(500 * time.Millisecond) {
 			client, err = rpc.NewNetworkClient(rpcAddress)
 			connected = err == nil // The client cannot be created until the node has started.
+			if time.Now().After(startTime.Add(2 * time.Minute)) {
+				testlog.Logger().Crit("failed to create a connect to node after 2 minute")
+			}
+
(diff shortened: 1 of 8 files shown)


integration/simulation/network/obscuro_node_utils.go
```diff
@@ -117,8 +116,7 @@ func startStandaloneObscuroNodes(params *params.SimParams, gethClients []ethadap
 
 	// start each obscuro node
 	for _, m := range obscuroNodes {
-		t := m
-		go t.Start()
+		_ = m.Start()
```
BedrockSquirrel, line 119: Worth checking the err response and panicking here? I think at the moment it would just silently swallow it if something went wrong at this point in the sim.

BedrockSquirrel (commented):
LGTM, thanks for doing this. Just a minor comment but also I think the failing test is relevant so perhaps something needs sorting there(starting standalone containers test).

integration/simulation/network/obscuro_node_utils.go
otherview: I would think It would too, but it might be misleading as errors are not being bubbled up.
Added a todo and a ticket. WTDT ?

BedrockSquirrel: What do you think would be misleading? At the very least I'd log it at ERROR level `"node might not have started properly, container errored - %w"`

I'm just worried that there might be unpredictable behaviour from a node in a bad state, so it might not fail the smoke tests but then have difficult to diagnose issues later for example.

otherview: Nothing will ever error from `_ = m.Start()`.
`m.Start()` calls `HostContainer.Start()` which always returns nil. 
And inside `HostContainer.Start()`, the `Start()` calls of each of the components do not return error. They instead rely on panic'ing.

That's why I think adding an error check here is misleading as the errors are never bubbled up.
It totally should be changed, but maybe in a different PR?

BedrockSquirrel (approved):
LGTM

PR #1020, opened 2023-01-05
Fix for issue deploying testnet

moraygrieve:

- Fixes mismatch in the load balancer deployment and corrects step title (was a copy and paste)


otherview (commented):
Should we change everything to be `Name-Number` instead of `NameNumber` ?

Asking this because there's a few other places that we're doing `Name-Number` like ` --public-ip-address-dns-name "obscuronode-${{ matrix.host_id }}-testnet-${{ GITHUB.RUN_NUMBER }}" \`

otherview (approved):

PR #1024, opened 2023-01-05
Adds native transfers to the tx injector

otherview:

- https://github.com/obscuronet/obscuro-internal/issues/1294

### Definition of done

### End-to-end tests

- Run the workflow

### DIFF
--- a/integration/simulation/transaction_injector.go
+++ b/integration/simulation/transaction_injector.go
@@ -134,6 +134,11 @@ func (ti *TransactionInjector) Start() {
 	})
 
 	wg.Go(func() error {
+		ti.issueRandomValueTransfers()
+		return nil
+	})
+
+	wg.Go(func() error {
 		ti.issueInvalidL2Txs()
 		return nil
 	})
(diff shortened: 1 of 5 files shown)


integration/simulation/transaction_injector.go
```diff
@@ -150,6 +155,55 @@ func (ti *TransactionInjector) Stop() {
...
 }
 
+// issueRandomValueTransfers creates and issues a number of L2 value transfer transactions proportional to the simulation time, such that they can be processed
+func (ti *TransactionInjector) issueRandomValueTransfers() {
+	for txCounter := 0; ti.shouldKeepIssuing(txCounter); txCounter++ {
+		fromWallet := ti.rndObsWallet()
+		toWallet := ti.rndObsWallet()
+		obscuroClient := ti.rpcHandles.ObscuroWalletRndClient(fromWallet)
+		// We avoid transfers to self, unless there is only a single L2 wallet.
+		for len(ti.wallets.SimObsWallets) > 1 && fromWallet.Address().Hex() == toWallet.Address().Hex() {
```
StefanIliev545, line 165: Should we really avoid this? I'm not sure how ethereum should handle those transactions, but either way I feel like if this must fail, it must be because of obscuro reverting the transaction rather than the test driver.

tools/networkmanager/inject_txs.go
```diff
@@ -215,4 +216,10 @@ func checkL2TxsSuccessful(ctx context.Context, rpcHandles *network.RPCHandles, t
 	} else {
 		println(fmt.Sprintf("Successfully injected %d withdrawals into the L1.", injectedWithdrawals))
 	}
+
+	if notFoundNativeTransfers != 0 {
```
StefanIliev545, line 220: This **if** contains a lot of negatives obfuscating the condition.

integration/simulation/transaction_injector.go
```diff
@@ -150,6 +155,55 @@ func (ti *TransactionInjector) Stop() {
...
+			Nonce:    fromWallet.GetNonceAndIncrement(),
+			Value:    big.NewInt(int64(testcommon.RndBtw(1, 500))),
+			Gas:      uint64(1_000_000),
+			GasPrice: gethcommon.Big1,
+			To:       &toWalletAddr,
+		}
+		// tx := ti.newObscuroTransferTx(fromWallet, toWallet.Address(), testcommon.RndBtw(1, 500))
+		tx, err := obscuroClient.EstimateGasAndGasPrice(txData)
+		if err != nil {
+			panic(err)
```
StefanIliev545, line 179: If the balance of the account is insufficient for the transfer, would the transaction get rejected and propagate an error? I imagine a wallet running out of eth is a possibility, so not sure if we want to panic.

StefanIliev545 (changes requested):
Looks good, just a couple of questions

integration/simulation/transaction_injector.go
otherview: Nah.. we shouldn't avoid this. It's just repeating the tx.
I think it might screw up the validation if we don't though..

tools/networkmanager/inject_txs.go
otherview: Yea, made a small change that makes it more readable I think. WDYT ?

StefanIliev545: It took me a bit to interpret notFoundNativeTransfers as a count (even if compared with 0). Given the condition perhaps `hasMissingNativeTransfers` would be slightly better

StefanIliev545 (approved):
LGTM

PR #1037, opened 2023-01-11
Added tasks for the wallet extension.

StefanIliev545:

### Definition of done

### End-to-end tests

- Run the workflow

### DIFF
--- a/contracts/hardhat.config.ts
+++ b/contracts/hardhat.config.ts
@@ -4,6 +4,8 @@ import "@nomicfoundation/hardhat-toolbox";
 import "hardhat-abi-exporter";
 import "@solidstate/hardhat-bytecode-exporter";
 
+import './tasks/wallet-extension';
+
 const abiExportPath = "./artifacts/abi/";
 const bytecodeExporterPath = "./artifacts/bin/"
 
--- a/contracts/package.json
+++ b/contracts/package.json
@@ -16,6 +16,7 @@
     "@solidstate/hardhat-bytecode-exporter": "^1.1.1",
     "hardhat": "^2.12.4",
     "hardhat-abi-exporter": "^2.10.1",
+    "node-docker-api": "^1.1.22",
     "ts-node": "^10.9.1",
     "typescript": "^4.9.4"
   }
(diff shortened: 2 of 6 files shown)


tudor-malene (approved):
LGTM.
maybe @otherview can have a look as well

otherview (commented):
Why allow Hardhat to orchestrate the wallet extension ? Or in other words, do we want to tie HardHat and the download/build/launch of a docker image together ?
Similarly, hardhat does not orchestrate starting an Ethereum node. It only takes an address and it's blind to anything else.
Implicitly it seems to enforce the need for docker, docker build / docker get, to be done by the task user.

Maybe a simpler scope of HH to register/manage the communication/ VK register with the WE could make these less coupled while providing a lot of help in dev usage ?

StefanIliev545:
> Why allow Hardhat to orchestrate the wallet extension ? Or in other words, do we want to tie HardHat and the download/build/launch of a docker image together ? Similarly, hardhat does not orchestrate starting an Ethereum node. It only takes an address and it's blind to anything else. Implicitly it seems to enforce the need for docker, docker build / docker get, to be done by the task user.
> 
> Maybe a simpler scope of HH to register/manage the communication/ VK register with the WE could make these less coupled while providing a lot of help in dev usage ?

Docker should automatically pull the image when doing `docker run` if it's not available. IMO it makes sense for hardhat to manage the wallet extension from the perspective of ultimately wanting those bits to be extracted as a plugin; 
Hardhat doesn't manage an ethereum node if the network is external, but it still has the integrated one for convenience. Ultimately the end game for the plugin I see is the same - hardhat obscuro:test - spawns wallet extension, hardhat test does not. While we can do it with bash scripts on the CI, they don't make for the greatest UX when doing stuff manually.

otherview:
It's not about scripts. HH is the orchestrator when it's only required for it to be the client of the WE.
At some stage, we had the simulation run on docker too, but we've since backed away from it.

On one hand, we're making the HH coupled with the docker app and docker build image.
And on the other hand instead of starting a wallet and leaving it running while operations are being done, now we need to be weary that HH might start a new wallet.

We can probably change that once we identify the need for a specific HH plugin, but I don't think that's a need atm.

All of that said. I am happy to merge this IF we agree to use this as a dev utility.

PR #1038, opened 2023-01-11
Hardhat deploy plugin and deployment of the layer 1 bits.

StefanIliev545:

In order to make hardhat capable of deploying to our environments.

This one does not deploy to L2 as the wallet extension bits are a prerequisite. 

Hardhat deploy plugin is added. 
The contracts for L1 and testnet specific ones are added.
Dockerfile for the hardhat deployer and a script replacing the one using the old contract deployer.

### Definition of done

### End-to-end tests

- Run the workflow


contracts/deployment_scripts/testnet/001_deploy_tokens.ts
```diff
@@ -0,0 +1,40 @@
...
+        contract: "WrappedERC20",
+        args: [ "POC", "POC" ],
+        log: true,
+    });
+
+
+    await deployments.execute('HOCERC20', {
+        from: deployer,
+        log: true,
+    }, "issueFor", deployer, "1000000000000000000000000000000");
```
otherview, line 30: Similarly, this contract can be executed with:
```
await lock.unlockTime()
```

contracts/hardhat.config.ts
```diff
@@ -31,7 +31,29 @@ const config: HardhatUserConfig = {
     path: abigen.bytecodeExporterPath,
     runOnCompile: true,
     clear: true,
+  },
+  namedAccounts: {
```
otherview, line 35: Are these requirements from the deploy plugin ?

contracts/hardhat.config.ts
```diff
@@ -31,7 +31,29 @@ const config: HardhatUserConfig = {
...
+  console.log(`Failed parsing "config/networks.json" with reason - ${e}`);
+}
+
+try {
+  if (process.env.NETWORK_JSON != null) {
+    config.networks = JSON.parse(process.env.NETWORK_JSON!!);
+  }
+} catch (e) {
+  console.log(`Unable to parse networks configuration from environment variable. Reason is ${e}`);
+}
```
otherview, line 57: Is reading from env and from file a requirement ?
Could the networks.json file be described in this file ?

testnet/testnet-deploy-l1-contracts.sh
```diff
@@ -53,56 +54,33 @@ then
...
-    --nodePort=${l1port} \
-    --l1Deployment \
-    --contractName="Layer1ERC20" \
-    --privateKey=${pkstring}\
-    --constructorParams="Hocus,HOC,1000000000000000000000000000000,${mgmtContractAddr}"
-# storing the contract address to the .env file
-hocErc20Addr=$(docker logs --tail 1 hocerc20deployer)
-echo "HOCERC20ADDR=${hocErc20Addr}" >> "${testnet_path}/.env"
-echo ""
+mgmtContractAddr=$(docker logs --tail 5 hh-l1-deployer | grep -e 'ManagementContract' | cut -c 121-162)
```
otherview, line 80: Can a comment be added here ?

contracts/deployment_scripts/layer1/001_deploy_management_contract.ts
```diff
@@ -0,0 +1,22 @@
+import {HardhatRuntimeEnvironment} from 'hardhat/types';
+import {DeployFunction} from 'hardhat-deploy/types';
```
otherview, line 2: Not seeing the value of using the deploy package. Can you ping across some examples ? 
I would prefer to avoid future-proofing, better to address the needs as we have them.

otherview (commented):
Overall looks good.
The github workflows I believe need to be updated.
I think it's possible to reduce some components.

testnet/hardhatdeployer.Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+COPY --from=0 /home/go-obscuro/tools/walletextension/bin /home/go-obscuro/tools/walletextension/bin
+COPY ./contracts /home/go-obscuro/contracts
+
+WORKDIR /home/go-obscuro/contracts
+RUN rm package-lock.json || true
+RUN npm install
+RUN npx hardhat compile
+COPY ./contracts /home/go-obscuro/contracts
+VOLUME /home/go-obscuro/contracts/deployments
```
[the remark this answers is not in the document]
StefanIliev545: The volume is so that the result of the deployment is persisted. This is to be used later by the L2 deployment that needs to link the bridges. The plugin saves a ton of stuff in this volume. Also if you run twice it will skip deploying scripts that have already been deployed (but will verify everything is there and so on)

contracts/deployment_scripts/testnet/001_deploy_tokens.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
+    const { 
+        deployments, 
+        getNamedAccounts
+    } = hre;
+
+    const {deployer} = await getNamedAccounts();
+
+    await deployments.deploy('HOCERC20', {
```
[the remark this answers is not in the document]
StefanIliev545: Mainly that the plugin persists what it has deployed, makes it accessible by name in other deployments and knows on which network it is deployed so it can manage complicated bits. Also it has some fail safes to prevent deploying when stuff is out of order.

contracts/hardhat.config.ts
StefanIliev545: Not required, but it's nice. Essentially allows you to declare which key belongs to which named account. Otherwise the plugin introduces it.

contracts/hardhat.config.ts
StefanIliev545: I've had the networks json described in here, but it felt like it was bloating the file. When default configs for testnet, simulation and so on are added at least.

contracts/deployment_scripts/layer1/001_deploy_management_contract.ts
StefanIliev545: At the stage of only deploying to L1 there isn't much of a value, apart from the checks if contracts exist and not deploying multiple times, but when deploying to L1 and L2 and requiring to maintain connection and deployed state across two networks it makes it easier due to being able to define companion networks and then calling functions like

```
layer1.execute(...
layer2.execute(...
```

The HRE is not really good by itself for multiple networks.

contracts/deployment_scripts/testnet/001_deploy_tokens.ts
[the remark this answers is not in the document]
StefanIliev545: In the example you gave, hardhat when started instantiates some singleton stuff to the network it was started with and caches a ton of stuff so `hre.ethers` always points to a single network. We'd have to manually instantiate stuff when we want to orchestrate cross network deployments. But yeah, we can do it entirely through ethers, it would just be additional code and manually managing what account is connected and so on.

testnet/hardhatdeployer.Dockerfile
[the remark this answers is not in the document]
otherview: In the past, I've used volume to quickly test changes without rebuilding the whole container, by mounting the volume on the host. I think if that's the approach that's fine, even desirable.

While there is a great caching benefit of using volume it's not desirable to have stateful containers.
Each execution should not read any state from previous executions.

otherview:
Just capturing the offline discussion here
- Move the TS to JS
- Remove the Volume
- Keeping the HH Deploy plugin
- Execution is now able to receive it it's deploying to L1, L2 or both

Please update if something is missing @StefanIliev545  :)

contracts/hardhat.config.ts
otherview: Cool, got it. What about the reading from env and file ? is the file read connected with the docker volume ?

otherview (commented):
Looks good.

Before merging please make sure it works with deploying to dev-testnet and please check with @moraygrieve if it does not impact the obscuro-tests. It shouldn't, but worth checking

tudor-malene (approved):
lgtm (if it works)

I haven't looked in detail, but I trust Pedro has.

PR #1055, opened 2023-01-20
Each rollup references its batches, rather than a set of transactions

joeldudley:

Rollups should point to the batches they are created from. They no longer need to contain the transaction information, since that is stored in the batch.

Each rollup only references one batch for now, as the mapping between rollups and batches is one-to-one. In the future, a single rollup will reference multiple batches.

### Definition of done

### End-to-end tests

- Run the workflow


go/common/headers.go
```diff
@@ -154,36 +154,6 @@ func (r *RollupHeader) Hash() L2RootHash {
 	return hash
 }
 
-func (r *RollupHeader) ToBatchHeader() *BatchHeader {
```
joeldudley, line 157: This was only used for some unit tests, so I removed it.

BedrockSquirrel (commented):
I'm a bit confused here, I thought we needed to be able to go from rollup -> full batches (not by lookup but the full data (compressed)).

So by my understanding, ExtRollup can't just have batch hashes on it, it needs the full tx data of each batch. I may have just missed something, or maybe this is a stepping stone towards that somehow (bit odd if it were though because it had the tx data before).

joeldudley:
I'm not sure. I thought the rollups were just a commitment to the batches distributed by the sequencer.

@tudor-malene - wdyt? Do we need the full compressed batches in the rollup, or just the hashes of the batches we're committing to?

joeldudley:
@tudor-malene Ok thanks, will rework.

go/enclave/core/rollup.go
```diff
@@ -44,29 +38,20 @@ func (r *Rollup) IsGenesis() bool {
...
-		Header:       encryptedRollup.Header,
-		Transactions: transactionBlobCrypto.Decrypt(encryptedRollup.EncryptedTxBlob),
-	}
-}
-
-func (r *Rollup) ToBatch() *Batch {
-	return &Batch{
-		Header:       r.Header.ToBatchHeader(),
-		Transactions: r.Transactions,
+		Header: encryptedRollup.Header,
```
BedrockSquirrel, line 55: Should this populate the batches here?

BedrockSquirrel (commented):
LGTM, small thing I think was missed

BedrockSquirrel (approved):
LGTM, thanks

PR #1056, opened 2023-01-20
Requires the full set of batches to be specified when creating a rollup from a set of batches

joeldudley:

### Definition of done

### End-to-end tests

- Run the workflow

### DIFF
--- a/go/common/rollups.go
+++ b/go/common/rollups.go
@@ -1,6 +1,7 @@
 package common
 
 import (
+	"sort"
 	"sync/atomic"
 )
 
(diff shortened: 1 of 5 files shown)


go/common/rollups.go
```diff
@@ -22,3 +22,12 @@ func (r *ExtRollup) Hash() L2RootHash {
 	r.hash.Store(v)
 	return v
 }
+
+// ExtRollupFromExtBatches converts a set of ExtBatch into an ExtRollup. The head batch is the one stored in the
+// ExtRollup header's `HeadBatchHash` field.
+func ExtRollupFromExtBatches(headBatch *ExtBatch, additionalBatches []*ExtBatch) *ExtRollup {
```
BedrockSquirrel, line 28: I think specifying the headBatch here will be awkward to use. Seems reasonable to make it explicit that the batches in a rollup should be in their natural ordering and then just set the last one to be the head I'd say.

(It would be painful on the reading side to have to sort the batches after reading them from the rollup too).

BedrockSquirrel (commented):
small comment

go/common/rollups.go
joeldudley: I disagree. I worry we'll get bugs because the ordering is wrong.

How about I sort based on the batch numbers in the method, then use the highest number?

BedrockSquirrel: You have to process batches in the right order, you have to retrieve them from storage in the right order. Feels like you'd have to work pretty hard to get them in the wrong order?

But yeah I definitely think they should be written into the rollup in the correct order so no need to split off the end one into a separate param. Can sort them if you think it's necessary but I think we should be able to convince ourselves that they will be ordered already (and if not we should fix the weirdness)

joeldudley: I think it's safest to order them. We can reason they'll arrive in good order now, but if in the future we allow them to come in random order, we'll forget to update here. If it turns out to be a performance concern, we can address.

BedrockSquirrel: Yeah fair enough, we should guarantee they're ordered when written to the L1 so this seems like a sensible place to assert that.

BedrockSquirrel (approved):
LGTM

PR #1065, opened 2023-01-23
Temporarily allow product key sealing of EDB credentials

BedrockSquirrel:

- While we continue developing the upgrade process for nodes in production we need a solution that allows testnet nodes to be upgraded without losing access to their state
- Sealing the EDB credentials with the enclave product key rather than the enclave unique ID provides some security but a determined attacker with insider access would be able to access the secrets so this is solely a stop-gap solution for testnet
- A task has been created here https://github.com/obscuronet/obscuro-internal/issues/1377 with todos in the code that point to that to ensure this doesn't get left in

- Change the code to seal EDB credentials with product key rather than unique ID
- Make it clear with the code and the todo comments that this is a temporary solution that will be rectified before mainnet

### Definition of done

### End-to-end tests

- Run the workflow


tudor-malene (approved):
lgtm

PR #1071, opened 2023-01-24
Juiced the docker. It's now fast.

StefanIliev545:

https://github.com/obscuronet/obscuro-internal/issues/1336

Reduced image sizes: 

```
testnetobscuronet.azurecr.io/obscuronet/enclave            latest    7497b303fd24   6 minutes ago    243MB
testnetobscuronet.azurecr.io/obscuronet/enclave_debug      latest    33c6285d0586   7 minutes ago    633MB
testnetobscuronet.azurecr.io/obscuronet/host               latest    c2ae2b1743fb   7 minutes ago    33.4MB
testnetobscuronet.azurecr.io/obscuronet/gethnetwork        latest    52c72bcd9aa7   19 minutes ago   622MB
testnetobscuronet.azurecr.io/obscuronet/contractdeployer   latest    40e6c07fd13c   26 minutes ago   23.4MB
testnetobscuronet.azurecr.io/obscuronet/obscuroscan        latest    f22ba55edac9   27 minutes ago   19.9MB
testnetobscuronet.azurecr.io/obscuronet/walletextension    latest    77cc666e4d43   27 minutes ago   17.8MB
testnetobscuronet.azurecr.io/obscuronet/hardhatdeployer    latest    b3a999af06e6   45 minutes ago   477MB
```

Builds now use multiple stages - They run a bit faster from scratch now. Incremental rebuilds are much more rapid: 

`=> [build-enclave 3/4] RUN --mount=type=cache,target=/root/.cache/go-build ego-go build  10.5s` down from `300s`

### Definition of done

### End-to-end tests

- Run the workflow


tudor-malene (commented):
I'll defer to @otherview .

Nice work, Stefan!

.github/workflows/manual-deploy-testnet-l1.yml
```diff
@@ -74,7 +74,7 @@ jobs:
 
       - name: 'Build and push image'
         run: |
-          docker build -t ${{env.L1_DOCKER_BUILD_TAG}} -f testnet/gethnetwork.Dockerfile  .
+          DOCKER_BUILDKIT=1 docker build -t ${{env.L1_DOCKER_BUILD_TAG}} -f testnet/gethnetwork.Dockerfile  .
```
otherview, line 77: Can a comment be added explaining the env var ?

.github/workflows/manual-deploy-testnet-l2.yml
```diff
@@ -89,13 +89,13 @@ jobs:
 
       - name: 'Build and push obscuro node images'
         run: |
-          docker build -t ${{env.L2_ENCLAVE_DOCKER_BUILD_TAG}} -f dockerfiles/enclave.Dockerfile  .
+          DOCKER_BUILDKIT=1 docker build -t ${{env.L2_ENCLAVE_DOCKER_BUILD_TAG}} -f dockerfiles/enclave.Dockerfile  .
```
otherview, line 92: Should the var setting be done outside the scope of the cmd ?

```
run: |
          DOCKER_BUILDKIT=1
          docker build -t ${{env.L2_ENCLAVE_DOCKER_BUILD_TAG}} -f dockerfiles/enclave.Dockerfile  .
          docker build -t ${{env.L2_ENCLAVE_DOCKER_BUILD_TAG}} -f dockerfiles/enclave.Dockerfile  .

File: dockerfiles/enclave.Dockerfile [hunk 3]
```diff
@@ -1,31 +1,39 @@
-FROM ghcr.io/edgelesssys/ego-dev:latest
+FROM ghcr.io/edgelesssys/ego-dev:latest AS build-base
 # on the container:
 #   /home/obscuro/data       contains working files for the enclave
 #   /home/obscuro/go-obscuro contains the src
 #
```
otherview, line 5: This should be updated. Like specified that it's the struct of the execution stage not to be confused with the build stage.

dockerfiles/enclave.Dockerfile
```diff
@@ -1,31 +1,39 @@
-FROM ghcr.io/edgelesssys/ego-dev:latest
+FROM ghcr.io/edgelesssys/ego-dev:latest AS build-base
```
otherview, line 1: Why have the intermediate `build-base` step ? Can these steps just exist in the `build-enclave` ?

dockerfiles/enclave.Dockerfile
```diff
@@ -1,31 +1,39 @@
...
+# Trigger new build stage for compiling the enclave
+FROM build-base as build-enclave
 COPY . .
 
-# build binary
 WORKDIR /home/obscuro/go-obscuro/go/enclave/main
-RUN ego-go build
+
+# Build the enclave using the cross image build cache.
+RUN --mount=type=cache,target=/root/.cache/go-build \
```
otherview, line 24: I'm weary that this might break (like builderx did on the faucet) and we had to strip it out for a simpler docker build.

What if it was narrowed to the debug images given that's mainly for local speed improvement ?

dockerfiles/enclave.debug.Dockerfile
```diff
@@ -24,12 +25,19 @@ COPY go.mod .
 COPY go.sum .
 RUN go mod download
 
+FROM get-dependencies as build-enclave
```
otherview, line 28: What happens if this image is built without the `enclave.Dockerfile` image ?
Will it know to build the  `enclave.Dockerfile` image so to have the `build-enclave` available ?

otherview (changes requested):
It's a great improvement and clean up .

It seems like dockerfiles are no longer isolated but instead, the dockerfiles stages are co-dependent.
While it speeds up things, I'd prefer to take the speed hit and avoid split stages across dockerfiles. Each dockerfile should be buildable on it's own.

dockerfiles/enclave.debug.Dockerfile
StefanIliev545: It doesn't need the enclave.Dockerfile. There is no way to share stuff between dockerfiles (without building images first and having multiple steps)

dockerfiles/enclave.Dockerfile
StefanIliev545: There is no real need for it, just seems cleaner this way where stages only do smaller bits of work.

dockerfiles/enclave.Dockerfile
StefanIliev545: This shouldn't break (or in fairness do anything) in the CI process as the runner will always be a fresh one. Also the main speed improvement is coming from the enclave building which takes 300~ seconds normally (signing is 30 seconds)

There is a standard github action for docker build & push that has configurations to allow cross run sharing of the cache, but not much point moving to them.

otherview:
> > It's a great improvement and clean up .
> > It seems like dockerfiles are no longer isolated but instead, the dockerfiles stages are co-dependent. While it speeds up things, I'd prefer to take the speed hit and avoid split stages across dockerfiles. Each dockerfile should be buildable on it's own.
> 
> Each dockerfile is buildable on its own still. There is no codependency and no way to have codependency between dockerfiles, apart from building one and tagging it and then using `FROM image-tag` which is the standard behaviour.
> 
> The split stages allow docker to figure out what can be ran in parallel, for example building the geth binary does not need to wait for downloading the modules of the obscuro projects. Docker figures this out with the multi stage builds and launches both to run at the same time. Same goes for `npm install` in parallel to `go mod download => build wallet extension` for the new contract deployer.
> 
> While the stages have redundant code across different docker files, they don't produce intermediate shared images. The naming of those stages is entirely locally resolved within a dockerfile.

otherview:
fat fingered

otherview:
> > It's a great improvement and clean up .
> > It seems like dockerfiles are no longer isolated but instead, the dockerfiles stages are co-dependent. While it speeds up things, I'd prefer to take the speed hit and avoid split stages across dockerfiles. Each dockerfile should be buildable on it's own.
> 
> Each dockerfile is buildable on its own still. There is no codependency and no way to have codependency between dockerfiles, apart from building one and tagging it and then using `FROM image-tag` which is the standard behaviour.
> 
> The split stages allow docker to figure out what can be ran in parallel, for example building the geth binary does not need to wait for downloading the modules of the obscuro projects. Docker figures this out with the multi stage builds and launches both to run at the same time. Same goes for `npm install` in parallel to `go mod download => build wallet extension` for the new contract deployer.
> 
> While the stages have redundant code across different docker files, they don't produce intermediate shared images. The naming of those stages is entirely locally resolved within a dockerfile.

That's right. I misread the docker files and assumed you were pushing one `build-base/system` and reusing it given the same docker registry. Scrap that please.

After reviewing the second time I get the logical separation of system setup and build setup. Can we add a header to the top of each dockerfile with a one line explaning of the build stages ?

dockerfiles/enclave.debug.Dockerfile
otherview: Yep, misread the dockerfiles. I was thinking of layer re-usage in the docker registry.

dockerfiles/enclave.Dockerfile
otherview: OOC, why is `--mount=type=cache,target=/root/.cache/go-build  ego-go build` faster ?

dockerfiles/enclave.Dockerfile
otherview: Repeated, but having a top header explaining the stages would be awesome :)

dockerfiles/enclave.Dockerfile
StefanIliev545: It shares the go compiler build cache across the layer rebuilds of the image and thus allows the go compiler to work incrementally rather than recompile everything from scratch everytime. This is a speedup only for the local builds obviously as there isn't any cache on the CI. The speedup on the CI is because of the docker builder being able to launch downloads in parallel. (the CI l2 image build drops from 9 mintues to 6, not that big of a difference)

otherview: Cool, thanks for the info :)

otherview (approved):
Good stuff @StefanIliev545 !

otherview:
Please make sure you've tested these in the CI before merging 🙏 @StefanIliev545

PR #1088, opened 2023-02-02
After enclave restart, replay batches to restore missing stateDB cache

BedrockSquirrel:
### Why this change is needed

After an enclave is restarted if the stateDB cache wasn't fully flushed then it fails to process batches because it can't find previous stateDB to build onto.

Ethereum has this problem too, when its node starts it moves its head backwards until it finds a block with stateDB cache intact then replays forwards. We do something similar here.

After starting up an enclave, if there is a head batch in the DB then it makes sure it has stateDB for that batch.

If not it steps backwards along the batch-chain, building up a list of batches to replay until it reaches genesis or finds a batch with existing stateDB.

Then it steps forwards through the list of batches and processes their transactions again to restore the stateDB caches.

### PR checks pre-merging

### DIFF
--- a/go/enclave/enclave.go
+++ b/go/enclave/enclave.go
@@ -188,6 +188,11 @@ func NewEnclave(
 		genesis,
 		logger,
 	)
+	// ensure cached chain state data is up-to-date using the persisted batch data
+	err = chain.ResyncStateDB()
+	if err != nil {
+		logger.Crit("failed to resync L2 chain state DB after restart", log.ErrKey, err)
+	}
 
 	jsonConfig, _ := json.MarshalIndent(config, "", "  ")
 	logger.Info("Enclave service created with following config", log.CfgKey, string(jsonConfig))
--- a/go/enclave/l2chain/l2_chain.go
+++ b/go/enclave/l2chain/l2_chain.go
@@ -566,8 +566,9 @@ func (oc *ObscuroChain) ResyncStateDB() error {
 }
 
 // replayBatchesToValidState is used to repopulate the stateDB cache with data from persisted batches. Two step process:
-// 1. step backwards from head batch until we find a batch that is already in stateDB catch, builds list of batches to replay
+// 1. step backwards from head batch until we find a batch that is already in stateDB cache, builds list of batches to replay
 // 2. iterate that list of batches from the earliest, process the transactions to calculate and cache the stateDB
+// todo: get unit test coverage around this (and L2 Chain code more widely, see ticket #1416
 func (oc *ObscuroChain) replayBatchesToValidState() error {
 	// this slice will be a stack of batches to replay as we walk backwards in search of latest valid state
 	// todo: consider capping the size of this batch list using FIFO to avoid memory issues, and then repeating as necessary
--- a/go/enclave/l2chain/l2_chain.go
+++ b/go/enclave/l2chain/l2_chain.go
@@ -548,8 +548,12 @@ func (oc *ObscuroChain) processState(batch *core.Batch, txs []*common.L2Tx, stat
 func (oc *ObscuroChain) ResyncStateDB() error {
 	batch, err := oc.storage.FetchHeadBatch()
 	if err != nil {
-		oc.logger.Info("no head batch found in DB after restart", log.ErrKey, err)
-		return nil
+		if errors.Is(err, errutil.ErrNotFound) {
+			// there is no head batch, this is probably a new node - there is no state to rebuild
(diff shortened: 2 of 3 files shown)


go/enclave/enclave.go
```diff
@@ -188,6 +188,11 @@ func NewEnclave(
 		genesis,
 		logger,
 	)
+	// ensure cached chain state data is up-to-date using the persisted batch data
```
otherview, line 191: Should this happen on Start instead of on New ?

go/enclave/l2chain/l2_chain.go
```diff
@@ -543,6 +543,80 @@ func (oc *ObscuroChain) processState(batch *core.Batch, txs []*common.L2Tx, stat
 	return rootHash, executedTransactions, txReceipts, synthReceipts
 }
 
+// ResyncStateDB can be called to ensure stateDB data is available for the canonical L2 batch chain
+// After an (ungraceful) shutdown this method must be called to rebuild the stateDB data based on the persisted batches
+func (oc *ObscuroChain) ResyncStateDB() error {
+	batch, err := oc.storage.FetchHeadBatch()
+	if err != nil {
+		oc.logger.Info("no head batch found in DB after restart", log.ErrKey, err)
+		return nil
```
otherview, line 552: This might be misleading. Should the error be checked + surfaced ?

go/enclave/l2chain/l2_chain.go
```diff
@@ -543,6 +543,80 @@ func (oc *ObscuroChain) processState(batch *core.Batch, txs []*common.L2Tx, stat
...
+	return nil
+}
+
+// replayBatchesToValidState is used to repopulate the stateDB cache with data from persisted batches. Two step process:
+// 1. step backwards from head batch until we find a batch that is already in stateDB catch, builds list of batches to replay
+// 2. iterate that list of batches from the earliest, process the transactions to calculate and cache the stateDB
+func (oc *ObscuroChain) replayBatchesToValidState() error {
+	// this slice will be a stack of batches to replay as we walk backwards in search of latest valid state
+	// todo: consider capping the size of this batch list using FIFO to avoid memory issues, and then repeating as necessary
+	batchesToReplay := make([]*core.Batch, 0)
```
otherview, line 570: ooc, why make an empty slice ?

go/enclave/l2chain/l2_chain.go
```diff
@@ -543,6 +543,80 @@ func (oc *ObscuroChain) processState(batch *core.Batch, txs []*common.L2Tx, stat
...
+		// we don't need the return values, just want the post-batch state to be cached
+		oc.processState(batch, batch.Transactions, prevState)
+	}
+
+	return nil
+}
+
+func stateDBAvailableForBatch(storage db.Storage, hash *common.L2RootHash) bool {
+	_, err := storage.CreateStateDB(*hash)
+	return err == nil
```
otherview, line 617: If we create a statedb and it errors then it's not available for batch ? 
Can a comment be added explaining the underlying logic ?

otherview (commented):
We should have a unit test around this stuff.

go/enclave/enclave.go
BedrockSquirrel: Enclave doesn't have a start - agree that some of the stuff in this constructor probably belongs in a start though, but in a different PR perhaps?

go/enclave/l2chain/l2_chain.go
BedrockSquirrel: Cool! I didn't know you could do that, didn't know append worked with nil slices.

go/enclave/enclave.go
otherview: Fair point. Would add a TODO 👍

otherview (approved):

go/enclave/l2chain/l2_chain.go
```diff
@@ -566,8 +566,9 @@ func (oc *ObscuroChain) ResyncStateDB() error {
 }
 
 // replayBatchesToValidState is used to repopulate the stateDB cache with data from persisted batches. Two step process:
-// 1. step backwards from head batch until we find a batch that is already in stateDB catch, builds list of batches to replay
+// 1. step backwards from head batch until we find a batch that is already in stateDB cache, builds list of batches to replay
 // 2. iterate that list of batches from the earliest, process the transactions to calculate and cache the stateDB
+// todo: get unit test coverage around this (and L2 Chain code more widely, see ticket #1416
```
otherview, line 571:
```suggestion
// todo: get unit test coverage around this (and L2 Chain code more widely, see ticket #1416 )
```

PR #1089, opened 2023-02-03
Rollup lookup should not assume that L1 blocks go back to genesis

BedrockSquirrel:
### Why this change is needed

We were sometimes seeing failures before rollups were published where the rollup lookup would try to go back to genesis but it would run out of L1 blocks because the enclave didn't start from genesis, it started from the mgmt contract deployment block.

Return rollup not found when the L1 blocks run out.

### PR checks pre-merging

### DIFF
--- a/go/enclave/l2chain/l2_chain.go
+++ b/go/enclave/l2chain/l2_chain.go
@@ -913,11 +913,6 @@ func (oc *ObscuroChain) processRollups(block *common.L1Block) error {
 // Given a block, returns the latest rollup in the canonical chain for that block (excluding those in the block itself).
 func (oc *ObscuroChain) getLatestRollupBeforeBlock(block *common.L1Block) (*core.Rollup, error) {
 	for {
-		// We've reached the genesis block, so we cannot go back any further.
-		if block.Number().Cmp(big.NewInt(0)) == 0 {
-			return nil, nil //nolint:nilnil
-		}
-
 		blockParentHash := block.ParentHash()
 		latestRollup, err := oc.storage.FetchHeadRollupForBlock(&blockParentHash)
 		if err != nil && !errors.Is(err, errutil.ErrNotFound) {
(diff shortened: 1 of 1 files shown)


otherview (approved):

go/enclave/l2chain/l2_chain.go
```diff
@@ -929,7 +926,9 @@ func (oc *ObscuroChain) getLatestRollupBeforeBlock(block *common.L1Block) (*core
 
 		block, err = oc.storage.FetchBlock(block.ParentHash())
 		if err != nil {
-			return nil, fmt.Errorf("could not fetch parent block")
+			// No rollup found - no more blocks available (enclave does not read the L1 chain from genesis if it knows
+			// when management contract was deployed, so we don't keep going to block zero, we just stop when the blocks run out)
+			return nil, errNoRollupFound
```
otherview, line 931: Wait isn't this the block that was not found ?

BedrockSquirrel: It was scanning backwards through blocks looking for a rollup. If it runs out of blocks then there was no rollup. I'll update the comment to try to make that clearer

otherview: Also should we surface that block error ?

go/enclave/l2chain/l2_chain.go
```diff
@@ -37,6 +37,8 @@ import (
 	gethrpc "github.com/ethereum/go-ethereum/rpc"
 )
 
+var errNoRollupFound = errors.New("no rollups found")
```
otherview, line 40:
```suggestion
var errNoRollupFound = errors.New("no rollup found")
```

go/enclave/l2chain/l2_chain.go
```diff
@@ -913,23 +915,24 @@ func (oc *ObscuroChain) processRollups(block *common.L1Block) error {
...
 		block, err = oc.storage.FetchBlock(block.ParentHash())
 		if err != nil {
-			return nil, fmt.Errorf("could not fetch parent block")
+			if errors.Is(err, errutil.ErrNotFound) {
+				// No more blocks available (enclave does not read the L1 chain from genesis if it knows
+				// when management contract was deployed, so we don't keep going to block zero, we just stop when the blocks run out)
+				// We have now checked through the entire (relevant) history of the L1 and no rollups were found.
+				return nil, errNoRollupFound
+			}
+			return nil, fmt.Errorf("could not fetch prev block - %w", err)
```
otherview, line 935:
```suggestion
			return nil, fmt.Errorf("could not fetch parent block - %w", err)
```

otherview (approved):

PR #1094, opened 2023-02-06
Fix for contract CI go test integration. Reintroduced param names to …

StefanIliev545:
### Why this change is needed

The PR gate was not always recompiling the solidity contracts. This means that what the go code uses and what the deployed contracts look like may end up differ. 

Fixed a bug introduced by removing the param names to avoid warning. Added comments to ignore the warning using the plugin we already have. Added workflow steps to build the contracts with hardhat before compiling the code.

### PR checks pre-merging


tudor-malene (commented):
LGTM.
would be good if @otherview would have a look as well

.github/workflows/build-pr.yml
```diff
@@ -15,6 +15,19 @@ jobs:
...
+      - name: Install abigen
+        run: |
+          sudo add-apt-repository -y ppa:ethereum/ethereum
+          sudo apt-get update
+          sudo apt-get -y install ethereum
+
+      - name: Install dependencies
+        run: cd contracts && npm install
+      - name: Generate Solidity Contract Bindings
+        run: cd contracts && npx hardhat generate-abi-bindings --output-dir generated
```
otherview, line 30: This may lead to the committed generated files mismatching the solidity files. 

Instead of this step would add a "Does solidity match the generated files" test in https://github.com/obscuronet/obscuro-internal/issues/1424

go/enclave/enclave.go
```diff
@@ -1031,6 +1031,10 @@ func (e *enclaveImpl) produceBlockSubmissionResponse(block *types.Block, l2Head
 	if producedBatch != nil {
 		producedExtBatch = producedBatch.ToExtBatch(e.transactionBlobCrypto)
 		producedExtRollup = common.ExtRollupFromExtBatches([]*common.ExtBatch{producedExtBatch})
+		err := e.chain.SignRollup(producedExtRollup.Header)
+		if err != nil {
+			panic(err) // This should never happen!
+		}
```
otherview, line 1037: Should prolly move this to another PR ? and add tests 👀

go/enclave/l2chain/l2_chain.go
```diff
@@ -697,6 +697,16 @@ func (oc *ObscuroChain) signBatch(batch *core.Batch) error {
 	return nil
 }
 
+func (oc *ObscuroChain) SignRollup(header *common.RollupHeader) error {
```
otherview, line 700: Should prolly move this to another PR ? and add tests 👀

otherview (commented):
lgtm, guess there's some code that slipped in ?

otherview (approved):

PR #1101, opened 2023-02-08
Record latest rollup for every L1 block to avoid scanning back

BedrockSquirrel:
### Why this change is needed

Currently we scan back through the L1 chain doing db lookups for rollups. When no rollup has been published we scan back to the start of the chain on **every** L1 block, causing thrashing and bringing the network to a halt.

Record the latest rollup against every L1 hash (even if it's no rollup) so we don't have to scan back. FWIW we used to do something like this and stopped at some point.

This might not be super efficient but it fixes the problem for now and we can optimise once we've built more out imo.

### PR checks pre-merging

### DIFF
--- a/go/enclave/db/storage.go
+++ b/go/enclave/db/storage.go
@@ -3,6 +3,7 @@ package db
 import (
 	"bytes"
 	"crypto/ecdsa"
+	"errors"
 	"fmt"
 	"math/big"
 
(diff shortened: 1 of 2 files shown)


tudor-malene (approved):
lgtm.
I think this is similar to what we were doing before

StefanIliev545:
Would this work after restarting?

BedrockSquirrel:
Yep should be fine for restart. Enclave remembers which L1 blocks it's seen.

PR #1107, opened 2023-02-10
Add a single start script to launch testnet with default arguments

moraygrieve:
### Why this change is needed

Makes it easier for everyone to test starting up a local testnet if there is a single start script, e.g. that obscuro-test can call.


otherview (approved):
lgtm!

StefanIliev545 (approved):

PR #1112, opened 2023-02-14
halfway adding testnet

otherview:
### Why this change is needed

https://github.com/obscuronet/obscuro-internal/issues/1437

### PR checks pre-merging


go/node/cmd/main.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"github.com/obscuronet/go-obscuro/go/node"
+)
+
+func main() {
+	cliConfig := ParseConfigCLI()
+	// todo: allow for multiple operation (start, stop, status)
+
+	nodeCfg := node.NewNodeConfig(
+		node.WithNodeName(cliConfig.nodeName),
```
BedrockSquirrel, line 12: I love this pattern but feel like it mostly makes sense in a more DSL-style scenario. Especially one where most of the values will be defaulted and the caller will be just overriding one or two. Is that how you see this mostly being used?

go/node/docker.go
```diff
@@ -0,0 +1,152 @@
...
+	"fmt"
+
+	"github.com/obscuronet/go-obscuro/go/common/docker"
+)
+
+type DockerNode struct {
+	cfg *Config
+}
+
+func NewDockerNode(cfg *Config) (Node, error) {
```
BedrockSquirrel, line 13: Nice, so I see you've made this guy pretty much implements the NodeOperator interface, so should be easy to plug in as a networktest env

go/node/docker.go
```diff
@@ -0,0 +1,152 @@
...
+	return &DockerNode{
+		cfg: cfg,
+	}, nil // todo: add config validation
+}
+
+func (d *DockerNode) Start() error {
+	// TODO this should probably be removed in the future
+
+	err := d.startEdgelessDB()
```
BedrockSquirrel, line 23: what is that TODO saying should be removed? The startEdgelessDB call? I feel like it should probably be startDatabase() or something to be at the same level of abstraction as host and enclave.

go/node/docker.go
```diff
@@ -0,0 +1,152 @@
...
+		)
+	}
+
+	_, err := docker.StartNewContainer(d.cfg.nodeName+"-enclave", d.cfg.enclaveImage, cmd, exposedPorts, envs, devices)
+	return err
+}
+
+func (d *DockerNode) startEdgelessDB() error {
+	if !d.cfg.sgxEnabled {
+		return nil
```
BedrockSquirrel, line 136: Maybe worth a comment to say that on non-SGX hardware we use an in-mem sqlite database so no docker process required.

go/node/node.go
```diff
@@ -0,0 +1,5 @@
+package node
+
+type Node interface {
```
BedrockSquirrel, line 3: What uses this interface? Very hard to search and check, 226 matches in the browser search haha

integration/manualtests/tx_test.go
```diff
@@ -67,3 +77,60 @@ func TestIssueTxWaitReceipt(t *testing.T) {
...
+
+	fmt.Printf("Created Tx: %s \n", signedTx.Hash().Hex())
+	fmt.Printf("Checking for tx receipt for %s \n", signedTx.Hash())
+	var receipt *types.Receipt
+	for start := time.Now(); time.Since(start) < time.Minute; time.Sleep(time.Second) {
+		receipt, err = authClient.TransactionReceipt(ctx, signedTx.Hash())
+		if err == nil {
+			break
+		}
+		//if !errors.Is(err, ethereum.NotFound) {
```
BedrockSquirrel, line 138: should this be uncommented/deleted?

testnet/launcher/docker.go
```diff
@@ -0,0 +1,243 @@
...
+	return managementContractAddr, messageBusContractAddr, nil
+}
+
+// waitForHealthyNode retries continuously for the node to respond to a healthcheck http request
+func waitForHealthyNode(port int) error { // todo: hook the cfg
+	requestURL := fmt.Sprintf("http://localhost:%d", port)
+	reqBody := `{"method": "obscuro_health", "id": 1}`
+
+	fmt.Println("Waiting for obscuro node to be healthy...")
+	for startTime := time.Now(); time.Now().Before(startTime.Add(2 * time.Minute)); time.Sleep(time.Second) {
```
BedrockSquirrel, line 209: I haven't checked but I imagine this will come out nicer with the `retry.Do()` thing and moving some of the low level processing into a separate function. You've got a nestif ignore and I imagine you're getting close to needing a complexity ignore too.

BedrockSquirrel (commented):
Nice, loving this! Comments are mostly stream of consciousness thoughts, nothing major I don't think.

I'm going to look at what changes have been made since I started the review now and think about it at a higher level a bit.

otherview:
> @otherview can you update the readme section on starting up a local testnet?

Updated the info, let me know if I missed something 🙏

.github/workflows/manual-deploy-testnet-l2.yml
```diff
@@ -226,19 +235,21 @@ jobs:
...
-               --sequencerID=${{ secrets.GETHNETWORK_PREFUNDED_ADDR_0 }} \
-               --dev_testnet=${{ (github.event.inputs.testnet_type == 'dev-testnet') || (github.event_name == 'schedule') }} \
-               --p2p_public_address=obscuronode-${{ matrix.host_id }}-${{needs.build.outputs.RESOURCE_TESTNET_NAME}}-${{ GITHUB.RUN_NUMBER }}.uksouth.cloudapp.azure.com:10000'
+            && cd /home/obscuro/go-obscuro/ \
+            && sudo go run /home/obscuro/go-obscuro/go/node/cmd  \
+               -is_genesis=${{ matrix.is_genesis }} \
+               -node_type=${{ matrix.node_type }} \
+               -is_sgx_enabled=true \
+               -host_id=${{ secrets[matrix.node_pk_addr] }} \
+               -l1_addr=${{needs.build.outputs.L1_HOST}} \
```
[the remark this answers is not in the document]
otherview: Fixed ! 👍

go/common/docker/docker.go
```diff
@@ -0,0 +1,195 @@
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func waitAndPullRemoteImage(image string, cli *client.Client) error {
+	// Pull the image from remote
+	fmt.Printf("Image %s not found locally. Pulling from remote...\n", image)
```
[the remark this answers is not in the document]
otherview: 100% agree. Let me add a ticket so we can track this.

go/common/docker/docker.go
```diff
@@ -0,0 +1,195 @@
...
+	// Wait for the container to finish with a timeout of one minute
+	statusCh, errCh := cli.ContainerWait(context.Background(), containerID, container.WaitConditionNotRunning)
+	select {
+	case err = <-errCh:
+		if err != nil {
+			return err
+		}
+	case status := <-statusCh:
+		if status.StatusCode != 0 {
+			return fmt.Errorf("container exited with non-zero status code")
```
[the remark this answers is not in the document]
otherview: Fixed!

go/common/docker/docker.go
```diff
@@ -0,0 +1,195 @@
...
+		}
+	case <-time.After(timeout):
+		return fmt.Errorf("timeout waiting for container to finish")
```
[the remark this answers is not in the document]
otherview: fixed !

go/node/README.md
```diff
@@ -0,0 +1,3 @@
+# Node package
+
+This package is responsible for orchestrating the start of various components that render the obscuro node.
```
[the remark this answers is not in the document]
otherview: Yep makes sense 👍

go/node/cmd/main.go
otherview: I see this being used in multiple instances (like the testnet) that's why it's using functional options.
As for the defaults I'd preferably want to avoid them, they require implicit knowledge which I try to avoid.
We can, however, build that pattern on top of this one, I've used something similar in the past.

Something like

```
nodeCfg := node.NewNodeConfig( nodetypes.DefaultSequencer(), node.WithEnclaveImage(...) )
```

Where the ` nodetypes.DefaultSequencer()` preloads a bunch of defaults. But, these are defaults of the ` nodetypes.DefaultSequencer()` not of the `node`. If that makes sense ?

go/node/docker.go
otherview: Yeah, hopefully these all plugin and play nicely :)

go/node/docker.go
otherview: I added a print there, which for some reason didn't make it. 
It's printing the enclave execution options, helpful for debugging but should be removed.

go/node/docker.go
otherview: Good point ! Added !

integration/manualtests/tx_test.go
otherview: That's a bug I found actually - but you're right it needs more info

go/node/cmd/main.go
BedrockSquirrel: Cool cool, yeah that's a nice pattern

testnet/launcher/docker.go
otherview: I've actually used it previously, it's a good package to have, just forgot it :) Added it now .
I think it actually makes the code cleaner.

testnet/launcher/docker.go
```diff
@@ -206,40 +207,42 @@ func waitForHealthyNode(port int) error { // todo: hook the cfg
...
-			if h, ok := r.(map[string]interface{}); ok {
-				if overallHealth := h["OverallHealth"]; overallHealth != nil {
-					if health, ok := overallHealth.(bool); ok && health {
-						fmt.Println("obscuro node is ready")
-						return nil
+	return retry.Do(
+		func() error {
+			req, err := http.NewRequestWithContext(context.Background(), http.MethodGet, requestURL, bytes.NewBufferString(reqBody))
+			if err != nil {
+				return fmt.Errorf("client: could not create request: %w", err)
```
BedrockSquirrel, line 214: Maybe want to wrap this one in the `retry.FailFast(err)` wrapper, so it doesn't keep retrying to keep the behaviour you had before

BedrockSquirrel (approved):
LGTM, I guess we're hoping this will JustWork for the testnets etc. but hard to guess that stuff until it goes through

testnet/launcher/docker.go
otherview: I might be getting this wrong. Please let me know.
I want it to never fail until it hits the sucess / timeout. Because the node might be unresponsive or might not be healthy yet it will throw errors, that we will ignore.

Does that make sense ?

otherview:
> LGTM, I guess we're hoping this will JustWork for the testnets etc. but hard to guess that stuff until it goes through

Found 2 bug and a confusing flag so far, that it was a pain in the arse to verify so far. Big hopes for this testnet launcher approach 🤞  😅

testnet/launcher/docker.go
BedrockSquirrel: Yes that makes sense.

The way you had it before it would do that, except for errors from the NewRequestWithContext call, in that case you returned the error instead of doing a `continue` to try again.

I was just saying if you wanted to keep that behaviour for that one error check, then you could wrap it in a FailFast. But maybe you've decided that wasn't the right behaviour and better to always retry.

otherview: Ahh I see. It's a moot. The `http.NewRequestWithContext` should never fail.

moraygrieve (approved):

PR #1117, opened 2023-02-20
No need to request secret for enclave on host restart

BedrockSquirrel:
### Why this change is needed

Host only needs to request secret after startup if its enclave is waiting for a secret (i.e. when the node initially starts)

Check the status of enclave before requesting secret for it

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #1123, opened 2023-02-22
Save deployment container logs as build artifacts for fault finding

moraygrieve:
### Why this change is needed

On failure of the L1 or L2 contract deployment, extract and uploads the container logs as build artifacts so we can diagnose from the action UI.


BedrockSquirrel (approved):
LGTM

PR #1129, opened 2023-02-23
Run a schedule deploy on the L1, and trigger L2 if succesful

moraygrieve:
### Why this change is needed

We need to schedule L1 deployment followed by L2, rather than just doing an L2 deployment nightly. This adds a schedule trigger to L1 to deploy into dev-testnet, and if successul it will send a repository dispatch to perform the L2 deployment.

### DIFF
--- a/.github/workflows/manual-deploy-testnet-l1.yml
+++ b/.github/workflows/manual-deploy-testnet-l1.yml
@@ -14,6 +14,8 @@
 name: '[M] Deploy Testnet L1'
 
 on:
+  schedule:
+    - cron: '05 3 * * *'
   workflow_dispatch:
     inputs:
       testnet_type:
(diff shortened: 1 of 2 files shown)


.github/workflows/manual-deploy-testnet-l1.yml
```diff
@@ -96,3 +98,8 @@ jobs:
           ports: '8025 8026 9000 9001'
           cpu: 2
           memory: 8
+
+      - name: 'Dispatch trigger'
+        if: ${{ github.event_name == 'schedule' }}
```
moraygrieve, line 103: Note by default this if condition will only trigger on a successful deployment (i.e you dont need to add an extra condition on if success())

BedrockSquirrel (approved):
LGTM, I'll take your word for it on the cron syntax, I can see it's close to what it was for the old L2 deploy

PR #1130, opened 2023-02-23
Resilient to RPC requests while sequencer unknown

BedrockSquirrel:
### Why this change is needed

I hit issues while testing node restart where an RPC request at the wrong time could cause index out of bounds errors when looking for sequencer.

We'd probably want other safe guards against that but at the very least we shouldn't just panic, we should return sensible error and recover.

Return error when sequencer requested but not known

### PR checks pre-merging


StefanIliev545 (approved):
LGTM

PR #1131, opened 2023-02-23
Enclave should not be automatically stopped when host stopping

BedrockSquirrel:
### Why this change is needed

I don't think the lifecycles of host and enclave are so tightly coupled that we should always attempt to shutdown the enclave when shutting down the host.

When node operator starts up they start the enclave then the host, it would be cleaner I think to be able to shutdown the host then the enclave cleanly.

Remover enclave stop from host `Stop()`

### PR checks pre-merging


StefanIliev545 (approved):

BedrockSquirrel:
This branch is on hold unfortunately, turns out the way the sims are wired up they rely on the host stopping the enclave and the wiring is such a mess that it's not worth fixing right now for this.

PR #1133, opened 2023-02-24
Fix to the format of the dispatch event

moraygrieve:
### Why this change is needed

Fixed the format error in the dispatch event.


StefanIliev545 (approved):
lgtm

PR #1143, opened 2023-03-03
Update the go versions on actions to match the repo version

BedrockSquirrel:
### Why this change is needed

Seeing weird lint failures, noticed go version was older than repo, thought I'd update in case it's relevant.

Edit: also had to set `skip-go-installation: true` flag on the lint step to stop the weird failures.

### PR checks pre-merging

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -22,7 +22,7 @@ jobs:
     steps:
       - uses: actions/setup-go@v3
         with:
-          go-version: 1.17
+          go-version: 1.18
       - uses: actions/checkout@v3
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/golangci-lint.yml
```diff
@@ -22,11 +22,13 @@ jobs:
     steps:
       - uses: actions/setup-go@v3
         with:
-          go-version: 1.17
+          go-version: 1.18
       - uses: actions/checkout@v3
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v2
         with:
+          # golangci will install a stable go version by default, overriding the version specified above
```
otherview, line 30: image

I think we always use the latest github action, wondering if that's an issue.

BedrockSquirrel: We're using v2 which installs its own go version and still has the skip flag. I guess if we switched to golangci v3 then this goes away? But don't know if there are other changes to be aware of with that./

otherview: Maybe worth updating yeah :)

BedrockSquirrel:
@otherview that seems to have done the trick, good shout!

otherview (approved):

PR #1149, opened 2023-03-09
Set host and enclave loggers to INFO by default

BedrockSquirrel:
### Why this change is needed

At some point the enclave logs on testnet were set above INFO and we started losing visibility of the goings on.

We should continue to reduce logs where possible by selectively downgrading and removing log messages. If we still feel there are too many logs then perhaps we set testnet to WARN but keep dev testnet at INFO level.

Change default log levels to be human readable and set to INFO in the testnet docker launch script.

### PR checks pre-merging

### DIFF
--- a/go/node/docker.go
+++ b/go/node/docker.go
@@ -2,6 +2,7 @@ package node
 
 import (
 	"fmt"
+	"github.com/ethereum/go-ethereum/log"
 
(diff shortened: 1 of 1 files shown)


moraygrieve (approved):
LGTM

PR #1151, opened 2023-03-10
Small logging improvements, loosen isLatest restriction

BedrockSquirrel:
- Make the enclave logging a bit clearer
- Provide visibility on how far behind of L1 head we are
- Loosen `isLatest` to be "within 1 block of current head" since it's just a best-effort measure anyway (validators have no idea if sequencer had received the latest L1 block they've seen yet) - this change probably makes no difference in practice, especially when we decouple batch production

### PR checks pre-merging

### DIFF
--- a/go/ethadapter/blockprovider.go
+++ b/go/ethadapter/blockprovider.go
@@ -58,15 +58,17 @@ func (e *EthBlockProvider) StartStreamingFromHeight(height *big.Int) (*host.Bloc
 	return &host.BlockStream{Stream: streamCh, Stop: cancel}, nil
 }
 
-func (e *EthBlockProvider) IsLive(b *types.Block) bool {
+// IsLatest returns true iff the block has the same hash as the L1 head block (from the eth client)
+func (e *EthBlockProvider) IsLatest(b *types.Block) bool {
 	l1Head, err := e.ethClient.FetchHeadBlock()
 	if err != nil {
 		e.logger.Warn("unable to fetch head eth block - %w", err)
 		return false
 	}
+	isLatest := b.Hash() == l1Head.Hash()
 	// this log message is helpful for visibility on how far behind the block feeding is
-	e.logger.Info("L1 block provider live-monitoring", "currBlock", b.NumberU64(), "head", l1Head.NumberU64())
-	return b.NumberU64() >= l1Head.NumberU64()-1
+	e.logger.Info("L1 block provider live-monitoring", "currBlock", b.NumberU64(), "head", l1Head.NumberU64(), "isLatest", isLatest)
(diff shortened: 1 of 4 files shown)


go/ethadapter/blockprovider.go
```diff
@@ -58,9 +58,15 @@ func (e *EthBlockProvider) StartStreamingFromHeight(height *big.Int) (*host.Bloc
 	return &host.BlockStream{Stream: streamCh, Stop: cancel}, nil
 }
 
-func (e *EthBlockProvider) IsLive(h gethcommon.Hash) bool {
+func (e *EthBlockProvider) IsLive(b *types.Block) bool {
```
StefanIliev545, line 61: This function needs a comment I feel.

go/ethadapter/blockprovider.go
```diff
@@ -58,9 +58,15 @@ func (e *EthBlockProvider) StartStreamingFromHeight(height *big.Int) (*host.Bloc
...
 	l1Head, err := e.ethClient.FetchHeadBlock()
-	return err == nil && h == l1Head.Hash()
+	if err != nil {
+		e.logger.Warn("unable to fetch head eth block - %w", err)
+		return false
+	}
+	// this log message is helpful for visibility on how far behind the block feeding is
+	e.logger.Info("L1 block provider live-monitoring", "currBlock", b.NumberU64(), "head", l1Head.NumberU64())
+	return b.NumberU64() >= l1Head.NumberU64()-1
```
StefanIliev545, line 69: Is the minus one here necessary? Perhaps a comment to explain it, since this would mean we consider a new fork to the parent to be "live" I suppose?

StefanIliev545 (changes requested):

go/ethadapter/blockprovider.go
StefanIliev545: Perhaps renaming it to `isLatest` would also be good in order to be consistent with the naming inside the enclave. I missed initially that it was the same thing when reviewing

go/ethadapter/blockprovider.go
BedrockSquirrel: Yeah fair, was just to give a bit of wiggle-room on a pretty wooly concept but actually I'll put it back so it just checks that the block == headBlock. (Also added the isLatest flag to the log message so if any weird issues creep up with forks and stuff then we'll see it)

go/ethadapter/blockprovider.go
BedrockSquirrel: Good shout

StefanIliev545 (approved):
lgtm

PR #1154, opened 2023-03-14
Add upgrade option to node script

BedrockSquirrel:
### Why this change is needed

We replaced docker compose with go scripts that manage the docker containers.

The main script just starts the containers but we need to be able to upgrade existing nodes as well.

It became obvious as well that some metadata about the network is required when restarting. So now we persist network config data in a json file on the node VMs to be read when restarting/upgrading.

- add "nodeAction" argument to the main script so you can call "start" or "upgrade" (and in the future we may add things like "stop", "status")
- add network config json file to node VMs so relevant config is available for node arguments when upgrading them

### PR checks pre-merging

### DIFF
--- a/.github/workflows/manual-upgrade-testnet-l2.yml
+++ b/.github/workflows/manual-upgrade-testnet-l2.yml
@@ -154,7 +154,7 @@ jobs:
               -host_public_p2p_addr=obscuronode-${{ matrix.host_id }}-${{needs.build.outputs.RESOURCE_TESTNET_NAME}}-${{ github.event.inputs.github_deploy_number }}.uksouth.cloudapp.azure.com:10000 \
               -host_p2p_port=10000 \
               -enclave_docker_image=${{needs.build.outputs.L2_ENCLAVE_DOCKER_BUILD_TAG}} \
-              -host_docker_image=${{needs.build.outputs.L2_HOST_DOCKER_BUILD_TAG}}
+              -host_docker_image=${{needs.build.outputs.L2_HOST_DOCKER_BUILD_TAG}} \
               upgrade'
 
   check-obscuro-is-healthy:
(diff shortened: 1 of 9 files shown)


.github/workflows/manual-upgrade-testnet-l2.yml
```diff
@@ -109,11 +141,21 @@ jobs:
...
-              docker compose up host enclave
-            '
+              git clone --depth 1 -b ${{ steps.extract_branch.outputs.branch }} https://github.com/obscuronet/go-obscuro.git /home/obscuro/go-obscuro \
+            && cd /home/obscuro/go-obscuro/ \
+            && sudo go run /home/obscuro/go-obscuro/go/node/cmd  \
+              -is_genesis=${{ matrix.is_genesis }} \
+              -node_type=${{ matrix.node_type }} \
+              -is_sgx_enabled=true \
+              -host_id=${{ secrets[matrix.node_pk_addr] }} \
+              -l1_host=${{needs.build.outputs.L1_HOST}} \
```
StefanIliev545, line 151: Is port needed here? The L2 deployment failure on the initial setup phase was failing because the default was different than what was deployed so I'm guessing this should also be wired?

BedrockSquirrel: Ah interesting. It's not currently being set on the `manual-deploy-testnet-l2.yml` script so I think the default works for these nodes.

Here's a clipboard diff between the deploy and the upgrade scripts:
Screenshot 2023-03-14 at 12 32 00

StefanIliev545: Are the contract addresses injected from somewhere? As we most definitely would need them on an upgrade restart

BedrockSquirrel: Yeah that's what the json thing is in this pr. We persist the contract addresses on the VM at deploy time and read them for restarts

PR #1186, opened 2023-03-24
Wipe and re-clone obscuro code on testnet node upgrade

BedrockSquirrel:
### Why this change is needed

Upgrade needs to update the code to the specified branch so it should delete and reclone the code (could pull and potentially switch branch but this seems simpler given we've got depth 1 clone.

Delete obscuro code from VM before cloning on upgrade script.

Move the network config file to not be in the same dir as the checked out code by default.

### PR checks pre-merging


StefanIliev545 (approved):
lgtm

PR #1195, opened 2023-03-30
Use docker mount for enclave persistence

BedrockSquirrel:
### Why this change is needed

After upgrading the enclave couldn't find the persisted EDB credentials and was trying to reinitialise the database (which fails).

Eventually realised this is because the creds are just persisted into the enclave container which is discarded on upgrade.

This change mirrors the host persistence, a persistent docker mount on the VM that is made available into the enclave docker container.

Add enclave persistence mount and configure it into the enclave docker container and the SGX enclave within that.

### PR checks pre-merging


tudor-malene (approved):
lgtm (if it works)

BedrockSquirrel:
Dev testnet came up on this branch so looks like we're all good

PR #1205, opened 2023-04-04
Updates from second pass through reviewing todos

moraygrieve:
### Why this change is needed

Updates comments for todo's based on review


moraygrieve:
Looks like commits to generated files should not have been made?

PR #1214, opened 2023-04-06
Adding debug_TraceTransaction endpoint and funcionality

otherview:
### Why this change is needed

Part of https://github.com/obscuronet/obscuro-internal/issues/971
Enables the debug handler.
Implements the debugTransaction.
Allows for 4byte, call, prestate and tracer tracers - Js tracers are not registered (easily added if needed).

Copied debug_traceTransaction and tracers from geth.
Added a GetChainStateAtTransaction that returns the chain after executing N transactions in a block.
Hooked the Host to Enclave RPC communication
Added a new debug namespace
Moved the ConvertToEthHeader to common package so it can be reused outside the evm package

### PR checks pre-merging


BedrockSquirrel (approved):
LGTM, there's some hideous looking code blocks but it's copied from geth with a comment so think we can allow it haha. And as you rightly pointed out this code is fairly low risk at the moment as it doesn't touch the main flows through the code.

We shouldn't gloss over the fact that this provides a privacy backdoor for testing and any enclave with it enabled won't be able to join the production network.

PR #1216, opened 2023-04-06
Reconnect client on the block provider

otherview:
### Why this change is needed

https://github.com/obscuronet/obscuro-internal/issues/1582

### PR checks pre-merging

### DIFF
--- a/go/ethadapter/blockprovider.go
+++ b/go/ethadapter/blockprovider.go
@@ -35,8 +35,9 @@ func NewEthBlockProvider(ethClient EthClient, logger gethlog.Logger) *EthBlockPr
 
 // EthBlockProvider streams blocks from the ethereum L1 client in the order expected by the enclave (consecutive, canonical blocks)
 type EthBlockProvider struct {
-	ethClient EthClient
-	logger    gethlog.Logger
+	ethClient         EthClient
+	logger            gethlog.Logger
+	lastBlockReceived time.Time
 }
 
 // StartStreamingFromHash will look up the hash block, find the appropriate height (LCA if there have been forks) and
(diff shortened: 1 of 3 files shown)


go/ethadapter/blockprovider.go
```diff
@@ -207,12 +216,31 @@ func (e *EthBlockProvider) awaitNewBlock(ctx context.Context) (*types.Header, er
 			if errors.Is(err, ErrSubscriptionNotSupported) {
 				return nil, err
 			}
-			e.logger.Error("L1 block monitoring error", log.ErrKey, err)
+			e.logger.Error("L1 block monitoring error - Restarting L1 block Monitoring...", log.ErrKey, err)
+
+			if !e.ethClient.Alive() {
```
BedrockSquirrel, line 221: Maybe worth moving this block to be an e.ensureClientConnected() method or something since it's duplicated below?

integration/networktest/log.go
```diff
@@ -5,7 +5,7 @@ import (
 	"github.com/obscuronet/go-obscuro/integration/common/testlog"
 )
 
-// EnsureTestLogsSetUp calls Setup iff it hasn't already been called (some tests run tests within themselves, we don't want
+// EnsureTestLogsSetUp calls Setup if it hasn't already been called (some tests run tests within themselves, we don't want
```
BedrockSquirrel, line 8: That wasn't a typo, "iff" is shorthand for "if and only if" but tbf we should probably avoid that if it's not widely used so I don't mind the edit.

BedrockSquirrel (approved):
Seems reasonable to me, minor suggestion

PR #1227, opened 2023-04-13
Refactored the l2 chain code. Added sequencer and validator separations and components to enable them

StefanIliev545:
### Why this change is needed

### PR checks pre-merging


otherview (commented):
image

tudor-malene:
> 
> 
> ### PR checks pre-merging
> Please indicate below by ticking the checkbox that you have read and performed the required [PR checks](https://github.com/obscuronet/obscuro-internal/blob/main/dev-ops-docs/dev-pr-checks.md)
> 

description please. It will be useful in the future as well

go/enclave/actors/interfaces.go
```diff
@@ -0,0 +1,28 @@
...
+	CreateBatch(*common.L1Block) (*core.Batch, error)
+	CreateRollup() (*common.ExtRollup, error)
+	IsReady() bool
+
+	ObscuroActor
+}
+
+type ObsValidator interface {
+	ValidateAndStoreBatch(*core.Batch) error
+	ReceiveBlock(*common.BlockAndReceipts, bool) (*components.BlockIngestionType, error)
```
tudor-malene, line 24: isn't this included in ObscuroActor?

tudor-malene (changes requested):
I like the premise of the PR. 
I'll try to reason a bit from a high level.

1. When you initialise the enclave, based on something you determine your role (validator or sequencer). (Let's say that's a config for now, but it needs to be something else)

2. Based on your role, you initialise the right service.

3. Ideally this service would have a single interface. When something is not supported like "CreateBatch" on a validator, it should just panic.
By having a single interface, you can keep the code using it more simple.
Since the create batch, create rollup, etc are driven from the host, it should e ok, because the host will know what role it has.

go/common/rpc/converters.go
```diff
@@ -31,21 +31,28 @@ func FromAttestationReportMsg(msg *generated.AttestationReportMsg) *common.Attes
...
-	return generated.BlockSubmissionResponseMsg{
+	msg := &generated.BlockSubmissionResponseMsg{
 		ProducedBatch:           &producedBatchMsg,
 		ProducedRollup:          &producedRollupMsg,
 		SubscribedLogs:          subscribedLogBytes,
 		ProducedSecretResponses: ToSecretRespMsg(response.ProducedSecretResponses),
-	}, nil
+	}
+
+	if response.SubscribedLogs != nil {
```
otherview, line 48: Should this marshelling be the first thing done ?

go/common/enclave.go
```diff
@@ -109,6 +109,8 @@ type Enclave interface {
 
 	// DebugTraceTransaction returns the trace of a transaction
 	DebugTraceTransaction(hash gethcommon.Hash, config *tracers.TraceConfig) (json.RawMessage, error)
+
+	StreamBatches() chan *ExtBatch
```
otherview, line 113: Can a comment be added to this method ?

go/enclave/actors/interfaces.go
```diff
@@ -0,0 +1,28 @@
+package actors
```
otherview, line 1: Can a README be added to this package ? Small just explaining the goal of it

go/enclave/actors/interfaces.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"github.com/obscuronet/go-obscuro/go/common"
+	"github.com/obscuronet/go-obscuro/go/enclave/components"
+	"github.com/obscuronet/go-obscuro/go/enclave/core"
+)
+
+type ObscuroActor interface {
```
otherview, line 9: Can comments be added to the interfaces ?

go/enclave/actors/sequencer.go
```diff
@@ -0,0 +1,307 @@
...
+		hostID:            hostID,
+		chainConfig:       chainConfig,
+		enclavePrivateKey: enclavePrivateKey,
+		mempool:           mempool,
+		storage:           storage,
+		encryption:        encryption,
+	}
+}
+
+func (s *sequencer) IsReady() bool {
```
otherview, line 71: I think I know why this returns false, but can a comment be added to make it clear ?

otherview (changes requested):
Looks good, I like the seperation of concerns on the actors.

Can comments be added to the Public methods, interfaces and high level steps ? This would greatly improve the review and avoid missing some potential issues.

go/enclave/actors/sequencer.go
StefanIliev545: just needs to be finished/removed

go/config/host_config.go
```diff
@@ -213,5 +227,7 @@ func DefaultHostParsedConfig() *HostInputConfig {
 		MetricsHTTPPort:           14000,
 		UseInMemoryDB:             true,
 		DebugNamespaceEnabled:     false,
+		BatchInterval:             1 * time.Second,
```
BedrockSquirrel, line 230: I plucked these default intervals out of thin air and while they're configurable I don't think I set them anywhere. So worth a quick discussion on whether we're happy with these values? I guess we were typically on 2 second batches before when it was driven by the L1 rate.

go/enclave/components/consumer.go
BedrockSquirrel, line 1: I'd call these files like blockconsumer and batchproducer probably, easier to understand at first glance maybe.

go/enclave/components/consumer_test.go
```diff
@@ -1,28 +1,20 @@
...
 func TestInvalidBlocksAreRejected(t *testing.T) {
-	// There are no tests of acceptance of valid chains of blocks. This is because the logic to generate a valid block
-	// is non-trivial.
-	genesisJSON, err := core.DefaultGenesisBlock().MarshalJSON()
-	if err != nil {
-		t.Errorf("could not parse genesis JSON: %v", err)
-	}
-	logger := log.New(log.DeployerCmp, int(gethlog.LvlDebug), log.SysOut)
-	chain := ObscuroChain{l1Blockchain: NewL1Blockchain(genesisJSON, logger)}
+	t.Skipf("skipping test that relied on l1blockchain previously which hasn't been used in a while")
```
BedrockSquirrel, line 14: worth deleting the test or you want to leave it so we can replace it with a similar one?

go/enclave/components/interfaces.go
```diff
@@ -0,0 +1,139 @@
...
+// ComputedBatch - a structure representing the result of a batch
+// computation where `Batch` is the newly computed batch and `Receipts`
+// are the receipts for the executed transactions inside this batch.
+// The `Commit` function allows for committing the stateDB resulting from
+// the computation of the batch. One might not want to commit in case the
+// resulting batch differs than what is being validated for example.
+type ComputedBatch struct {
+	Batch    *core.Batch
+	Receipts types.Receipts
+	Commit   func(bool) (gethcommon.Hash, error)
```
BedrockSquirrel, line 53: We should include in one of the comments here what this bool parameter is. I only looked it up an hour ago and I've already forgotten :)

BedrockSquirrel (commented):
Got about 25% through and got to go for now.- looking good, minor stuff. Will go through the rest later/tomorrow.

go/host/container/cli_flags.go
```diff
@@ -64,5 +66,7 @@ func getFlagUsageMap() map[string]string {
 		useInMemoryDBName:            "Whether the host will use an in-memory DB rather than persist data",
 		levelDBPathName:              "Filepath for the levelDB persistence dir (can be empty if a throwaway file in /tmp/ is acceptable or if using InMemory DB)",
 		debugNamespaceEnabledName:    "Whether the debug names is enabled",
+		batchIntervalName:            "How many seconds between each batch",
```
BedrockSquirrel, line 69: If you want to use `ParseDuration` then should probably change these prompts to not mention "seconds", people can put any duration with units I guess.

go/host/host.go
```diff
@@ -863,6 +856,92 @@ func (h *host) handleBatches(encodedBatchMsg *common.EncodedBatchMsg) error {
...
+		select {
+		case resp, ok := <-streamChan:
+			if !ok {
+				stop()
+				h.logger.Warn("Batch streaming failed. Reconneting from latest received batch after 3 seconds")
+				time.Sleep(3 * time.Second)
+
+				if lastBatch != nil {
+					bHash := lastBatch.Hash()
+					streamChan, stop = h.EnclaveClient().StreamBatches(&bHash)
```
BedrockSquirrel, line 896: Tiny thing but can you switch this to `h.enclaveClient` instead of the public method, probably autocomplete by accident. Caught my eye cos it's different from the one two lines below so got me wondering why.

BedrockSquirrel (commented):
Few more comments

go/enclave/enclave.go
```diff
@@ -261,78 +285,138 @@ func (e *enclaveImpl) StopClient() error {
 	return nil // The enclave is local so there is no client to stop
 }
 
-// SubmitL1Block is used to update the enclave with an additional L1 block.
-func (e *enclaveImpl) SubmitL1Block(block types.Block, receipts types.Receipts, isLatest bool) (*common.BlockSubmissionResponse, error) {
-	if atomic.LoadInt32(e.stopInterrupt) == 1 {
-		return nil, nil //nolint:nilnil
+func (e *enclaveImpl) sendMissingMatches(fromHash *common.L2BatchHash, outChannel chan common.StreamBatchResponse) error {
```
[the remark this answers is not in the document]
StefanIliev545: This would find the batch in the fork and walk back the head until it reaches the same parent batch in both chains and then continue on.

go/enclave/l2chain/lite_l2_chain.go
```diff
@@ -0,0 +1,253 @@
...
+		logger:               logger,
+		GlobalGasCap:         5_000_000_000, // todo (#627) - make config
+		BaseFee:              gethcommon.Big0,
+		sequencerID:          sequencerID,
+		genesis:              genesis,
+		Registry:             registry,
+	}
+}
+
+type ChainInterface interface {
```
tudor-malene: shouldn't both the Lite and the non-lite L2 chains implement this interface?

go/enclave/l2chain/l2_chain.go
```diff
@@ -1,162 +1,82 @@
...
-		return nil, errors.New("head batch height is in the past compared to requested batch")
-	}
-
-	for batch.Number().Cmp(headBatch.Number()) != 0 {
-		batch, _ = oc.storage.FetchBatchByHeight(batch.NumberU64() + 1)
-		batches = append(batches, batch)
-	}
-
-	return batches, nil
+type ChainInterface interface {
```
tudor-malene, line 74: let's move this to an "intefraces" file and add a nice comment on it

go/enclave/l2chain/l2_chain.go
```diff
@@ -1,162 +1,82 @@
...
-
-	gethcommon "github.com/ethereum/go-ethereum/common"
-	gethcore "github.com/ethereum/go-ethereum/core"
-	gethlog "github.com/ethereum/go-ethereum/log"
-	gethrpc "github.com/ethereum/go-ethereum/rpc"
 )
 
-// ObscuroChain represents the canonical L2 chain, and manages the state.
 type ObscuroChain struct {
 	hostID      gethcommon.Address
```
tudor-malene, line 31: Some fields are no longer used. Like the "hostId", "nodeType" and others.
Let's get rid of them

tudor-malene (changes requested):
pushing a partial review.

go/enclave/components/interfaces.go
```diff
@@ -0,0 +1,138 @@
...
+	// (resulting in rewinding of one or more blocks that we had previously considered canonical)
+	Fork bool
+
+	// PreGenesis is true if there is no stored L1 head block.
+	// (L1 head is only stored when there is an L2 state to associate it with. Soon we will start consuming from the
+	// genesis block and then, we should only see one block ingested in a 'PreGenesis' state)
+	PreGenesis bool
+}
+
+type BlockConsumer interface {
```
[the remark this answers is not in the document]
tudor-malene: Let's prefix everything that has "Block" with "L1", to make it extra clear what we're talking about
"L1BlockXXX"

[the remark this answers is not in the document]
tudor-malene: "L1BlocksProcessor" and the main method "Process"

go/common/tracers/native/4byte.go
```diff
@@ -41,15 +41,15 @@ func init() {
...
-//   > debug.traceTransaction( "0x214e597e35da083692f5386141e69f47e973b2c56e7a8073b1ea08fd7571e9de", {tracer: "4byteTracer"})
-//   {
-//     0x27dc297e-128: 1,
-//     0x38cc4831-0: 2,
-//     0x524f3889-96: 1,
-//     0xadf59f99-288: 1,
-//     0xc281d19e-0: 1
-//   }
 //
+//	> debug.traceTransaction( "0x214e597e35da083692f5386141e69f47e973b2c56e7a8073b1ea08fd7571e9de", {tracer: "4byteTracer"})
```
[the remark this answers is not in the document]
StefanIliev545: gofumpt doesn't like those files.

go/enclave/services/interfaces.go
```diff
@@ -0,0 +1,42 @@
...
+	// SubmitTransaction - L2 obscuro transactions need to be passed here. Sequencers
+	// will put them in the mempool while validators might put them in a queue and monitor
+	// for censorship.
+	SubmitTransaction(*common.L2Tx) error
+}
+
+type Sequencer interface {
+	// CreateBatch - creates a new head batch for the parameter block or if there is no block
+	// provided, for the latest known L1 head block.
+	CreateBatch(*common.L1Block) (*core.Batch, error)
```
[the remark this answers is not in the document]
StefanIliev545: I'm still contemplating if the host can select what l1 head it wants to create a batch for (and rollup) since it should always know where the l1 chain sits at

go/enclave/components/sigverifier.go
```diff
@@ -0,0 +1,45 @@
+package components
```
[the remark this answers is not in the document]
StefanIliev545: I'm not sure, but it kind of holds state and needs access to the storage because of the `FetchAttestedKey`. Was thinking it can be extended further to remove the process network secrets stuff from the enclave.

go/enclave/services/sequencer.go
```diff
@@ -0,0 +1,329 @@
...
+	return batch, nil
+}
+
+func (s *sequencer) createNewHeadBatch(block *common.L1Block) (*core.Batch, error) {
+	headBatch, err := s.batchRegistry.GetHeadBatch()
+	if err != nil {
+		return nil, err
+	}
+
+	// We find the ancestral batch for the block as it might be different
```
tudor-malene, line 130: I have a feeling this logic is unnecessarily complex.
I'll try to restate the problem.

Problem: I want to add a batch to the l2 chain. 

Logic:
1. If the current l2 head points to an ancestor of the current l1 head, then I just carry on.

2. if the current l2 head points to a fork of the l1 head, then I go to the l1 block where the l1 fork happened. 
Try to find batches in the db for the current canonical fork ( maybe there are some already). If there are, then carry on from where they left off, but making sure to duplicate batches up to the maximum height already created.


Based on the above description. I'd say that a very important parameter of the sequencer is the "MaxL2Height". Basically if you're creating a batch with a height of less than that, you have to make sure to duplicate the already existing batch at that height.

tudor-malene (commented):
more partial comments

go/enclave/components/batch_registry.go
```diff
@@ -0,0 +1,238 @@
...
+	return br.storage.SetHeadBatchPointer(batch)
+}
+
+func (br *batchRegistry) updateBlockPointers(batch *core.Batch, receipts types.Receipts) error {
+	head, err := br.GetHeadBatchFor(batch.Header.L1Proof)
+
+	if err != nil && !errors.Is(err, errutil.ErrNotFound) {
+		return fmt.Errorf("unexpected error while getting head batch for block. Cause: %w", err)
+	} else if head != nil && batch.NumberU64() > head.NumberU64() {
+		return fmt.Errorf("inappropriate update from previous head with height %d to new head with height %d for same l1 block", head.NumberU64(), batch.NumberU64())
```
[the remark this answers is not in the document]
StefanIliev545: Me too, I dunno what I'm doing here

go/enclave/services/README.md
```diff
@@ -0,0 +1,5 @@
+### The Service package
```
otherview, line 1: In golang it's common practice to avoid generic packages like services.
Instead, for example, nodetype could work, where the `sequencer` and the `validator` are implementations of the `node_type.go` interface. 

If another name is prefered that's good too, it's just going in the direction of this package is a unit of work and not a collection of units. Hope that makes sense :)

go/enclave/services/sequencer.go
```diff
@@ -0,0 +1,324 @@
...
+		hostID:            hostID,
+		chainConfig:       chainConfig,
+		enclavePrivateKey: enclavePrivateKey,
+		mempool:           mempool,
+		storage:           storage,
+		encryption:        encryption,
+	}
+}
+
+func (s *sequencer) CreateBatch() error {
```
otherview, line 78: Would be cool to have unit tests around this where we mock bad data, forks or other weird scenarios and see the error / expected batches come out.

go/enclave/services/sequencer.go
```diff
@@ -0,0 +1,324 @@
...
+// should only create batches and stateDBs but not commit them to the database,
+// this is the responsibility of the sequencer. Refactor the code so genesis state
+// won't be committed by the producer.
+func (s *sequencer) initGenesis(block *common.L1Block) error {
+	batch, msgBusTx, err := s.batchProducer.CreateGenesisState(block.Hash(), s.hostID, uint64(time.Now().Unix()))
+	if err != nil {
+		return err
+	}
+
+	if err := s.mempool.AddMempoolTx(msgBusTx); err != nil {
```
otherview, line 110: no need to declare `err`. ( following few of lines too)

go/host/rpc/enclaverpc/enclave_client.go
```diff
@@ -367,6 +378,62 @@ func (c *Client) DebugTraceTransaction(hash gethcommon.Hash, config *tracers.Tra
...
+func (c *Client) StreamBatches(from *common.L2BatchHash) (chan common.StreamBatchResponse, func()) {
+	batchChan := make(chan common.StreamBatchResponse, 10)
+	cancelCtx, cancel := context.WithCancel(context.Background())
+
+	request := &generated.StreamBatchesRequest{}
+	if from != nil {
+		request.KnownHead = from.Bytes()
+	}
+
+	stream, err := c.protoClient.StreamBatches(cancelCtx, request)
```
otherview, line 390: The error typically gets returned

go/host/rpc/enclaverpc/enclave_client.go
```diff
@@ -367,6 +378,62 @@ func (c *Client) DebugTraceTransaction(hash gethcommon.Hash, config *tracers.Tra
...
+		}
+
+		cancel()
+		close(batchChan)
+	}
+
+	go func() {
+		defer stopIt()
+		for {
+			if stop {
```
otherview, line 412: Probably select is a better choice instead of the for loop

otherview (commented):
I think it's overall ok.

A few overall points:
- It looks soooo much better in terms of understanding the flow of data process
- Services/utils/components packages are not typically that used in go (the services/nodetype comment is a good example of what I mean)
- All comms are Host->Enclave direction and it seems like the Stream Batches breaks this convetion

go/host/host.go
```diff
@@ -264,13 +272,16 @@ func (h *host) Unsubscribe(id rpc.ID) {
 }
 
 func (h *host) Stop() {
+	h.logger.Info("Host received a stop command. Attempting shutdown...")
+	h.interrupter.Close()
+	h.shutdownGroup.Wait()
```
BedrockSquirrel, line 277: When you merge in main there's a utility method called `async.WaitTimeout()` that will wait for a WaitGroup with a timeout.

I think if we use that here with a clear error message when it does timeout then it'll help with some of the shutdown hanging issues on e2e tests etc.

When I do the HA work I'm hoping I can figure out what's hanging as a look more closely at the lifecycles of that stuff.

go/enclave/components/batch_registry.go
```diff
@@ -0,0 +1,238 @@
...
+
+func (br *batchRegistry) GetHeadBatchFor(blockHash common.L1BlockHash) (*core.Batch, error) {
+	return br.storage.FetchHeadBatchForBlock(blockHash)
+}
+
+func (br *batchRegistry) GetBatch(batchHash common.L2BatchHash) (*core.Batch, error) {
+	return br.storage.FetchBatch(batchHash)
+}
+
+func (br *batchRegistry) Subscribe() chan *core.Batch {
```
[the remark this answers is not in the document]
StefanIliev545: There should only be a single subscription (for simplicity) to the enclave and the assumption is that if a new one is opened then the host dropped the connection to the old one so it replaces it.

go/common/enclave.go
```diff
@@ -106,11 +106,20 @@ type Enclave interface {
 	// HealthCheck returns whether the enclave is in a healthy state
 	HealthCheck() (bool, error)
 
-	GenerateRollup() (*ExtRollup, error)
+	// CreateBatch - creates a new head batch extending the previous one for the latest known L1 head if the node is
+	// a sequencer. Will panic otherwise.
```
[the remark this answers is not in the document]
StefanIliev545: Not really, the host should know what it is calling. If its starts calling sequencer code on a validator then something somewhere is corrupted.

go/enclave/services/validator.go
```diff
@@ -0,0 +1,174 @@
...
+		Randomness:   incomingBatch.Header.MixDigest,
+		Creator:      incomingBatch.Header.Agg,
+		ChainConfig:  val.chainConfig,
+	})
+	if err != nil {
+		return fmt.Errorf("failed computing batch. Cause: %w", err)
+	}
+
+	if !bytes.Equal(cb.Batch.Hash().Bytes(), incomingBatch.Hash().Bytes()) {
+		return fmt.Errorf("batch is in invalid state. Incoming hash: %s  Computed hash: %s", incomingBatch.Hash().Hex(), cb.Batch.Hash().Hex())
```
tudor-malene, line 107: pls add todo here and ticket to generate the challenge at this point, and return it

go/enclave/components/interfaces.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+type RollupConsumer interface {
+	// ProcessL1Block - extracts the rollups from the block's transactions
+	// and verifies their integrity, saving and processing any batches that have
+	// not been seenp previously.
+	ProcessL1Block(b *common.BlockAndReceipts) ([]*core.Rollup, error)
+}
+
+// TODO - find a better place for the following stuff
```
tudor-malene, line 129: why not move them to the batch_producer

go/enclave/components/rollup_producer.go
```diff
@@ -0,0 +1,155 @@
...
+	TransactionBlobCrypto crypto.TransactionBlobCrypto
+
+	ObscuroChainID  int64
+	EthereumChainID int64
+
+	logger gethlog.Logger
+
+	storage db.Storage
+
+	registry      BatchRegistry
```
tudor-malene, line 36: the naming of this variable is inconsistent in the places where it's used.
Sometimes `registry` or `batchRegistry`.
Let's settle on `batchRegistry` everywhere

go/enclave/components/rollup_producer.go
```diff
@@ -0,0 +1,155 @@
...
+	blockConsumer L1BlockProcessor
```
tudor-malene, line 37: same for this.

go/enclave/components/rollup_producer.go
```diff
@@ -0,0 +1,155 @@
+package components
+
+// todo (@stefan) - once the cross chain messages based bridge is implemented remove this completely
```
tudor-malene, line 3: what's up with this todo?
what needs to be removed?

go/enclave/components/rollup_producer.go
```diff
@@ -0,0 +1,155 @@
...
+// fetchLatestRollup - Will pull the latest rollup based on the current head block from the database or return null
+func (re *rollupProducer) fetchLatestRollup() (*core.Rollup, error) {
+	b, err := re.blockConsumer.GetHead()
+	if err != nil {
+		return nil, err
+	}
+	return re.getLatestRollupBeforeBlock(b)
+}
+
+func (re *rollupProducer) CreateRollup() (*core.Rollup, error) {
```
tudor-malene, line 69: let's add a comment here with the problem we discussed around reorgs.
That is is the responsibility of the host to wait until it knows that the rollup it has contains only "final" batches that it pushes it to the l1

go/enclave/components/rollup_producer.go
```diff
@@ -0,0 +1,155 @@
...
+	if batches[len(batches)-1].Header.Hash() == hash {
+		return nil, fmt.Errorf("current head batch matches the rollup head bash")
+	}
+
+	newRollup := createNextRollup(rollup, batches)
+	return newRollup, nil
+}
+
+// getLatestRollupBeforeBlock - Given a block, returns the latest rollup in the canonical chain for that block (excluding those in the block itself).
+func (re *rollupProducer) getLatestRollupBeforeBlock(block *common.L1Block) (*core.Rollup, error) {
```
tudor-malene, line 98: this method is duplicated in the rollup_consumer

tudor-malene (changes requested):
partial review

go/enclave/components/block_processor.go
```diff
@@ -0,0 +1,110 @@
...
+		err = bc.crossChainProcessors.Remote.StoreCrossChainMessages(br.Block, *br.Receipts)
+		if err != nil {
+			return nil, errors.New("failed to process cross chain messages")
+		}
+	}
+
+	return ingestion, nil
+}
+
+func (bc *l1BlockProcessor) tryAndInsertBlock(br *common.BlockAndReceipts, isLatest bool) (*BlockIngestionType, error) {
```
tudor-malene, line 47: let's get rid of this method and move all this logic to "Process".

go/enclave/components/block_processor.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/obscuronet/go-obscuro/go/enclave/db"
+)
+
+type l1BlockProcessor struct {
+	storage              db.Storage
+	logger               gethlog.Logger
+	crossChainProcessors *crosschain.Processors
+}
+
+func NewBlockConsumer(storage db.Storage, cc *crosschain.Processors, logger gethlog.Logger) L1BlockProcessor {
```
tudor-malene, line 22: there's a naming discrepancy here. 
blockProcessor/Consumer

go/enclave/components/block_processor.go
```diff
@@ -0,0 +1,110 @@
...
+	return &l1BlockProcessor{
+		storage:              storage,
+		logger:               logger,
+		crossChainProcessors: cc,
+	}
+}
+
+func (bc *l1BlockProcessor) Process(br *common.BlockAndReceipts, isLatest bool) (*BlockIngestionType, error) {
```
tudor-malene, line 30: if it's blockProcessor then the variable should be "bp"

tudor-malene (commented):
more comments

tudor-malene (commented):
few more comments. Mostly for @otherview

go/enclave/l2chain/interfaces.go
```diff
@@ -0,0 +1,32 @@
...
+	gethrpc "github.com/ethereum/go-ethereum/rpc"
+	"github.com/obscuronet/go-obscuro/go/common/gethapi"
+	"github.com/obscuronet/go-obscuro/go/enclave/core"
+)
+
+// ObscuroChain - the interface that provides the data access layer to the obscuro l2.
+// Operations here should be read only.
+type ObscuroChain interface {
+	// GetBalance - Returns the balance of an address given the standard provided BlockNumber which in our case is batch number.
+	// Note this method also returns the address we should encrypt the balance for, unlike `GetBalanceAtBlock`
```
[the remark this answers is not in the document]
otherview: As far as I can tell, it's because the GetBalance is a User Method and GetBalanceAtBlock isn't.

These names are confusing. The GetBalance is a "special" function that does logic to figure out if the address is a contract or EOA so it can decide which keys it should encrypt with. 
The GetBalance calls the GetBalanceAtBlock to actually get the balance of the address.

Should the methods' names be distinguished ? One JUST fetches the balance and the other actually implements the rules required to safely return the balance to the user.

go/enclave/l2chain/interfaces.go
```diff
@@ -0,0 +1,32 @@
...
+	GetBalance(accountAddress gethcommon.Address, blockNumber *gethrpc.BlockNumber) (*gethcommon.Address, *hexutil.Big, error)
+
+	// GetBalanceAtBlock - will return the balance of a specific address at the specific given block number (batch number).
+	GetBalanceAtBlock(accountAddr gethcommon.Address, blockNumber *gethrpc.BlockNumber) (*hexutil.Big, error)
+
+	// ObsCall - The interface for executing eth_call RPC commands against obscuro.
+	ObsCall(apiArgs *gethapi.TransactionArgs, blockNumber *gethrpc.BlockNumber) (*gethcore.ExecutionResult, error)
+
+	// ObsCallAtBlock - Execute eth_call RPC against obscuro for a specific block (batch) number.
```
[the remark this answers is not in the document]
otherview: Same as above I believe. The `ObsCallAtBlock` is the actual method that runs the obs call. And the `ObsCall` seems to do some very small work of logging and error handling.

Removing the `ObsCall` could make things simpler, but we would lose that logging and impact the estimate gas. ( In  `ObsCall` we return err if `result.Failed()` in `ObsCall` we don't, and that's on purpose bc of the magic estimation function of geth).

Maybe making the methods names more explicit would make it clearer ?

otherview:
> few more comments. Mostly for @otherview

Perhaps a good time to distinguish what are User Focused methods and  System / Internal methods are ?

go/enclave/components/interfaces.go
StefanIliev545: wasn't clear at the time what the best place would be

go/common/enclave.go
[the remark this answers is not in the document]
otherview: Weird that that host can trigger panics on the enclave on-demand, no ?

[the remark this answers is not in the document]
StefanIliev545: From a security POV?

go/enclave/components/batch_producer.go
```diff
@@ -0,0 +1,202 @@
...
+		return nil, err
+	}
+
+	ccSuccessfulTxs, ccReceipts, err := bp.processTransactions(batch, len(successfulTxs), crossChainTransactions, stateDB, context.ChainConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := bp.verifyInboundCrossChainTransactions(crossChainTransactions, ccSuccessfulTxs, ccReceipts); err != nil {
+		return nil, fmt.Errorf("batch computation failed due to cross chain messages. Cause: %w", err)
```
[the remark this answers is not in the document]
StefanIliev545: This shouldn't be possible, we generate the synthetic transactions out of the rlp encoded events. It's just a data transfer so the data inside the message makes no difference whatsoever since it does not trigger any arbitary executions. Perhaps if the data is insanely big (but someone has to pay for this on l1) and l1 allows for a transaction with more size than we do. I need to think about it more, but in general there shouldn't be anything DoS worthy here.

go/enclave/components/batch_producer.go
```diff
@@ -0,0 +1,202 @@
...
+func (bp *batchProducer) verifyInboundCrossChainTransactions(transactions types.Transactions, executedTxs types.Transactions, receipts types.Receipts) error {
+	if transactions.Len() != executedTxs.Len() {
+		return fmt.Errorf("some synthetic transactions have not been executed")
+	}
+
+	for _, rec := range receipts {
+		if rec.Status == 1 {
+			continue
+		}
+		return fmt.Errorf("found a failed receipt for a synthetic transaction")
```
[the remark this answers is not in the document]
StefanIliev545: its a synthetic transaction so tx hash would be meaningless since there is no record of the tx going in

go/enclave/components/batch_registry.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+func (br *batchRegistry) updateSubscribers(batch *core.Batch) {
+	br.subscriptionMutex.Lock()
+	subscriptionChan := br.subscription
+	br.subscriptionMutex.Unlock()
+
+	if subscriptionChan != nil {
+		*subscriptionChan <- batch
+	}
```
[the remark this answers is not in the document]
StefanIliev545: In case unsubscribe() makes br.subscription nil while trying to push. subscriptionChannel would always point to an actual channel

go/enclave/components/batch_registry.go
```diff
@@ -0,0 +1,238 @@
...
+
+	// TODO - we probably dont want to spawn a goroutine everytime
+	go br.updateSubscribers(batch)
+
+	return nil
+}
+
+func (br *batchRegistry) updateSubscribers(batch *core.Batch) {
+	br.subscriptionMutex.Lock()
```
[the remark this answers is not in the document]
StefanIliev545: Unsubscribe() can change the br.subscription to nil between pushing and checking for nil

go/common/enclave.go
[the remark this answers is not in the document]
tudor-malene: It's cool. The host can kill the enclave anytime it likes anyway.
This does not give it additional powers

go/enclave/rpc_server.go
```diff
@@ -251,6 +257,40 @@ func (s *RPCServer) DebugTraceTransaction(_ context.Context, req *generated.Debu
...
+		}
+
+		encoded, err := json.Marshal(batchResp)
+		if err != nil {
+			s.logger.Error("Error marshalling batch response", log.ErrKey, err)
+			return nil
+		}
+
+		if err := stream.Send(&generated.EncodedUpdateResponse{Batch: encoded}); err != nil {
+			s.logger.Error("Failed streaming batch back to client", log.ErrKey, err)
```
tudor-malene, line 284: logger.Info

go/enclave/components/batch_registry.go
```diff
@@ -0,0 +1,324 @@
...
+	"github.com/obscuronet/go-obscuro/go/enclave/core"
+	"github.com/obscuronet/go-obscuro/go/enclave/db"
+)
+
+type batchRegistryImpl struct {
+	storage db.Storage
+	logger  gethlog.Logger
+
+	batchSubscription *chan *core.Batch
+	eventSubscription *chan uint64
```
tudor-malene, line 26: commment

tudor-malene (approved):
lgtm
(after the few comments)

PR #1239, opened 2023-04-21
Updates after backlog review part three

moraygrieve:


tudor-malene (commented):
lgtm, after fixing the whitespaces to keep the linter happy

PR #1241, opened 2023-04-28
Remove error from proto message payload

otherview:
### Why this change is needed

https://github.com/obscuronet/obscuro-internal/issues/1644

Removes the internal message from the proto messages payload.

Having methods return `Type{ someInfoValues, error}, error` is often confusing. In internal methods any error should be surfaced using `Type{ someInfoValues }, error `. 

Connected with https://github.com/obscuronet/obscuro-internal/issues/1524 .

- Status endpoint now returns `StatusResponse{Status: int32(status)}, err`
- Status RPC client now returns `nil, err` if there was an issue retrieving the status.
- InitEnclave similar to above
- DebugEventLogRelevancy similar to above
- DebugTraceTransaction similar to above
- CreateRollup similar to above
- HealthCheck similar to above

### PR checks pre-merging


BedrockSquirrel (approved):
lgtm

PR #1242, opened 2023-04-28
Add wait group Wait() timeout function to util

BedrockSquirrel:
### Why this change is needed

Host is using a waitgroup for shutdown in Stefan's big PR and things keep hanging on shutdown, this feels like it might help but I didn't want to further bloat the PR with this refactor.

I found another async-related function and pulled it into the new util package as well.

### PR checks pre-merging

### DIFF
--- a/go/common/async/async.go
+++ b/go/common/async/async.go
@@ -3,28 +3,12 @@ package async
 import (
 	"fmt"
 	"sync"
-	"sync/atomic"
 	"time"
 )
 
 type (
 	ScheduledFunc func()
 )
-
-// ScheduleInterrupt runs the function after the delay and can be interrupted
-func ScheduleInterrupt(delay time.Duration, interrupt *int32, fun ScheduledFunc) {
-	ticker := time.NewTicker(delay)
-
-	go func() {
-		<-ticker.C
-		if atomic.LoadInt32(interrupt) == 1 {
-			return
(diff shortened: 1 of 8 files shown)


go/common/async/async.go
```diff
@@ -0,0 +1,53 @@
...
+	go func() {
+		<-ticker.C
+		if atomic.LoadInt32(interrupt) == 1 {
+			return
+		}
+
+		fun()
+		ticker.Stop()
+	}()
+}
```
otherview, line 27: I think you can delete this method because the interrupt it checked on the `ScheduledFunc`.

integration/ethereummock/node.go
```diff
@@ -351,7 +353,7 @@ func (m *Node) startMining() {
 
 			// Generate a random number, and wait for that number of ms. Equivalent to PoW
 			// Include all rollups received during this period.
-			common.ScheduleInterrupt(m.cfg.PowTime(), interrupt, func() {
+			async.ScheduleInterrupt(m.cfg.PowTime(), interrupt, func() {
```
otherview, line 356: This can be replaced by `Schedule` since the interrupt is checked inside the function

otherview (approved):
Good stuff ! I think the `ScheduleInterrupt` can be removed

PR #1251, opened 2023-05-08
Add InMemory Node to the node package

otherview:
### Why this change is needed

Creating in-memory versions of the obscuro node is somewhat messy right now.
The common practice is to directly create a node using the core struct, which is becoming larger and harder to maintain.

This PR cleans up the existing node package and allows to create in-memory instances of the Obscuro node.
Also changes the node runner test to use the node package instead of manually creating the structs.

- Adds the InMemory Node type to the Node package
- Cleans up the Node interface
- Noderunner now uses this structure

### PR checks pre-merging

### DIFF
--- a/go/node/in_memory_node.go
+++ b/go/node/in_memory_node.go
@@ -15,10 +15,10 @@ type InMemNode struct {
 	host    *hostcontainer.HostContainer
 }
 
-func NewInMemNode(cfg *Config) (*InMemNode, error) {
+func NewInMemNode(cfg *Config) *InMemNode {
 	return &InMemNode{
 		cfg: cfg,
-	}, nil // todo: add config validation
+	}
 }
 
 func (d *InMemNode) Start() error {
--- a/integration/noderunner/noderunner_test.go
+++ b/integration/noderunner/noderunner_test.go
@@ -3,7 +3,6 @@ package noderunner
 import (
 	"encoding/hex"
 	"fmt"
-	gethcommon "github.com/ethereum/go-ethereum/common"
 	"net/http"
 	"testing"
 	"time"
@@ -17,6 +16,7 @@ import (
 	"github.com/obscuronet/go-obscuro/integration/common/testlog"
 	"github.com/obscuronet/go-obscuro/integration/eth2network"
 
+	gethcommon "github.com/ethereum/go-ethereum/common"
 	gethlog "github.com/ethereum/go-ethereum/log"
 )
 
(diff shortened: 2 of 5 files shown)


go/node/in_memory_node.go
```diff
@@ -0,0 +1,90 @@
...
+	hostcontainer "github.com/obscuronet/go-obscuro/go/host/container"
+)
+
+type InMemNode struct {
+	cfg     *Config
+	enclave *enclavecontainer.EnclaveContainer
+	host    *hostcontainer.HostContainer
+}
+
+func NewInMemNode(cfg *Config) (Node, error) {
```
BedrockSquirrel, line 18: We should return the `*InMemNode` here imo rather than `Node`, "Accept interfaces, return structs"

go/node/node.go
```diff
@@ -0,0 +1,7 @@
+package node
+
+type Node interface {
```
BedrockSquirrel, line 3: I don't think this interface is needed, is it used anywhere?

integration/noderunner/noderunner_test.go
```diff
@@ -86,20 +75,15 @@ func TestCanStartStandaloneObscuroHostAndEnclave(t *testing.T) {
 		panic(err)
 	}
 
-	go func() {
-		err := enclavecontainer.NewEnclaveContainerFromConfig(enclaveConfig).Start()
-		if err != nil {
-			panic(err)
-		}
-	}()
+	inMemNode, _ := node.NewInMemNode(nodeCfg)
```
BedrockSquirrel, line 78: We should check the error here, else risk nil pointering below and making it harder to debug in future.

BedrockSquirrel (changes requested):
minor comments, seems good generally

go/node/in_memory_node.go
otherview: Good point !

go/node/node.go
otherview: It's a good idea to make sure that Node implementations follow it. 
I've changed the noderunner to use the interface and added a todo to have the testrunner run the docker instance tests.

On a side note, we're slowing starting to build a good suite of long-running tests that we should probably run every night?

integration/noderunner/noderunner_test.go
BedrockSquirrel: I think if the method returns an error type the caller should assume it might return an error. Relying on devs checking the usages when they update it seems like a pointless risk.

If you don't think it needs it then scrap the error return maybe?

go/node/node.go
BedrockSquirrel: Fair enough. Not sure what you mean about the long-running tests, I'll ping on discord.

integration/noderunner/noderunner_test.go
otherview: Yeah makes sense 👍

BedrockSquirrel (approved):
LGTM

PR #1254, opened 2023-05-09
InMem Node used in the socket simulation.

otherview:
### Why this change is needed

- Socket simulation should use the in memory nodes instead of directly calling the structs
- Socket simulation now also uses actual addresses and PKs instead of using hacks

### PR checks pre-merging


StefanIliev545:
I think this breaks full network sim logging - do a full run and things get overwritten and lost from the enclave. Seems to be because we create a new logger everywhere post this PR, while previously the testlog.Logger() was shared

otherview:
Good catch ! I've created a task to track that bug.

PR #1255, opened 2023-05-10
Testnet deploy: install gcc and allow time for VMs to initialise

BedrockSquirrel:
### Why this change is needed

Dev testnet deployments are consistently failing with:
```
# github.com/ethereum/go-ethereum/crypto/secp256k1
cgo: C compiler \"gcc\" not found: exec: \"gcc\": executable file not found in $PATH
# github.com/ethereum/go-ethereum/rpc
cgo: C compiler \"gcc\" not found: exec: \"gcc\": executable file not found in $PATH
# github.com/edgelesssys/ego/internal/attestation
cgo: C compiler \"gcc\" not found: exec: \"gcc\": executable file not found in $PATH
# github.com/mattn/go-sqlite3
cgo: C compiler \"gcc\" not found: exec: \"gcc\": executable file not found in $PATH
```

I think this is because the node package which is called directly (docker launcher) now has a transitive dependency on sqlite and edgeless because of the In-Mem node we added to it.

- Install gcc as part of the prep for the launch
- make sure both nodes sleep for 1minute (this is an attempt to fix an issue where key ubuntu files were missing on the VM, maybe too soon after creation?)

### PR checks pre-merging

### DIFF
--- a/.github/workflows/manual-deploy-testnet-l2.yml
+++ b/.github/workflows/manual-deploy-testnet-l2.yml
@@ -195,7 +195,7 @@ jobs:
         id: extract_branch
 
         # The Azure API will sometimes supersede PUT requests that come in close together. This sleep will stagger the VM requests.
-        # It expects host_id to be an int and then multiplies it by 30s (i.e. host 0: sleep 0, host 1: sleep 60,...)
+        # It expects host_id to be an int and then multiplies it by 60s (i.e. host 0: sleep 0, host 1: sleep 60,...)
       - name: 'Stagger VM creation'
         shell: bash
         run: sleep `expr ${{matrix.host_id}} \* 60`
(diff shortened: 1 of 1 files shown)


.github/workflows/manual-deploy-testnet-l2.yml
```diff
@@ -223,9 +223,10 @@ jobs:
           inlineScript: |
             az vm open-port -g Testnet -n "${{needs.build.outputs.RESOURCE_STARTING_NAME}}-${{ matrix.host_id }}-${{ GITHUB.RUN_NUMBER }}"  --port 13000,13001,6060,6061,10000
 
-      - name: 'Wait for Obscuro sequencer to start'
+        # To overcome issues with missing dependencies on the VM we need to wait for the VM to be ready
+      - name: 'Allow time for VM initialization'
         shell: bash
-        run: if [ 0 != ${{ matrix.host_id }} ]; then echo "Sleeping while sequencer starts" & sleep 60; fi
+        run: sleep 60
```
otherview, line 229: We might still need the `Wait for Obscuro sequencer to start` to avoid the validator issuing a request secret before the sequencer has started ?

BedrockSquirrel: There is a step higher up (`Stagger VM creation`) that staggers them by 30 seconds already, I can bump that to 60 if you like?

This sleep had a bug in it until we fixed it this week so neither of them were sleeping, so I think the 30s offset must have been doing the trick.

otherview (approved):

PR #1261, opened 2023-05-12
Updated to run the upgrade nightly

moraygrieve:
### Why this change is needed

Run a schedule dev-testnet upgrade nightly, and use only repository dispatches

### DIFF
--- a/.github/workflows/manual-deploy-testnet-l2.yml
+++ b/.github/workflows/manual-deploy-testnet-l2.yml
@@ -352,4 +352,4 @@ jobs:
     steps:
       - name: 'Trigger Faucet deployment'
         run: |
-          curl -XPOST -H "Authorization: Bearer ${{ secrets.GH_TOKEN }}" -H "Accept:application/vnd.github" -H "Content-Type:application/json" https://api.github.com/repos/obscuronet/faucet/dispatches --data '{ "event_type": "${{ github.event.inputs.testnet_type/-/_ }}_deployed", "client_payload": { "ref": "${{ github.ref }}", "env": "testnet", "event": "${{ github.event_name }}" }'
+          curl -XPOST -H "Authorization: Bearer ${{ secrets.GH_TOKEN }}" -H "Accept:application/vnd.github" -H "Content-Type:application/json" https://api.github.com/repos/obscuronet/faucet/dispatches --data '{ "event_type": "${{ github.event.inputs.testnet_type/-/_ }}_deployed", "client_payload": { "ref": "${{ github.ref }}", "event": "${{ github.event_name }}" }'
--- a/.github/workflows/manual-deploy-testnet-l2.yml
+++ b/.github/workflows/manual-deploy-testnet-l2.yml
@@ -350,12 +350,6 @@ jobs:
     needs:
       - check-obscuro-is-healthy
     steps:
-      - name: 'Trigger Faucet deployment for testnet'
-        if: ${{ github.event.inputs.testnet_type  == 'testnet' }}
+      - name: 'Trigger Faucet deployment'
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/manual-upgrade-testnet-l2.yml
```diff
@@ -184,3 +190,18 @@ jobs:
...
           ./.github/workflows/runner-scripts/wait-node-healthy.sh --host=obscuronode-0-${{needs.build.outputs.RESOURCE_TESTNET_NAME}}-${{needs.build.outputs.VM_BUILD_NUMBER}}.uksouth.cloudapp.azure.com
           ./.github/workflows/runner-scripts/wait-node-healthy.sh --host=obscuronode-1-${{needs.build.outputs.RESOURCE_TESTNET_NAME}}-${{needs.build.outputs.VM_BUILD_NUMBER}}.uksouth.cloudapp.azure.com --timeout=60
+
+  deploy-faucet:
+    runs-on: ubuntu-latest
+    needs:
+      - check-obscuro-is-healthy
+    steps:
+      - name: 'Trigger Faucet deployment for testnet'
+        if: ${{ github.event.inputs.testnet_type  == 'testnet' }}
```
BedrockSquirrel, line 200: Bit of a shame to have use two steps with 'if' statements if there's just one minor difference between the two (the event type is dev_testnet_deployed), I think something like this would work instead of the if:

```
# event_type field requires the testnet type but with an underscore
eventtype="${github.event.inputs.testnet_type/-/_}_deployed"

curl -XPOST -H "Authorization: Bearer ${{ secrets.GH_TOKEN }}" -H "Accept:application/vnd.github" -H "Content-Type:application/json" https://api.github.com/repos/obscuronet/faucet/dispatches --data '{ "event_type": "${eventtype}", "client_payload": { "ref": "${{ github.ref }}", "env": "${{ github.event.inputs.testnet_type }}", "event": "${{ github.event_name }}" }'
```
Tested the replace locally:
image

BedrockSquirrel: You can use the same trick to avoid the branching on the health check step above

.github/workflows/manual-deploy-testnet-l2.yml
```diff
@@ -355,15 +348,8 @@ jobs:
...
-        run: |
-          curl -XPOST -H "Authorization: Bearer ${{ secrets.GH_TOKEN }}" -H "Accept:application/vnd.github" -H "Content-Type:application/json" https://api.github.com/repos/obscuronet/faucet/actions/workflows/manual-deploy-${{ github.event.inputs.testnet_type }}-faucet.yml/dispatches --data '{"ref": "main" }'
-
-      - name: 'Trigger Faucet deployment on schedule'
-        if: ${{ github.event_name == 'schedule' }}
+      - name: 'Trigger Faucet deployment'
         run: |
-          curl -XPOST -H "Authorization: Bearer ${{ secrets.GH_TOKEN }}" -H "Accept:application/vnd.github" -H "Content-Type:application/json" https://api.github.com/repos/obscuronet/faucet/dispatches --data '{ "event_type": "l2_dev_deployment", "client_payload": { "ref": "main", "env": "dev-testnet" }'
+          curl -XPOST -H "Authorization: Bearer ${{ secrets.GH_TOKEN }}" -H "Accept:application/vnd.github" -H "Content-Type:application/json" https://api.github.com/repos/obscuronet/faucet/dispatches --data '{ "event_type": "${{ github.event.inputs.testnet_type/-/_ }}_deployed", "client_payload": { "ref": "${{ github.ref }}", "env": "testnet", "event": "${{ github.event_name }}" }'
```
BedrockSquirrel, line 355: Does the "env" field need to be variable too? To say like `"env": "${{github.event.inputs.testnet_type}}"`

moraygrieve: yes, tbh easiest to just remove that anyway

BedrockSquirrel (approved):
LGTM

PR #1262, opened 2023-05-12
Fix syntax

moraygrieve:
### Why this change is needed

Fixes a syntax issue in the github workflows


BedrockSquirrel (approved):
lgtm if that works, I find the braces and brackets in these scripts super confusing haha

PR #1269, opened 2023-05-16
Storage updates atomicity

StefanIliev545:
### Why this change is needed

Storage operations that should be atomic are not, which is causing failures due to race conditions.

Moved the responsibility for who controls the database batch.

### PR checks pre-merging


tudor-malene (commented):
I must say I'm not a fan of this solution. Feels complicated.

What if we make all the methods of the storage take as first parameter a "*sql.Batch".

If it is nil, then it will create one, and then close it at the end of the function.
If it is not nil, then it will use it.

This gives you control from the caller if you want to execute multiple methods in the same batch or not

PR #1279, opened 2023-05-19
Enclave: always update L1 head on submit block

BedrockSquirrel:
### Why this change is needed

The enclave should always be operating from the tip of the L1 that it knows about. So we should treat every submitted L1 block as the new head.

- Remove the isLatest gate on updating L1 head
- Minor refactor of the block ingest method

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #1292, opened 2023-05-26
Standardise block and batch log references

BedrockSquirrel:
### Why this change is needed

As discussed in discord, it helps our datadog debugging if we stick to consistent referencing in the logs for block, batches etc.

This change didn't cover the whole codebase, just looking for the log messages that I'd found it hard to tie together and any others that I spotted along the way.

Changing logs to use the standard key constants from the `log.go` file

Adding some references to batch hash in log messages to help us keep track of when batches arrive where.

Getting rid of `.Hex()` in places (.String() which gets called by the logger and by `%s` operator uses .Hex())

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #1307, opened 2023-05-30
Only publish batches when up-to-date with L1

BedrockSquirrel:
### Why this change is needed

We should not publish batches while in 'catch-up', if there are issues with the sequencer it will halt batch production rather than carry on ignoring L1 activity that it's unaware of.

Add a flag to the host that is updated to true with every successful submission of an L1 block.

### PR checks pre-merging

### DIFF
--- a/go/host/host.go
+++ b/go/host/host.go
@@ -7,6 +7,7 @@ import (
 	"math/big"
 	"os"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/ethereum/go-ethereum/core/types"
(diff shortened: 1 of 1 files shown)


go/host/host.go
```diff
@@ -63,6 +64,7 @@ type host struct {
 	stopControl *stopcontrol.StopControl
 
 	l1BlockProvider hostcommon.ReconnectingBlockProvider
+	l1UpToDate      atomic.Bool                     // Whether the last submitted L1 block was the head L1 block
```
otherview, line 67: `l1UpToDate` -> `l1HeadProcessError` ? Seems more descriptive.

otherview (approved):
lgtm, the var name can be more descriptive imo

go/host/host.go
BedrockSquirrel: It's not about the error, it's about whether our enclave has received the current L1 head.

PR #1364, opened 2023-06-27
enclave-guardian: add enclave status tracker to host

BedrockSquirrel:
### Why this change is needed

Added a simple state machine that tracks the state of an enclave so host can change its behaviour accordingly. This will be within the enclave-guardian package but getting it in early because it makes the PRs for the other services easier.

Add an enclave status tracker that gets updated based on the hosts data about the enclave. Not being used for much yet but eventually it drives and guards all behaviour of host towards enclave.

### PR checks pre-merging

### DIFF
--- a/go/host/enclave/state.go
+++ b/go/host/enclave/state.go
@@ -1,0 +1,184 @@
+package enclave
+
+import (
+	"fmt"
+	"sync"
+
+	gethcommon "github.com/ethereum/go-ethereum/common"
+	gethlog "github.com/ethereum/go-ethereum/log"
+	"github.com/obscuronet/go-obscuro/go/common"
+	"github.com/obscuronet/go-obscuro/go/common/gethutil"
+)
+
+// This state machine compares the state of the enclave to the state of the world and is used to determine what actions can be taken with the enclave.
+// It records the last known status code of the enclave. It also records the l1 head and the l2 head that it believes the
+// enclave has processed, optimistically updating these after successful actions and verifying the status when errors occur.
+
+// Usage notes:
+// - The status is updated by the host when the enclave successfully processed blocks and batches
+// - The status is updated when we receive a status from the enclave
+// - The status is **not** updated immediately when it receives blocks/batches from the outside world (this is to avoid flickering between catch-up and live when new blocks/batches arrive)
+// - The state should be notified of a live block/batch arrival before notifying if it is successfully processed
+// - If unexpected error occurs when interacting with the enclave, then status should be requested and this state updated with the result
+
+// Status is the status of the enclave from the host's perspective (including what it knows of the outside world)
+type Status int
+
+const (
+	// Live - enclave is up-to-date with known external data. It can process L1 and L2 blocks as they arrive and respond to requests.
+	Live Status = iota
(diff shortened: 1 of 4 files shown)


go/host/enclave/state.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+// OnDisconnected is called if the enclave is unreachable/not returning a valid Status
+func (s *StateTracker) OnDisconnected() {
+	s.m.Lock()
+	defer s.m.Unlock()
+	s.setStatus(Disconnected)
+}
+
+// when enclave is operational, this method will update the status based on comparison of current chain heads with enclave heads
```
otherview, line 132: Why execute the set status in this unified method ?
What if the execute the set status was in each of the operations ? The lock is already adquired at that stage. Would that be simpler ?

go/host/host.go
```diff
@@ -413,16 +416,20 @@ func (h *host) startProcessing() {
 	for {
 		select {
 		case b := <-blockStream.Stream:
-			isLive := h.l1BlockProvider.IsLatest(b) // checks whether the block is the current head of the L1 (false if there is a newer block available)
+			l1Head, err := h.ethClient.FetchHeadBlock()
+			if err != nil {
```
otherview, line 420: This should prolly return / panic, otherwise its a nill point exception

go/host/host.go
```diff
@@ -1074,3 +1093,20 @@ func (h *host) validateConfig() {
 		h.logger.Crit("the host must specify a public P2P address")
 	}
 }
+
+// this function should be fired off in a new goroutine whenever the status of the enclave needs to be verified
+// (e.g. if we've seen unexpected errors from the enclave client)
+func (h *host) checkEnclaveStatus() {
+	// only allow one status request at a time, if lock unavailable then abandon attempt
+	if h.statusReqLock.TryLock() {
```
otherview, line 1101: image

I don't see anything wrong with it per-say but given the documentations and purely from a conservative POV I would probably suggest using an atomic int instead.

```go
if !atomic.CompareAndSwapUint32(&lock, 0, 1) {
        return  // it's locked
    }                                               
defer atomic.StoreUint32(&locker, 0)          
doTheThing()
```

go/host/host.go
```diff
@@ -413,16 +416,20 @@ func (h *host) startProcessing() {
...
+			}
+			h.enclaveState.OnReceivedBlock(l1Head.Hash())
+			isLive := l1Head.Hash() == b.Hash()
 			err = h.processL1Block(b, isLive)
 			if err != nil {
 				// handle the error, replace the blockStream if necessary (e.g. if stream needs resetting based on enclave's reported L1 head)
 				blockStream = h.handleProcessBlockErr(b, blockStream, err)
-				// failed to update the L1 head, so assume we're behind
-				h.l1UpToDate.Store(false)
+				// failed to update the L1 head, check enclave state
```
otherview, line 429:
```suggestion
				// failed to update the L1 head, spawn a goroutine to check enclave state
```

go/host/enclave/state.go
BedrockSquirrel: Yeah, good shout

otherview (approved):

PR #1372, opened 2023-07-04
Setup generic services in the host

BedrockSquirrel:
### Why this change is needed

We want the host to be a container that wires together a bunch of services running that depend on each other. It shouldn't have much logic itself.

So it's useful for the host to be able to treat these services the same as each other for starting/stopping/health checks.

Add a way to register services on the host. Make start/stop loop through services that are registered.

### PR checks pre-merging

### DIFF
--- a/go/host/host.go
+++ b/go/host/host.go
@@ -54,8 +54,9 @@ const (
 
 // Implementation of host.Host.
 type host struct {
-	config  *config.HostConfig
-	shortID uint64
+	config   *config.HostConfig
+	shortID  uint64
+	services map[string]hostcommon.Service // host services - registered by name for health checks, start and stop
 
 	p2p               hostcommon.P2P        // For communication with other Obscuro nodes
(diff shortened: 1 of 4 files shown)


go/host/host.go
```diff
@@ -1115,3 +1125,22 @@ func (h *host) catchUpL1Block() bool {
...
+
+func (h *host) GetService(name string) hostcommon.Service {
+	service, ok := h.services[name]
+	if !ok {
+		h.logger.Crit("requested service not registered", "name", name)
+	}
+	return service
+}
+
+func (h *host) l1Repo() hostcommon.L1BlockRepository {
```
BedrockSquirrel, line 1144: These lookups don't need to be used all over the place (I'm hoping we won't have the lookups at all).

When the enclave-guardian is created for example it will have the l1Repo injected into it. It's just used in here atm because the logic hasn't moved out yet.

tudor-malene (approved):
lgtm

PR #1379, opened 2023-07-10
Add subscription manager for sub/unsub

BedrockSquirrel:
### Why this change is needed

A few of these services in host refactor need to subscribe to be notified of things, a basic subscription manager helps keep the code for that tidied away and means they don't need to worry about thread-safety.

Add a generic sub/unsub subscription manager, use it for block repository.

### PR checks pre-merging

### DIFF
--- a/go/common/host/services.go
+++ b/go/common/host/services.go
@@ -26,8 +26,8 @@ type Service interface {
 
 // L1BlockRepository provides an interface for the host to request L1 block data (live-streaming and historical)
 type L1BlockRepository interface {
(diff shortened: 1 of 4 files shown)


BedrockSquirrel:
Note: at the moment it's just the host subscribing to the L1 repository and given that they have the same lifecycle I didn't bother holding onto the unsubscribe function to cleanup at shutdown.

Enclave-guardian and other services should unsubscribe when they are stopped.

tudor-malene (approved):
lgtm

PR #1388, opened 2023-07-18
Add batch repository to host using seq no as key

BedrockSquirrel:
### Why this change is needed

Part of the enclave-guardian prep (splitting out services so that logic can be per-enclave instead of dumped in host.go

These changes are all on the host side:
- Switch batch catch-up and batch p2p requests to use SeqNumber
- Add batch repository service
- Make P2P a service which has subscribers for different message types. BatchRepo subscribes for batch data and requests, the host still subscribes for transactions (eventually that will be the enclave-guardian directly)
- Rework mock p2p so the in-mem tests work

The only change on the enclave side is that the Status method now return the L2 head as a seq num instead of a hash. (And tweaking some comments).

### PR checks pre-merging


tudor-malene (commented):
looks good in principle.
Discussed on discord to add a few more clarifications/checks on the sequencer/validator code paths

tudor-malene (approved):
lgtm

PR #1393, opened 2023-07-19
fix issue with host failing to stop after error

BedrockSquirrel:
### Why this change is needed

Seeing issues with node restart tests that suggested host was failing to stop. In particular I saw failure to stop enclave with client causing an error which meant the host didn't finish stopping.

Log errors while stopping rather than returning them.

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #1394, opened 2023-07-20
Scheduled trigger of the faucet fix for schedule

moraygrieve:


BedrockSquirrel (approved):
LGTM, but should we add the same condition to the 'testnet' check above? (line 42)

PR #1395, opened 2023-07-20
Try feeding consecutive blocks if failing after a fork

BedrockSquirrel:
### Why this change is needed

We have an edgecase after returning to an L1 fork where using the state machine to decide what block to send next is a bit fragile. We should instead try consecutive canonical blocks from the LCA until we get to one that the enclave hasn't seen

Also, this disables a flaky obscuroscan test until we get a chance to make it more robust.

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #1397, opened 2023-07-20
Obscuroscan v2 deploy 

otherview:
### Why this change is needed

https://github.com/obscuronet/obscuro-internal/issues/1958

- Adds a dockerfile for obscuroscan
- Adds a manual trigger to deploy it to azure

### PR checks pre-merging

### DIFF
--- a/.github/workflows/manual-deploy-obscuro-scan-2.yml
+++ b/.github/workflows/manual-deploy-obscuro-scan-2.yml
@@ -1,0 +1,69 @@
+# Deploys Obscuro Scan on Azure for Testnet
+# Builds the Obscuro Scan image, pushes the image to dockerhub and starts the Obscuro Scan on Azure
+
+name: '[M] Deploy Obscuro Scan 2 Testnet'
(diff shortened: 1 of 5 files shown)


moraygrieve:
Is the intention is to have the v2 in parallel to the current implementation, and then when complete remove the old one?

otherview:
> Is the intention is to have the v2 in parallel to the current implementation, and then when complete remove the old one?

Yep, that is correct!

BedrockSquirrel (approved):
Looks reasonable to me, don't know about the port numbers but guess you can figure that all out when we're productionising. Looks like atm it's exposing FE on 81 and backend on 82 from the VM?

otherview:
> Looks reasonable to me, don't know about the port numbers but guess you can figure that all out when we're productionising. Looks like atm it's exposing FE on 81 and backend on 82 from the VM?

Yes, when you run it "locally" the FE takes port 81 and BE 82, that's because the faucet is already taking port 80 in the local testnet launcher.

PR #1401, opened 2023-07-24
Put in some logging

moraygrieve:
Calling a re-usable workflow from a schedule _should_ pass in the testnet_type in the inputs context according to all documentation, but it doesn't seem to be working. There are reported bugs of if conditions in re-usable workflows called from a workflow_call, so I want to put in a little more logging and up the schedule to try and fix today.


otherview (approved):
lgtm - would add in suggesting the use of the `run-name` too. We seem to use it in some places

Eg. `run-name: '[M] Deploy Obscuro Scan Testnet ( ${{ github.event.inputs.testnet_type }} )'`

PR #1403, opened 2023-07-24
Hook obscuro scan other endpoints

otherview:
### Why this change is needed

https://github.com/obscuronet/obscuro-internal/issues/1962
https://github.com/obscuronet/obscuro-internal/issues/1965

### PR checks pre-merging

### DIFF
--- a/tools/obscuroscan_v2/frontend/src/lib/config.dev.js
+++ b/tools/obscuroscan_v2/frontend/src/lib/config.dev.js
@@ -1,5 +1,5 @@
 class Config {
-    static backendServerAddress = "http://127.0.0.1:8080"
+    static backendServerAddress = "http://127.0.0.1:43910"
 }
 
 export default Config
--- a/tools/obscuroscan_v2/frontend/src/stores/counterStore.js
+++ b/tools/obscuroscan_v2/frontend/src/stores/counterStore.js
@@ -14,14 +14,14 @@ export const useCounterStore = defineStore({
         async fetchCount() {
             this.loading = true;
             try {
-                let response = await fetch( Config.backendServerAddress+'/count/contracts/');
-                let data = await response.json();
-                this.totalContractCount = data.count;
+                const totContractResp = await fetch( Config.backendServerAddress+'/count/contracts/');
+                const totContractData = await totContractResp.json();
+                this.totalContractCount = totContractData.count;
                 console.log("Fetched "+this.totalContractCount);
 
-                response = await fetch( Config.backendServerAddress+'/count/transactions/');
-                data = await response.json();
-                this.totalTransactionCount = data.count;
+                const totTxResp = await fetch( Config.backendServerAddress+'/count/transactions/');
+                const totTxData = await totTxResp.json();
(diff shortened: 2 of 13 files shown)


tools/obscuroscan_v2/frontend/src/lib/config.dev.js
```diff
@@ -1,5 +1,5 @@
 class Config {
-    static backendServerAddress = "http://127.0.0.1:8080"
+    static backendServerAddress = "http://127.0.0.1:43910"
```
BedrockSquirrel, line 2: What's this port about? You need to configure that somewhere locally to run dev server or it just works?

tools/obscuroscan_v2/frontend/src/stores/counterStore.js
```diff
@@ -13,11 +14,16 @@ export const useCounterStore = defineStore({
         async fetchCount() {
             this.loading = true;
             try {
-                console.log(Config.backendServerAddress)
-                const response = await fetch( Config.backendServerAddress+'/count/contracts/');
-                const data = await response.json();
-                this.count = data.count;
-                console.log("Fetched "+this.count);
+                let response = await fetch( Config.backendServerAddress+'/count/contracts/');
+                let data = await response.json();
```
BedrockSquirrel, line 18: Is it safe to have `response` and `data` be overwritten variables like this? I think general principle to avoid risk of weird bugs is just to use `const` unless you've got a good reason.

tools/obscuroscan_v2/frontend/src/stores/priceStore.js
```diff
@@ -0,0 +1,43 @@
...
+        ethPriceUSD: null,
+        loading: false,
+        pollingInterval: 5000,  // 5 seconds
+        timer: null,
+    }),
+    actions: {
+        async fetchCount() {
+            this.loading = true;
+            try {
+                let response = await fetch( 'https://api.coingecko.com/api/v3/simple/price?ids=ethereum&vs_currencies=usd');
```
BedrockSquirrel, line 16: Again, best to use `const` everywhere imo. Just stops people from reading the code from having to wonder if there are shenanigans afoot.

BedrockSquirrel (approved):
LGTM - minor optional niggles

tools/obscuroscan_v2/frontend/src/lib/config.dev.js
otherview: It's a placeholder. It should point to the api port for the dev environment. I guess I'll change, and set it as default, once I hook up the launcher to obscuroscan ?

tools/obscuroscan_v2/frontend/src/stores/counterStore.js
otherview: Totaly right, fixed it 👍

tools/obscuroscan_v2/frontend/src/stores/priceStore.js
otherview: Fixed 👍

PR #1417, opened 2023-08-02
allow enclave to produce handlers for both old and new message types

zkokelj:
### Why this change is needed

This change is needed, because current implementation only allows for old message formats signed by users (`vk<vk>`) and returns following error for new format used by Obscuro Gateway `Register <userID> for <account>`.

`unable to create VK encryptor - unable to create vk encryption for request - invalid viewing key signature for requested address`

For the transition period I want to allow all formats not to break compatibility with WE and will delete the support when WE endpoints are removed and Obscuro Gateway is the only way to use it.

Changes in the constructor for `VKHandler` + added and updated tests

### PR checks pre-merging

### DIFF
--- a/go/common/viewingkey/viewing_key.go
+++ b/go/common/viewingkey/viewing_key.go
@@ -97,6 +97,14 @@ func Sign(userPrivKey *ecdsa.PrivateKey, vkPubKey []byte) ([]byte, error) {
 
 // GenerateSignMessage creates the message to be signed
 // vkPubKey is expected to be a []byte("0x....") to create the signing message
+// todo (@ziga) Remove this method once old WE endpoints are removed
 func GenerateSignMessage(vkPubKey []byte) string {
 	return SignedMsgPrefix + hex.EncodeToString(vkPubKey)
+}
+
+// GenerateSignMessageOG creates the message to be signed by Obscuro Gateway (new format)
+// format is expected to be "Register <userID> for <Account>"
+func GenerateSignMessageOG(vkPubKey []byte, addr *gethcommon.Address) string {
+	userID := crypto.Keccak256Hash(vkPubKey).Bytes()
+	return fmt.Sprintf("Register %s for %s", hex.EncodeToString(userID), addr.Hex())
 }
--- a/go/enclave/vkhandler/vk_handler.go
+++ b/go/enclave/vkhandler/vk_handler.go
@@ -25,9 +25,10 @@ type VKHandler struct {
 // New creates a new viewing key handler
 // checks if the signature is valid
 // as well if signature matches account address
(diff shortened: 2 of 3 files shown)


go/common/viewingkey/viewing_key.go
```diff
@@ -97,6 +97,14 @@ func Sign(userPrivKey *ecdsa.PrivateKey, vkPubKey []byte) ([]byte, error) {
 
 // GenerateSignMessage creates the message to be signed
 // vkPubKey is expected to be a []byte("0x....") to create the signing message
+// todo (@ziga) Remove this method once old WE endpoints are removed
```
otherview, line 100: Let's add a github issue lest we forget :)

go/enclave/vkhandler/vk_handler.go
```diff
@@ -25,9 +25,10 @@ type VKHandler struct {
 // New creates a new viewing key handler
 // checks if the signature is valid
 // as well if signature matches account address
+// todo (@ziga) - function now accepts both old and new messages
 func New(requestedAddr *gethcommon.Address, vkPubKeyBytes, accountSignatureHexBytes []byte) (*VKHandler, error) {
 	// Recalculate the message signed by MetaMask.
-	msgToSign := viewingkey.GenerateSignMessage(vkPubKeyBytes)
+	msgToSign := viewingkey.GenerateSignMessageOG(vkPubKeyBytes, requestedAddr)
```
otherview, line 31: Wont' this return err if it's the old message type ?

go/enclave/vkhandler/vk_handler_test.go
```diff
@@ -34,7 +34,64 @@ func TestVKHandler(t *testing.T) {
 	assert.NoError(t, err)
 }
 
-func TestSignAndCheckSignature(t *testing.T) {
+func TestVKHandlerOGMessageFormat(t *testing.T) {
```
otherview, line 37: Should this be a table test ? 
https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

go/enclave/vkhandler/vk_handler.go
zkokelj: No, I've tested both cases in the tests.
(We receive address, pubKey and signature)
Message affects just a hash and wrong address is returned in that case a few lines below.

go/common/viewingkey/viewing_key.go
zkokelj: added: https://github.com/orgs/obscuronet/projects/1/views/36?pane=issue&itemId=34873032

otherview (approved):
good stuff 👍

PR #1426, opened 2023-08-07
Ziga/fix og user id and address problems

zkokelj:
### Why this change is needed

Previously there was an error on how userID was calculated and different options of upper/lower chars in addresses were not addressed.

- new (correct) way of calculating userID
- lowercase address in message

### PR checks pre-merging

### DIFF
--- a/go/common/viewingkey/viewing_key.go
+++ b/go/common/viewingkey/viewing_key.go
@@ -4,6 +4,7 @@ import (
 	"crypto/ecdsa"
 	"encoding/hex"
 	"fmt"
+	"strings"
 
 	"github.com/ethereum/go-ethereum/accounts"
 	"github.com/ethereum/go-ethereum/crypto"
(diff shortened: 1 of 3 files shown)


go/common/viewingkey/viewing_key.go
```diff
@@ -106,5 +107,5 @@ func GenerateSignMessage(vkPubKey []byte) string {
 // format is expected to be "Register <userID> for <Account>"
 func GenerateSignMessageOG(vkPubKey []byte, addr *gethcommon.Address) string {
 	userID := crypto.Keccak256Hash(vkPubKey).Bytes()
-	return fmt.Sprintf("Register %s for %s", hex.EncodeToString(userID), addr.Hex())
+	return fmt.Sprintf("Register %s for %s", hex.EncodeToString(userID), strings.ToLower(addr.Hex()))
```
otherview, line 110: Good catch - worth adding this to the function comment too imo

otherview (approved):
lgtm, small nit to add the lowering to the function method.

PR #1430, opened 2023-08-08
Reduce retention to 1 day on failure only

moraygrieve:
### Why this change is needed

We regularly fill up our usage on actions storage and it looks like the ci-logs are the culprit. This PR only uploads the logs as artifacts on failure, and only with a retention of 1 day.


tudor-malene (approved):
lgtm

PR #1431, opened 2023-08-08
Refactor host construction to receive a factory for each service

BedrockSquirrel:
### Why this change is needed

Host was in a halfway state between logic in host.go and logic split into services. The wiring up and initialisation was a bit of a mess.

Simplify the host wiring:
- Host now has no logic, everything is in services
- Host receives a factory for each service, it creates them (with personalised loggers), starts them, monitors them, stops them
- Services declare which services they depend on with a private serviceLocator interface
- Service factories know how to create everything the service needs from config
- In-mem test has different factories that inject the mocked p2p/eth client/enclave accessor

### PR checks pre-merging


go/common/host/services.go
```diff
@@ -31,7 +23,7 @@ const (
 // They should be resilient to services availability, because the construction ordering is not guaranteed.
 type Service interface {
 	Start() error
-	Stop() error
+	Stop()
```
otherview, line 26: In the past we had issues with tests because stop was not stopping correctly. We would either ignore the error or swallow it.

Prolly best to always check to for errors ?

go/common/host/services.go
```diff
@@ -157,3 +149,16 @@ type LogSubscriptionManager interface {
...
 	SendLogsToSubscribers(result *common.EncryptedSubscriptionLogs)
 }
+
+// Metrics service doesn't currently expose any methods internally to the host, its role is just to serve metrics data via HTTP
+type Metrics interface {
+	Registry() metrics.Registry
+}
+
+// RPCServer doesn't currently expose any methods internally to the host, its role is just to serve requests
+type RPCServer interface{}
```
otherview, line 159: Is this meant to be and interface{} ?

go/host/container/servicefactories.go
```diff
@@ -0,0 +1 @@
+package container
```
otherview, line 1: Leftover ?

go/common/metrics/metrics.go
```diff
@@ -17,39 +23,55 @@ var (
 	_threadCreateProfile                  = pprof.Lookup("threadcreate")
 )
 
+type metricsServiceLocator interface {
```
otherview, line 26: Why is this a private interface ?

What if this was it's own package ?

go/common/metrics/metrics.go
```diff
@@ -17,39 +23,55 @@ var (
...
+	DB() *db.DB
+}
+
 // Service provides the metrics for the host
 // it registers the gethmetrics Registry
 // and handles the metrics server
 type Service struct {
 	registry gethmetrics.Registry
+	sl       metricsServiceLocator
```
otherview, line 35: Then here it would be ok to import the Service interface

go/host/container/host_container.go
```diff
@@ -2,93 +2,49 @@ package container
...
+	Host   hostcommon.Host
+	Logger gethlog.Logger
 }
 
 func (h *HostContainer) Start() error {
-	h.metricsService.Start()
-
 	// make sure the rpc server has a host to render requests
-	err := h.host.Start()
+	err := h.Host.Start()
```
otherview, line 31: Perhaps outside the scope of this PR but would be cool if the container Start started the component as a go routine with a channel and took care of that kind of messaging/lifecycle. 
I think we do that in obscuroscan container, but will double check :)

go/host/container/host_container.go
```diff
@@ -2,93 +2,49 @@ package container
...
-)
-
 type HostContainer struct {
-	host           hostcommon.Host
-	logger         gethlog.Logger
-	metricsService *metrics.Service
-	rpcServer      clientrpc.Server
+	// Host and Logger are temporarily public to allow fudge in the test wiring
+	// todo (@matt) make these private again once we no longer need to wire the Stop method into the host factories
+	Host   hostcommon.Host
```
otherview, line 25: This might be a pointer that something is funky here. 
Typically the containers never provide access to their components.

go/host/container/host_container.go
otherview, line 31: This kinda defeats the purpose of the container.

It's just a passtrhough now because it has no underlying logic.

go/host/container/host_container.go
```diff
@@ -2,93 +2,49 @@ package container
...
-
-	if h.rpcServer != nil {
-		// rpc server cannot be stopped synchronously as it will kill current request
-		go func() {
-			// make sure it's not killing the connection before returning the response
-			time.Sleep(time.Second) // todo review this sleep
-			h.rpcServer.Stop()
-		}()
-	}
+	h.Host.Stop()
```
otherview, line 43: Similar here. 

Before there was a Host that took some dependecies but did not manage the life cycle of the dependencies.

Which meant that the Host was cleaner as it's responsabilites had a smaller scope.
Now the Host handles the dependencies.

go/host/db/hostdb.go
```diff
@@ -38,25 +41,42 @@ type DB struct {
 	blockReads  gethmetrics.Gauge
 }
 
+func (db *DB) Start() error {
+	return nil
+}
+
 // Stop is especially important for graceful shutdown of LevelDB as it may flush data to disk that is currently in cache
-func (db *DB) Stop() error {
+func (db *DB) Stop() {
```
otherview, line 49: We should prolly surface these errors and handle them at the stop handler at the highest level.

go/host/db/hostdb.go
```diff
@@ -38,25 +41,42 @@ type DB struct {
...
 }
 
-func CreateDBFromConfig(cfg *config.HostConfig, regMetrics gethmetrics.Registry, logger gethlog.Logger) (*DB, error) {
+func (db *DB) HealthStatus() hostcommon.HealthStatus {
+	// always healthy for now, satisfies Service interface
+	return &hostcommon.BasicErrHealthStatus{ErrMsg: ""}
+}
+
+// InitialiseMetrics registers the host DB metrics with the given registry (this is called by the metrics service when it starts)
+func (db *DB) InitialiseMetrics(registry gethmetrics.Registry) {
```
otherview, line 63: If this happens everytime the DB is created, why make a new method ?

Can it be that the problem is that the inmem db is using a struct direct initialization instead of a function ? I mean it would be easy to do a check for nil registry right ?

otherview: Perhaps if this was controlled by the host container then it would be easy to manage the dependecies. Like the registry would be created and then when the db is created with the registry injection in it.

otherview (commented):
Had a good first pass, still have to dig more on the integration, tests and tools.

Perhaps we should rethink the service approach. 
Hard to see the benefit, and it seems they are making the flow of deps more complicated.
I wrote more reasons in more detail, which I'll ping you, but the core imo is:

- It's making the host and it's components more complex

- The host is no longer taking care of host operations, it's the center piece of communication between services.

Mainly we want the host to not care about dependencies.
Or the dependecies of dependencies.(Like db and metrics).
This allows us to have loose coupling between components and that makes it's easier to isolate whats happening.

go/common/host/services.go
BedrockSquirrel: All we can do is log the error if it gets returned, I think the contract should be that the services try their best to stop any go-routines they own and log any errors in the process.

go/common/host/services.go
BedrockSquirrel: I was keen to give it a type but the RPCServer service doesn't expose any methods to the rest of the host currently

BedrockSquirrel:
Abandoned this part of the refactor for now as team discussions ongoing about host architecture.

PR #1433, opened 2023-08-08
Avoid panics when sequencer host thinks it is behind the enclave

BedrockSquirrel:
### Why this change is needed

Sometimes we hit the following panic while running:
`enclave service is not a validator but validator was requested!`

This is caused by the host attempting to submit a batch to a sequencer enclave. This seems to be caused by the host briefly flickering into an L2 catch-up state because it thinks it is behind on batches (which shouldn't be possible for the sequencer).

Avoid L2 catchup in the sequencer host so we do not panic. Instead log a warning so we  have visibility to better understand why this is happening.

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #1434, opened 2023-08-08
Obscuroscan - Add Pagination in the list calls (Public and Encrypted)

otherview:
### Why this change is needed

https://github.com/obscuronet/obscuro-internal/issues/2017

- Adds Pagination to CustomQuery and PublicTransaction calls
- `PrivateQueryResponse` and `PublicTxData` have now a Total field ( so we can paginate )
- Added pagination to the FE tables
- Refactored the Polling mechanism used across different stores

### PR checks pre-merging


go/enclave/enclave.go
```diff
@@ -1344,38 +1344,60 @@ func (e *enclaveImpl) GetReceiptsByAddress(encryptedParams common.EncryptedParam
 		return responses.AsPlaintextError(fmt.Errorf("unable to decode eth_getStorageAt params - %w", err)), nil
 	}
 
-	// Parameters are [ViewingKey, Address, TBD, TBD]
+	// Parameters are [ViewingKey, PrivateCustomQueryHeader, PrivateCustomQueryArgs, null]
```
zkokelj, line 1347: Why is required number of parameters == 4 (with last being null). Is this because we reuse getStorageAt?

go/enclave/genesis/testnet_genesis.go
```diff
@@ -7,6 +7,7 @@ import (
 	gethcommon "github.com/ethereum/go-ethereum/common"
 )
 
+// 0xA58C60cc047592DE97BF1E8d2f225Fc5D959De77 ->
```
zkokelj, line 10: It is unclear what this means. Is it an address with testnet prefundedPK?

integration/obscuroscan/obscuroscan_test.go
```diff
@@ -93,19 +103,28 @@ func TestObscuroscan(t *testing.T) {
...
+	assert.Equal(t, uint64(1), publicTxsObj.Result.Total)
+
+	issueTransactions(
+		t,
+		fmt.Sprintf("ws://127.0.0.1:%d", startPort+integration.DefaultHostRPCWSOffset),
+		wallet.NewInMemoryWalletFromConfig("8dfb8083da6275ae3e4f41e3e8a8c19d028d32c9247e24530933782f2a05035b", integration.ObscuroChainID, testlog.Logger()),
+		100,
+	)
 
+	// time.Sleep(time.Hour)
```
zkokelj, line 127: this comment can be removed (unless you use it often to test things locally)

zkokelj (approved):
Nice work! LGTM :)
Left a few small comments

StefanIliev545 (approved):
No gulag

go/enclave/enclave.go
otherview: Yep. The length of the params slice is 4, even though the 4th position is null. 
This avoids a few problems we had in past were we would receive params slices with a length of 3 and then had to implement weird logic to accommodate for it.

integration/obscuroscan/obscuroscan_test.go
otherview: Yes, I typically test with that. And because the test isn't enabled it's probably fine to keep it there for now.

PR #1438, opened 2023-08-10
Fix bug with restarting enclaves sqlite connection

BedrockSquirrel:
### Why this change is needed

There was an issue with restarting nodes that use sqlite for the enclave where they couldn't find the sqlite file after restart and so created a new database.

Tweak the format of the sqlite3 connection to match the docs so the filename is findable after restart.

Before the change the files were being saved like this:
image

After the change:
image

### PR checks pre-merging


go/enclave/enclave.go
```diff
@@ -109,6 +109,9 @@ func NewEnclave(
 	mgmtContractLib mgmtcontractlib.MgmtContractLib,
 	logger gethlog.Logger,
 ) common.Enclave {
+	jsonConfig, _ := json.MarshalIndent(config, "", "  ")
```
BedrockSquirrel, line 112: I moved this config dump higher up because I was hitting crit errors in the constructor method below before it reached the config log msg, making it hard to know what was going on.

StefanIliev545: there is this library i added, sdumps or whatever that should be better

StefanIliev545 (approved):

go/enclave/enclave.go
BedrockSquirrel: Nice yeah I've seen that around, the `litter` one. Not sure how well it will work with log messages though because of the line breaks, looks like all the current usages are for stdout prints.

But yeah agree we should try it when we get some time cos it is a PITA to read. I'll leave it for this PR though cos was just moving it earlier.

PR #1449, opened 2023-08-14
Enclave should report latest batch received even if unexecuted

BedrockSquirrel:
### Why this change is needed

The enclave reports in its status to the host what the latest L2 batch it has received is. But we recently updated that head batch query used for the status to only include executed batches.

So the host gets this status and thinks the enclave has lost the batch it just sent somehow and puts it into catchup. This means the host is constantly flickering to catchup unnecessarily.

 Use current seq number for the status reporting instead of head batch.

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #1451, opened 2023-08-15
p2p batch request data is historic not live

BedrockSquirrel:
### Why this change is needed

Response batches message has flag live=true which isn't catastrophic but might make L2 catch up much slower because L2 repo won't realise that p2p response came back.

Correct the `live` flag to `false` in batch data response.

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #1452, opened 2023-08-15
Fix wiring of L1 start hash

BedrockSquirrel:
### Why this change is needed

Configured L1 start hash wasn't making it to the enclave guardian so we were still streaming from genesis of L1.

### PR checks pre-merging

### DIFF
--- a/go/host/enclave/state.go
+++ b/go/host/enclave/state.go
@@ -126,12 +126,7 @@ func (s *StateTracker) OnEnclaveStatus(es common.Status) {
 	s.m.Lock()
 	defer s.m.Unlock()
 	s.enclaveStatusCode = es.StatusCode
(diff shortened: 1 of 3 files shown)


go/host/enclave/state.go
```diff
@@ -126,12 +126,7 @@ func (s *StateTracker) OnEnclaveStatus(es common.Status) {
 	s.m.Lock()
 	defer s.m.Unlock()
 	s.enclaveStatusCode = es.StatusCode
-	if es.L1Head != gethutil.EmptyHash {
```
BedrockSquirrel, line 129: We don't need this hack any more, if enclave has no head we don't try to pretend its head is the configured head, we just check for that and use configured start hash.

tudor-malene (approved):
lgtm

PR #1459, opened 2023-08-16
Clarify in-mem flag is for host DB

BedrockSquirrel:
### Why this change is needed

[Issue 2058: WithInMemFlag seemingly ignored](https://github.com/obscuronet/obscuro-internal/issues/2058)

Actually turned out that flag is for the host DB to be in-memory. And think we have found a fix for the persisted sqlite db issues on the enclave so will leave that as hardcoded to persisted for now (lots of the settings are hardcoded where they until the tests need flexibility).

Rename the in-mem flag method to make it clear it is for the host DB.

### PR checks pre-merging

### DIFF
--- a/go/node/config.go
+++ b/go/node/config.go
@@ -247,7 +247,7 @@ func WithPCCSAddr(s string) Option {
 	}
 }
 
-func WithInMemoryDB(b bool) Option {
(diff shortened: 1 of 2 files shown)


moraygrieve (approved):
LGTM

PR #1466, opened 2023-08-17
Update p2p stopping mechanism

BedrockSquirrel:
### Why this change is needed

Probably not an issue, but if P2P service is stopped we should abandon any threads that are trying to refresh the peer list.

- Update P2P stopper to use atomic.Bool as more convenient
- Have refresh p2p list bail out if service is stopped while it is retrying

### PR checks pre-merging

### DIFF
--- a/go/host/p2p/p2p.go
+++ b/go/host/p2p/p2p.go
@@ -82,8 +82,8 @@ type Service struct {
 	txSubscribers    *subscription.Manager[host.P2PTxHandler]
 	batchReqHandlers *subscription.Manager[host.P2PBatchRequestHandler]
 
-	listener          net.Listener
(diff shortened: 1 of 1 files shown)


tudor-malene (approved):
lgtm

go/host/p2p/p2p.go
```diff
@@ -107,8 +107,7 @@ func (p *Service) Start() error {
 	}
 
 	p.logger.Info("P2P server started listening", "bindAddress", p.ourBindAddress, "publicAddress", p.ourPublicAddress)
-	i := int32(0)
-	p.listenerInterrupt = &i
+	p.running.Store(true)
```
otherview, line 110: We should use the stopcontrol package like in the host

PR #1472, opened 2023-08-19
Bug: nil pointer when eth new block not found

BedrockSquirrel:
### Why this change is needed

When using Infura RPC endpoint for sepolia data we were getting nil pointers with the new block lookup. This shouldn't happen but if it does we shouldn't nil pointer and should be able to recover.

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #1481, opened 2023-08-24
Upgrading geth to 1.12

otherview:
### Why this change is needed

https://github.com/obscuronet/obscuro-internal/issues/2076

- Upgrades geth to 1.12

### PR checks pre-merging


.golangci.yml
```diff
@@ -31,7 +31,7 @@ linters:
     - goheader
     - goimports
     - revive
-    - gomoddirectives
+    #- gomoddirectives
```
BedrockSquirrel, line 34: Worth a note or something for what this linter is or why it's commented?

go/enclave/components/batch_registry.go
```diff
@@ -132,7 +132,7 @@ func (br *batchRegistry) GetBatchStateAtHeight(blockNumber *gethrpc.BlockNumber)
 
 func (br *batchRegistry) GetBatchAtHeight(height gethrpc.BlockNumber) (*core.Batch, error) {
 	var batch *core.Batch
-	switch height {
+	switch height { //
```
BedrockSquirrel, line 135: `¯\_(ツ)_/¯`

go/enclave/components/batch_registry.go
```diff
@@ -148,6 +148,7 @@ func (br *batchRegistry) GetBatchAtHeight(height gethrpc.BlockNumber) (*core.Bat
 			return nil, fmt.Errorf("batch with requested height %d was not found. Cause: %w", height, err)
 		}
 		batch = headBatch
+	case gethrpc.SafeBlockNumber, gethrpc.FinalizedBlockNumber:
```
BedrockSquirrel, line 151: Looks like this returns `nil, nil` at the mo, is that intentional? I guess it should return an 'unsupported' error if we're not going to handle it just yet.

BedrockSquirrel (commented):
I know you're still tidying up so might still be getting to these but just had quick skim through while I was here.

Also could you bump the geth version in eth2_binaries.go to `1.12.2` - shouldn't cause any issues from what I've seen, just stops old version showing up in logs when we run tests.

And I didn't bother commenting about removing the `Println`s and doing something with the commented out stuff cos I guess you're already on that.

go/enclave/components/batch_registry.go
otherview: Not sure I follow, I just added these as default behavior to avoid the lint error.
Should I add a condition to return 'unsuported' ?

BedrockSquirrel: They don't fall through to default though, golang's weird with that isn't it (https://go.dev/play/p/SjqwN7NZflI)

so anyone calling with that param will get nil, nil back. I'd suggest adding these two to the one above like:

```
// for now we are treating `finalized` and `safe` heads the same as latest head
case gethrpc.LatestBlockNumber, gethrpc.SafeBlockNumber, gethrpc.FinalizedBlockNumber:
```

otherview: Yeah now I see what you mean, I forget to use the `fallthrough` keyword.

But that suggestion makes sense, fixed 👍

BedrockSquirrel (approved):
LGTM

PR #1483, opened 2023-08-25
Bump eth2network

otherview:
### Why this change is needed

- Bumps geth to 1.12.2 
- Fixes the hardhat deployer static call
- Fixes the way the Health endpoint was called
- Updated the dlv version in the enclave_debug docker image

### DIFF
--- a/testnet/launcher/docker.go
+++ b/testnet/launcher/docker.go
@@ -1,12 +1,9 @@
 package launcher
 
 import (
-	"bytes"
-	"context"
-	"encoding/json"
 	"fmt"
-	"io"
-	"net/http"
+	"github.com/obscuronet/go-obscuro/go/obsclient"
+	"github.com/obscuronet/go-obscuro/go/rpc"
 	"time"
 
 	"github.com/obscuronet/go-obscuro/go/common/retry"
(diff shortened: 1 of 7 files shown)


testnet/launcher/docker.go
```diff
@@ -210,49 +207,30 @@ func deployL1Contracts() (*node.NetworkConfig, error) {
...
 			if err != nil {
 				return err
 			}
-
-			response := map[string]interface{}{}
-			err = json.Unmarshal(resBody, &response)
-			if err != nil {
-				return err
+			if health {
+				fmt.Println("obscuro node is ready")
```
zkokelj, line 228: Is this println here intentionally? 
If yes maybe capitalise Obscuro (`obscuro node is ready` -> `Obscuro node is ready`)

testnet/launcher/docker.go
```diff
@@ -210,49 +207,30 @@ func deployL1Contracts() (*node.NetworkConfig, error) {
...
-						if health, ok := overallHealth.(bool); ok && health {
-							fmt.Println("obscuro node is ready")
-							return nil
-						}
-					}
-				}
-			}
 			return fmt.Errorf("node OverallHealth is not good yet")
-		}, retry.NewTimeoutStrategy(2*time.Minute, 1*time.Second),
+		}, retry.NewTimeoutStrategy(5*time.Minute, 1*time.Second),
```
zkokelj, line 233: Is there a reason to increase timeout here?

zkokelj (approved):
LGTM! Left 2 minor comments

testnet/launcher/docker.go
otherview:
```suggestion
				fmt.Println("Obscuro node is ready")
```

testnet/launcher/docker.go
otherview: I saw some false positives when running the nodes locally.

StefanIliev545 (approved):

PR #1486, opened 2023-08-28
use versioning with Obscuro Gateway API

zkokelj:
### Why this change is needed

We want to have versions in the Obscuro Gateway API to allow new versions in the future.

As path for `eth` requests change to `/v1` we can serve frontend on `/` instead of `/og`

/v1/ is used before Obscuro Gateway endpoints (except health, ready and old walletExtension endpoints that will be removed soon)

### PR checks pre-merging

### DIFF
--- a/tools/walletextension/api/staticOG/javascript.js
+++ b/tools/walletextension/api/staticOG/javascript.js
@@ -6,10 +6,10 @@ const idAddAllAccounts = "addAllAccounts";
 const idRevokeUserID = "revokeUserID";
 const idStatus = "status";
 const idUserID = "userID";
-const pathJoin = "/join/";
-const pathAuthenticate = "/authenticate/";
-const pathQuery = "/query/";
-const pathRevoke = "/revoke/";
+const pathJoin = "/v1/join/";
+const pathAuthenticate = "/v1/authenticate/";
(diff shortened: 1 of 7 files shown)


tools/walletextension/api/staticOG/javascript.js
```diff
@@ -6,10 +6,10 @@ const idAddAllAccounts = "addAllAccounts";
 const idRevokeUserID = "revokeUserID";
 const idStatus = "status";
 const idUserID = "userID";
-const pathJoin = "/join/";
-const pathAuthenticate = "/authenticate/";
-const pathQuery = "/query/";
-const pathRevoke = "/revoke/";
+const pathJoin = "/v1/join/";
```
BedrockSquirrel, line 9: Maybe good to have API version constant like in the Go to build all these URLs with?

BedrockSquirrel (approved):
LGTM - optional comment

otherview (commented):
LGTM - Since we're expanding on the routes and URL can a comment block or a readme file be created with the routes and a super tiny description on what they do ?

tools/walletextension/api/staticOG/javascript.js
zkokelj: Added, thanks!

PR #1487, opened 2023-08-29
Uses a prebuilt obscuro node image

otherview:
### Why this change is needed

https://github.com/obscuronet/obscuro-internal/issues/1449

Improves the deploy time by having a pre-built image instead of installing everything on the fly.
Should also reduce install-time flakyness, like errors on apt-get update / docker install, that we have seen in the past.

Defaults to using the ObscuroConfUbuntu image
Added a text file on how to create these images for future reference.

### PR checks pre-merging


StefanIliev545 (approved):
LGTM

PR #1491, opened 2023-08-31
Add a IsIncomingP2PEnabled Flag

otherview:
### Why this change is needed

https://github.com/obscuronet/obscuro-internal/issues/2103

To test the system with p2p down.

- Adds an IsIncomingP2PEnabled
- Sets one node disabled in both sims
- Adds an NoIncomingP2P implementation of P2P
- Fixed a bug where failing to send batches to a host would delay the whole processing of the sequencer

### PR checks pre-merging


go/enclave/events/subscription_manager.go
```diff
@@ -151,7 +151,6 @@ func (s *SubscriptionManager) GetSubscribedLogsForBatch(batch *core.Batch, recei
 			if relevant {
 				relevantLogsForSub = append(relevantLogsForSub, logItem)
 			}
-			s.logger.Info(fmt.Sprintf("Subscription %s. Account %s. Log %v. Extracted addresses: %v. Relevant: %t", id, sub.Account, logItem, userAddrs, relevant))
```
tudor-malene, line 154: why is this gone?
This log is super useful.

go/host/enclave/guardian.go
```diff
@@ -198,7 +198,6 @@ func (g *Guardian) mainLoop() {
 		// check enclave status on every loop (this will happen whenever we hit an error while trying to resolve a state,
 		// or after the monitoring interval if we are healthy)
 		g.checkEnclaveStatus()
-		g.logger.Trace("mainLoop - enclave status", "status", g.state.GetStatus())
```
tudor-malene, line 201: @BedrockSquirrel , I assume this log is somewhat useful

go/host/p2p/p2p.go
```diff
@@ -356,12 +355,15 @@ func (p *Service) broadcast(msg message) error {
 	copy(currentAddresses, p.peerAddresses)
 	p.peerAddressesMutex.RUnlock()
 
-	var wg sync.WaitGroup
 	for _, address := range currentAddresses {
-		wg.Add(1)
-		go p.sendBytesWithRetry(&wg, address, msgEncoded) //nolint: errcheck
+		closureAddr := address
```
tudor-malene, line 359: what was the purpose of the WaitGroup?

go/host/p2p/no_inbound_p2p.go
```diff
@@ -0,0 +1,263 @@
...
+	"github.com/obscuronet/go-obscuro/go/common/host"
+	"github.com/obscuronet/go-obscuro/go/common/log"
+	"github.com/obscuronet/go-obscuro/go/common/retry"
+	"github.com/obscuronet/go-obscuro/go/common/subscription"
+	"github.com/obscuronet/go-obscuro/go/config"
+
+	gethlog "github.com/ethereum/go-ethereum/log"
+)
+
+type NoInboundP2P struct {
```
tudor-malene, line 24: how much of this is duplicated compared to the normal implementation?
If a lot, then it would make sense to initialise that with a flag

tudor-malene (changes requested):
looks good in principle.
A couple of questions.

go/host/enclave/guardian.go
BedrockSquirrel: Yeah very useful, and only trace level

go/enclave/events/subscription_manager.go
otherview: Good point, putting it back in.
Seemed like the log was spinning before fixing the p2p thing.
image

go/host/enclave/guardian.go
otherview: Same thing, tks for the find, putting it back in.
Seemed like the log was spinning before fixing the p2p thing.
image

go/host/p2p/p2p.go
otherview: I believe it was to issue the payload to all nodes in parallel and wait until the whole broadcast was finished.
With a 2 min retry and one node with p2p off, this had a weird effect on the sequencer which would not update the host db in proper time. 

When running the sim with this code, all nodes (host db) have height ~80 but the sequencer (host db) stays at high ~60

go/host/p2p/no_inbound_p2p.go
otherview: Yeah, I like using a flag, did that in the p2p in-memory; was keeping them separate to make sure I wasn't introducing any bugs in the existing p2p. 

I'll push all under a flag tomorrow 👍

tudor-malene (approved):
lgtm

PR #1508, opened 2023-09-11
hardhat: ensure token script doesn't spin when waiting

BedrockSquirrel:
### Why this change is needed

Small bug in the sleep in this TS script means that we spin and create lots of promises instead of waiting 1sec between each one.

### PR checks pre-merging


contracts/deployment_scripts/testnet/layer2/001_whitelist_tokens.ts
```diff
@@ -95,8 +95,7 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
         setTimeout(fail, 30_000)
         const messageBusContract = (await hre.ethers.getContractAt('MessageBus', '0x526c84529b2b8c11f57d93d3f5537aca3aecef9b'));
         while (await messageBusContract.callStatic.verifyMessageFinalized(messages[1]) != true) {
-            console.log(`Messages not stored on L2 yet, retrying...`);
```
otherview, line 98: Perhaps it's better to leave the log to see what's happening no ?

If we're tweaking this - we shouldn't do forever loops. What if this had an exit/fail condition ?

otherview (approved):

PR #1514, opened 2023-09-13
Obscuro Gateway Database Deploy

otherview:
### Why this change is needed

https://github.com/obscuronet/obscuro-internal/issues/2158

Adds a deployment for mariadb into a vm.

### PR checks pre-merging

### DIFF
--- a/.github/workflows/manual-deploy-obscuro-gateway-database.yml
+++ b/.github/workflows/manual-deploy-obscuro-gateway-database.yml
@@ -1,0 +1,120 @@
+# Deploys Obscuro Gateway's Database on Azure for Testnet
+# Starts the Obscuro Gateway's database on Azure VM
+
+name: '[M] Deploy Obscuro Gateway Database'
+run-name: '[M] Deploy Obscuro Gateway Database ( ${{ github.event.inputs.testnet_type }} )'
+on:
+  workflow_dispatch:
+    inputs:
+      testnet_type:
+        description: 'Testnet Type'
+        required: true
+        default: 'dev-testnet'
+        type: choice
+        options:
+          - 'dev-testnet'
+          - 'testnet'
+
+jobs:
+  build-and-deploy:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/manual-deploy-obscuro-gateway-database.yml
```diff
@@ -0,0 +1,120 @@
...
+          inlineScript: |
+            az vm run-command invoke -g Testnet -n "${{env.RESOURCE_STARTING_NAME}}-OG-MariaDB-${{ GITHUB.RUN_NUMBER }}"  \
+            --command-id RunShellScript \
+            --scripts 'mkdir -p /home/obscuro \
+            && sudo apt-get update \
+            && sudo apt-get install -y gcc \
+            && sudo snap refresh && sudo snap install --channel=1.18 go --classic \
+            && curl -fsSL https://get.docker.com -o get-docker.sh && sh ./get-docker.sh \
+            && git clone --depth 1 -b ${{ env.BRANCH_NAME }} https://github.com/obscuronet/go-obscuro.git /home/obscuro/go-obscuro \
+            && docker network create --driver bridge node_network || true \
```
zkokelj, line 99: `|| true ` suppresses any errors. Is there any purpose for doing it?
I guess if any of the steps fail the deployment will not be sucessful

zkokelj (approved):
LGTM!  Minor comment on suppressing errors.

.github/workflows/manual-deploy-obscuro-gateway-database.yml
otherview: Good point. 
It does suppress errors but only for that if clause, so if other clauses fail then the command fails. 

We want to forcibly create a node_network so that if we do create more containers they can intra-container communicate between them. 

It's more of a standard, everywhere that we create docker containers we create the `node-network` (it then guarantees that all containers can communicate between them). Not sure if it's worth removing. Will remove it if you feel strongly about it.

PR #1525, opened 2023-09-14
Ziga/og fix frontend issues from workshop

zkokelj:
### Why this change is needed

In this PR I fixed a few frontend bugs and added new features:

- support for mobile (https://github.com/obscuronet/obscuro-internal/issues/2138)
- use getStorageAt instead of local storage to get userID of an user (https://github.com/obscuronet/obscuro-internal/issues/2120)
- show user ID (part of https://github.com/obscuronet/obscuro-internal/issues/2133)
- show all accounts from Metamask and display which of them are connected to the gateway

While doing this I fixed 2 bugs on OG backend:
- use address 0x0...0 for getting UserID from getStorageAt (using anything that is not an address results in Metamask error)
- allow Gateway to use defaultUser for the requests before / after joining (to avoid Metamask "infinite" loading while trying to get `eth_blockNumber` - especially problematic after revoking as we are calling it with userID which is not valid anymore, but is still in the rpcURL in Metamask)

### PR checks pre-merging


tools/walletextension/common/constants.go
```diff
@@ -27,28 +27,28 @@ const (
...
+	PathHealth                       = "/health/"
+	WSProtocol                       = "ws://"
+	DefaultUser                      = "defaultUser"
+	UserQueryParameter               = "u"
+	AddressQueryParameter            = "a"
+	MessageFormatRegex               = `^Register\s(\w+)\sfor\s(\w+)$`
+	MessageUserIDLen                 = 64
+	SignatureLen                     = 65
+	PersonalSignMessagePrefix        = "\x19Ethereum Signed Message:\n%d%s"
+	GetStorageAtUserIDRequestAddress = "0x0000000000000000000000000000000000000000"
```
otherview, line 49: Why change the string to a hex ?

tools/walletextension/test/wallet_extension_test.go
```diff
@@ -295,8 +295,11 @@ func TestGetStorageAtForReturningUserID(t *testing.T) {
 	respJoin := makeHTTPEthJSONReqWithPath(walletHTTPPort, "v1/join")
 	userID := string(respJoin)
 
+	specialAddressToGetUserID := "0x0000000000000000000000000000000000000000"
+	randomAddress := "0x123ABC0000000000000000000000000000000000"
```
otherview, line 299: Why use`specialAddressToGetUserID` instead of a string ?

tools/walletextension/wallet_extension.go
```diff
@@ -307,7 +307,27 @@ func (w *WalletExtension) DeleteUser(hexUserID string) error {
...
+	if err != nil {
+		w.Logger().Error(fmt.Errorf("error getting user's private key (%s), %w", hexUserID, err).Error())
+		return false
+	}
+
+	if len(key) > 0 {
+		return true
+	}
+
+	return false
```
otherview, line 327:
```suggestion
	return  len(key) > 0
```

tools/walletextension/common/constants.go
zkokelj: Before the change we were using `getUserID` as an address for getting userID with getStorageAt function.
Metamask (and probably also some other tools) check this value and return error if it is not an address, that is why I decided to use GetStorageAt with address `0x0...0` as an address to get userID

tools/walletextension/test/wallet_extension_test.go
zkokelj: see the answer above + decided to make a variable and not hardcode it in all cases separately

tools/walletextension/wallet_extension.go
zkokelj: fixed, thanks

tudor-malene (approved):
lgtm

tools/walletextension/common/constants.go
otherview: > Metamask (and probably also some other tools) check this value and return error if it is not an address,

Did you get an error ? Can you post it, I'm wondering bc I didn't get one when implementing this for `listPersonalTransactions`

otherview: Can you try `walletStore.provider.send` instead ?
Ie. in obscuro scan we use `const personalTxData = await walletStore.provider.send('eth_getStorageAt', ["listPersonalTransactions", requestPayload, null])`

zkokelj: The error above is actually recent bug in metamask (https://github.com/MetaMask/metamask-extension/issues/20847).
In another browser I have following error: 
image

And it appears also if I use:
`const provider = window.ethereum;

    // Define the parameters
    const params = [
        'getUserID',  // Replace with the contract address
        getRandomIntAsString(0, 1000)
    ];

    // Create JSON-RPC payload
    const payload = {
        method: 'eth_getStorageAt',
        params: params,
        id: new Date().getTime(),  // Unique id for the request
        jsonrpc: '2.0'
    };

    // Use `sendAsync` to send the JSON-RPC request
    provider.sendAsync(payload, (error, result) => {
        if (error) {
            console.error('Error:', error);
        } else if (result.error) {
            console.error('Result Error:', result.error);
        } else {
            console.log('Storage value:', result.result);
        }
    });`
    
    
    `.send` method is/will be deprecated soon: https://eips.ethereum.org/EIPS/eip-1193#send-deprecated

otherview: Perhaps that error can be from the provider usage? Is this using `ethers`? 

Wonder if using something like we did on obscuro_scan would help https://github.com/obscuronet/go-obscuro/blob/f4913937b1f876aebb26eb915134a6a76856372c/tools/obscuroscan_v2/frontend/src/components/MetaMaskConnectButton.vue ?

> .send method is/will be deprecated soon:
> https://eips.ethereum.org/EIPS/eip-1193#send-deprecated

Yeah I think it's deprecated for external providers, one should use sendAsync, but it should be fine to use with ethers.
image

zkokelj: fixed + reverted back to getUserID

otherview (commented):
Minor bits. 

One question, perhaps for a different PR. How does the connect to metamask process happen ? Typically there's a button for it right ?

tools/walletextension/api/staticOG/javascript.js
```diff
@@ -108,39 +108,111 @@ async function revokeUserID(userID) {
...
+    max = Math.floor(max);
+    const randomInt = Math.floor(Math.random() * (max - min + 1)) + min;
+    return randomInt.toString();
+}
+
+
+async function getUserID() {
+    const provider = new ethers.providers.Web3Provider(window.ethereum);
+    const userID = await provider.send('eth_getStorageAt', ["getUserID", getRandomIntAsString(0, 1000)])
+    return userID
```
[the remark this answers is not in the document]
zkokelj: Will do a test, but if I was using the same value as the second parameter the request was not sent on every refresh (making the UI not working properly some of the time - until it decided to make another request).
I was looking at request and they were not coming every time they we called it from javascript. I guess there is/(was with previous method) some caching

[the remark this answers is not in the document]
zkokelj: Tested again. Setting second parameter to static value doesn't make a request every time it should.
Here is an example - I refreshed page multiple times (javascript always says request sent), but the gateway received a `eth_getStorageAt` request only once & on other refreshes it receives only `net_version` request.
This makes the gateway frontend sometimes using userID that is not valid anymore (for example if I revoke userID and refresh a page.. it still shows me same page as before revoking)

```
received a request:  eth_getStorageAt
received a request:  net_version
received a request:  net_version
received a request:  net_version
received a request:  net_version
received a request:  net_version
received a request:  net_version
received a request:  net_version
received a request:  net_version
received a request:  net_version
received a request:  net_version
received a request:  net_version
```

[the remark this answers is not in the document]
zkokelj: I agree it is not an ideal solution, but the fastest one in our case.
(Setting a timeout & cache timeout  in Web3Provider is not possible: https://github.com/ethers-io/ethers.js/discussions/2349)

tools/walletextension/api/staticOG/javascript.js
```diff
@@ -108,39 +108,111 @@ async function revokeUserID(userID) {
...
+}
+
+async function requestAccounts() {
+    try {
+        const accounts = await ethereum.request({ method: 'eth_requestAccounts' });
+        return accounts[0]; // Currently selected account
```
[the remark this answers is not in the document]
zkokelj: fixed.

tools/walletextension/api/staticOG/javascript.js
```diff
@@ -108,39 +108,111 @@ async function revokeUserID(userID) {
...
+    } else {
+        const statusArea = document.getElementById(idStatus);
+        statusArea.innerText = 'Please install MetaMask to use Obscuro Gateway.';
+    }
+}
+
+async function populateAccountsTable(document, tableBody, userID) {
+    tableBody.innerHTML = '';
+    const accounts = await ethereum.request({method: 'eth_requestAccounts'});
+    for (const account of accounts) {
```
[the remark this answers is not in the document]
zkokelj: ethers provider used now

zkokelj:
Current version is not working in Google Chrome (changing `getUserID` to `0x0000000000000000000000000000000000000000` makes it work).

Even using ethers we get error: 
image
which makes gateway useless

tools/walletextension/api/staticOG/javascript.js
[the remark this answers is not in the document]
otherview: Sure that's fine 👍  Good investigating I would do the same

otherview:
> Current version is not working in Google Chrome (changing `getUserID` to `0x0000000000000000000000000000000000000000` makes it work).
> 
> Even using ethers we get error: image which makes gateway useless

The issue is a bit hidden to be fair. It comes from infura / an rpc response, not from Metamask.
If Obscuroscan uses this method while Metamask is connected to Ethereum it gets the same response.

A try-catch seems to do the trick.
```
async function getUserID() {
    try {
        return await provider.send('eth_getStorageAt', ["getUserID", getRandomIntAsString(0, 1000), null])
    }catch (e) {
        console.log(e)
        return null;
    }
}
```

otherview (approved):

PR #1537, opened 2023-09-20
Using .env files for centralizing github enviroment variables

otherview:
### Why this change is needed

As per title, takes @BedrockSquirrel PR and uses a .env file to load the variables.
Now variables and values are in the code, only secrets are in GH.


BedrockSquirrel:
As discussed in discord, I'm not against this but not clear to me that having these checked in is any better than having them in the GH envs.

Also if you want to keep the URLs non-secret then you'll need to find a way to provide the sepolia ones differently so bit of work still required.

So I'd suggest parking this for now and revisiting soon as part of a wider config rethink (but I'm not against it going in if others think it is an improvement worth making).

otherview:
Typically I choose to have as much as possible of code and configs in the codebase instead of in managed services.
I can see pros and cons of using GH Environment Variables and Environment Secrets vs using .env files + 'env'_named secrets. But no deciding factor tbh.

General consensus is to use GH Environment Variables and Environment Secrets, we'll try that and can always change if needed 👏

PR #1544, opened 2023-09-21
Tweaks to obs gw

otherview:
### Why this change is needed

- Enforce a more sensitive max password attempts 
- Adding more info to User log handler


otherview:
deferred to https://github.com/obscuronet/go-obscuro/pull/1549

PR #1546, opened 2023-09-21
Add flag to configure L1 chainID for docker nodes

BedrockSquirrel:
### Why this change is needed

We need to wire sepolia chain ID from the github action script but it's not a CLI option at that level :(

### PR checks pre-merging

### DIFF
--- a/go/node/config.go
+++ b/go/node/config.go
@@ -55,7 +55,11 @@ type Config struct {
 }
 
 func NewNodeConfig(opts ...Option) *Config {
-	defaultConfig := &Config{}
+	defaultConfig := &Config{
+		batchInterval:  "1s",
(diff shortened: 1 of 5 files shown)


go/node/config.go
```diff
@@ -51,10 +51,15 @@ type Config struct {
 	l1BlockTime               time.Duration
 	batchInterval             string
 	rollupInterval            string
+	l1ChainID                 int
 }
 
 func NewNodeConfig(opts ...Option) *Config {
-	defaultConfig := &Config{}
+	defaultConfig := &Config{
```
BedrockSquirrel, line 58: I finally figured out how the defaults should be set for this level of wiring.

PR #1563, opened 2023-09-28
Setup faucet for sepolia env

BedrockSquirrel:
### Why this change is needed

- Sepolia testing is easier with a faucet.
- We can probably use this faucet for early partners and users too.
- The testnet deploy was showing red for sepolia because of the faucet step

Add config

### PR checks pre-merging

### DIFF
--- a/tools/faucet/cmd/cli.go
+++ b/tools/faucet/cmd/cli.go
@@ -2,8 +2,9 @@ package main
 
 import (
 	"flag"
-	"github.com/ethereum/go-ethereum/params"
 	"math/big"
+
+	"github.com/ethereum/go-ethereum/params"
 
 	"github.com/obscuronet/go-obscuro/tools/faucet/faucet"
 )
--- a/tools/faucet/cmd/cli.go
+++ b/tools/faucet/cmd/cli.go
@@ -32,6 +32,10 @@ const (
 	serverPortName    = "serverPort"
 	serverPortDefault = 80
 	serverPortUsage   = "Port where the web server binds to"
+
+	defaultAmountName    = "defaultAmount"
+	defaultAmountDefault = 100.0
+	defaultAmountUsage   = "Default amount of token to fund (in ETH)"
 )
 
 func parseCLIArgs() *faucet.Config {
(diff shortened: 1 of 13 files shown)


tools/faucet/cmd/cli.go
```diff
@@ -32,6 +32,10 @@ const (
 	serverPortName    = "serverPort"
 	serverPortDefault = 80
 	serverPortUsage   = "Port where the web server binds to"
+
+	defaultAmountName    = "defaultAmount"
+	defaultAmountDefault = 100.0
```
otherview, line 37: float ?

tools/faucet/cmd/cli.go
```diff
@@ -41,15 +45,25 @@ func parseCLIArgs() *faucet.Config {
 	faucetPK := flag.String(faucetPKName, faucetPKDefault, faucetPKUsage)
 	jwtSecret := flag.String(jwtSecretName, jwtSecretDefault, jwtSecretUsage)
 	serverPort := flag.Int(serverPortName, serverPortDefault, serverPortUsage)
+	defaultAmount := flag.Float64(defaultAmountName, defaultAmountDefault, defaultAmountUsage)
```
otherview, line 48: Any need for it to be a float ?

tools/faucet/cmd/cli.go
```diff
@@ -41,15 +45,25 @@ func parseCLIArgs() *faucet.Config {
...
 	}
 }
+
+func toWei(amount *float64) *big.Int {
+	amtFloat := new(big.Float).SetFloat64(*amount)
+	weiFloat := new(big.Float).Mul(amtFloat, big.NewFloat(1e18))
+	// don't care about the accuracy here, float should have less than 18 decimal places
+	wei, _ := weiFloat.Int(nil)
+	return wei
+}
```
otherview, line 69: Perhaps `gethcommon.ParseEther(ethValue)` is a better alternative.

tools/faucet/faucet/faucet.go
```diff
@@ -13,18 +13,18 @@ import (
...
 
 const (
-	_timeout       = 60 * time.Second
-	OBXNativeToken = "obx"
-	WrappedOBX     = "wobx"
-	WrappedEth     = "weth"
-	WrappedUSDC    = "usdc"
+	_timeout              = 60 * time.Second
+	NativeToken           = "eth"
+	DeprecatedNativeToken = "obx" // leaving this in temporarily for tooling that is getting native funds using `/obx` URL
```
otherview, line 24: better to rip the band aid straight off imo

tools/faucet/cmd/cli.go
```diff
@@ -32,6 +32,10 @@ const (
 	serverPortName    = "serverPort"
 	serverPortDefault = 80
 	serverPortUsage   = "Port where the web server binds to"
+
+	defaultAmountName    = "defaultAmount"
+	defaultAmountDefault = 100.0
+	defaultAmountUsage   = "Default amount of token to fund (in ETH)"
```
otherview, line 38: it's a pet annoyance, I'll leave it to you consideration. The flags receive ETH but the config struct receives WEI. Worth normalizing one way or the other imo.

tools/faucet/webserver/web_server.go
```diff
@@ -24,80 +25,38 @@ type requestAddr struct {
...
+	// authed endpoint
+	r.POST("/auth/fund/:token", jwtTokenChecker(jwtSecret, faucetServer.Logger), fundingHandler(faucetServer, defaultAmount))
 
-		// fund the address
-		addr := common.HexToAddress(req.Address)
-		if err := faucetServer.Fund(&addr, token, amount); err != nil {
-			errorHandler(c, fmt.Errorf("unable to fund request %w", err), faucetServer.Logger)
-			return
-		}
+	r.POST("/fund/:token", fundingHandler(faucetServer, defaultAmount))
```
otherview, line 35: Can a comment be added here to take care of this endpoint before deploying to sepolia ?

tools/faucet/cmd/cli.go
BedrockSquirrel: That does sound better but I can't find it. I've changed 1e18 magic number to use geth's `params.Ether` but seems like in geth they mostly just do the multiplication.

tools/faucet/faucet/faucet.go
BedrockSquirrel: As discussed with Moray, we want to make it less painful to transition. No need to break the discord faucet and e2e tests unnecessarily even if it's an easy fix. I'll add a todo to remove it.

tools/faucet/cmd/cli.go
BedrockSquirrel: I feel like we should mostly use explicit big.Int for token values everywhere in code, seems to be mostly what geth does.

But the CLI is a human configured interface, imo it makes sense for it to use human readable values but not to leak that into other layers. If I'm checking the config for each env I don't want to have to count the zeroes ideally.

tools/faucet/cmd/cli.go
otherview: Could swear there was a ParseEther(string), but had a double look and nothing. params.Ether sounds good.

otherview (approved):

PR #1564, opened 2023-09-29
Pedro/fe tweaks

otherview:
### Why this change is needed

https://github.com/obscuronet/obscuro-internal/issues/2221

Making obscuro scan closer to the sentient labs view.

image


zkokelj (approved):
LGTM

PR #1577, opened 2023-10-03
Add balance endpoint to faucet for monitoring

BedrockSquirrel:
### Why this change is needed

We need to be notified when sepolia faucet is getting low on funds but the account balance can't be easily monitored because you need signed viewing key.

Endpoint will let us add it to the periodic monitoring.

### PR checks pre-merging


otherview (approved):
lgtm

PR #1612, opened 2023-10-17
Lock hh-deploy version to fix config compatability bug

BedrockSquirrel:
### Why this change is needed

Locking hardhat-deploy version fixes an error we see: https://github.com/obscuronet/obscuro-internal/issues/2294

### PR checks pre-merging


StefanIliev545 (approved):

otherview:
There are still some deps that aren't locked. Is it worth locking all deps so this doesn't happen again ?

PR #1614, opened 2023-10-20
Cais gateway UI changes

CaisManai:
Revamped the UI and fixed several UI bugs. Only tested on desktop.


tools/walletextension/api/staticOG/index.html
```diff
@@ -11,25 +11,58 @@
...
+    <h2>Welcome to the Obscuro Gateway</h2>
+    <form>
+        <div id="information">A 3 click process to setup encrypted communication between MetaMask and the Obscuro Network.
+            Upon hitting Begin, you'll be asked to:
+            <ol>
+                <li>
+                    Add the Obscuro Testnet network as a new network to MetaMask. Check the Network URL base is https://testnet.obscu.ro and the chain ID is 443 and hit approve  
+                </li>
+                <li>
+                    Allow MetaMask to switch networks to the Obscuro Testnet network
```
zkokelj, line 30: We should probably mention that users need to select which accounts they want to connect? (for the first time users)

tools/walletextension/api/staticOG/javascript.js
```diff
@@ -271,22 +275,35 @@ const initialize = async () => {
...
 
     function displayOnlyJoin() {
         joinButton.style.display = "block"
         revokeUserIDButton.style.display = "none"
         accountsTable.style.display = "none"
+        informationArea.style.display = "block"
+        informationArea2.style.display = "none"
+
+        beginBox.style.visibility = "visible";
+        spinner.style.visibility = "hidden";
```
zkokelj, line 297: When I add accounts and refresh page it only shows us the spinner

tools/walletextension/api/staticOG/javascript.js
0xjba, line 1: Please revert back to the Hard coded gateway address & any usage of it.
Line 1, 14-18, Line 34

0xjba (changes requested):
Please revert back the Hard coded gateway address & any usage of it.
Line 1, 14-18, Line 34

tools/walletextension/api/staticOG/style.css
```diff
@@ -35,11 +203,13 @@ input { display: table-cell; }
 
 /* basic style for the accounts table*/
 table {
```
zkokelj, line 205: Maybe center the table and use same style as the rest of the webpage:

image

In the table we display all the accounts that are connected to metamask (for this URL), even if they are not connected to Obscuro (connected = False), but we don't give an option to the user to add them. Is this ok?

zkokelj (commented):
Nice improvements! 

Mentioned a few things in the comments, but nothing major (not working after refresh is the biggest issue in my opinion).

tools/walletextension/api/staticOG/style.css
CaisManai: It's not ideal, but it's okay for this first version. I'll raise a ticket for it.

PR #1626, opened 2023-10-27
ObscuroScan: load eth price on init

BedrockSquirrel:
### Why this change is needed

Currently it takes one minute before the eth price is requested so we see a loading circle. The other data also has a slight (1 second) delay before it loads because it waits for its polling interval.

Change the poller semantics to make the request immediately when started and then at the interval thereafter.

### PR checks pre-merging

### DIFF
--- a/tools/obscuroscan_v2/frontend/src/stores/priceStore.js
+++ b/tools/obscuroscan_v2/frontend/src/stores/priceStore.js
@@ -24,7 +24,7 @@ export const usePriceStore = defineStore({
             }
         },
 
-        async startPolling() {
+        startPolling() {
             this.poller.start();
         },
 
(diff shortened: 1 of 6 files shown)


tools/obscuroscan_v2/frontend/src/stores/priceStore.js
```diff
@@ -9,7 +9,7 @@ export const usePriceStore = defineStore({
         poller: new Poller(() => {
             const store = usePriceStore();
             store.fetch();
-        }, 60*Config.pollingInterval)
+        }, Config.pricePollingInterval)
```
otherview, line 12: Is there a need to have the price updated at the second ? Also, I think it happened to me getting limit rated for polling too fast, might want to avoid the client limit rated for polling too fast too.

otherview (approved):

PR #1628, opened 2023-10-30
Ziga/og deduplicate events

zkokelj:
### Why this change is needed

After enabling users to subscribe with multiple accounts there are scenarios when users can get duplicated events (one per account instead of one per user). We want to deduplicate them for the users to have similar experience as with Ethereum.

Implementation and usage of circular buffer for deduplication.

### PR checks pre-merging

### DIFF
--- a/tools/walletextension/subscriptions/deduplication_circular_buffer.go
+++ b/tools/walletextension/subscriptions/deduplication_circular_buffer.go
@@ -1,0 +1,43 @@
+package subscriptions
+
+import "github.com/ethereum/go-ethereum/common"
+
+// LogKey uniquely represents a log (consists of BlockHash, TxHash, and Index)
+type LogKey struct {
+	BlockHash common.Hash // Not necessary, but can be helpful in edge case of block reorg.
+	TxHash    common.Hash
+	Index     uint
+}
+
+// CircularBuffer is a data structure that uses a single, fixed-size buffer as if it was connected end-to-end.
+type CircularBuffer struct {
+	data []LogKey
+	size int
+	end  int
+}
+
+// NewCircularBuffer initializes a new CircularBuffer of the given size.
+func NewCircularBuffer(size int) *CircularBuffer {
+	return &CircularBuffer{
+		data: make([]LogKey, size),
+		size: size,
+		end:  0,
+	}
+}
+
+// Push adds a new LogKey to the end of the buffer. If the buffer is full,
+// it overwrites the oldest data with the new LogKey.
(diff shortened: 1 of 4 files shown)


tools/walletextension/subscriptions/subscriptions.go
```diff
@@ -75,13 +74,29 @@ func (sm *SubscriptionManager) HandleNewSubscriptions(clients []rpc.Client, req
...
 func readFromChannelAndWriteToUserConn(channel chan common.IDAndLog, userConn userconn.UserConn, userSubscriptionID gethrpc.ID, logger gethlog.Logger) {
+	buffer := NewCircularBuffer(wecommon.DeduplicationBufferSize)
 	for data := range channel {
 		jsonResponse, err := prepareLogResponse(data, userSubscriptionID)
 		if err != nil {
 			logger.Error("could not marshal log response to JSON on subscription.", log.SubIDKey, data.SubID, log.ErrKey, err)
 			continue
 		}
 
+		// create unique identifier for current log
```
otherview, line 85: Do we need to json.marshal if it turns out to be a duplicate ?

tools/walletextension/subscriptions/deduplication_circular_buffer.go
```diff
@@ -0,0 +1,43 @@
...
+
+// CircularBuffer is a data structure that uses a single, fixed-size buffer as if it was connected end-to-end.
+type CircularBuffer struct {
+	data []LogKey
+	size int
+	end  int
+}
+
+// NewCircularBuffer initializes a new CircularBuffer of the given size.
+func NewCircularBuffer(size int) *CircularBuffer {
```
otherview, line 20: Why a circular buffer versus a limitless slice ?

otherview: What happens to the user experience if the buffer starts removing messages ?

Perhaps it would be simpler from a user experience and system operation if the subscription is knocked down when the buffer reaches a certain size ?

otherview (changes requested):

tools/walletextension/subscriptions/deduplication_circular_buffer.go
otherview: Sorted offline 👍

tools/walletextension/subscriptions/subscriptions.go
zkokelj: changed order

PR #1635, opened 2023-11-04
ObscuroScan: serve oldest batches first

BedrockSquirrel:
### Why this change is needed

We want to show oldest batches first in obscuro scan, very rare that someone would want to look at the first batches of the network.

Also added the batch tx hashes so we can use them for tx count and to link to txs potentially.

### PR checks pre-merging

### DIFF
--- a/go/host/db/batches.go
+++ b/go/host/db/batches.go
@@ -132,8 +132,8 @@ func (db *DB) GetBatchBySequenceNumber(sequenceNumber *big.Int) (*common.ExtBatc
 	return db.GetBatch(*batchHash)
 }
 
-// GetBatchListing returns latest BatchListingResponse given a pagination. For example, page 0, size 10 will return the
-// latest 10 batches.
+// GetBatchListing returns latest batches given a pagination.
+// For example, page 0, size 10 will return the latest 10 batches.
 // todo change this when the db changes - this is not super performant
 func (db *DB) GetBatchListing(pagination *common.QueryPagination) (*common.BatchListingResponse, error) {
 	// fetch the total batches so we can paginate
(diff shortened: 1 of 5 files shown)


go/host/db/batches.go
```diff
@@ -132,25 +132,30 @@ func (db *DB) GetBatchBySequenceNumber(sequenceNumber *big.Int) (*common.ExtBatc
 	return db.GetBatch(*batchHash)
 }
 
-// GetBatchListing returns BatchListingResponse given a pagination
+// GetBatchListing returns latest BatchListingResponse given a pagination. For example, page 0, size 10 will return the
+// latest 10 batches.
```
otherview, line 136: Would it make sense to make all listings return in reverse order and having a comment making that super clear ?

tools/obscuroscan_v2/frontend/src/components/BatchDataGrid.vue
```diff
@@ -47,14 +48,9 @@ export default {
   setup() {
     const store = useBatchStore()
 
-    // Start polling when the component is mounted
+    // Reload batch data onMount
     onMounted(() => {
-      store.startPolling()
```
otherview, line 52: I wonder if this has the desired effect.. It seems like the `RotatingBatchesItem` is also activating the `store.startPolling()`

otherview (approved):

go/host/db/batches.go
BedrockSquirrel: Yeah good point, I'll see if I can make this more general, the rollups should behave the same.

go/host/db/blocks.go
```diff
@@ -74,19 +88,12 @@ func (db *DB) GetBlockListing(pagination *common.QueryPagination) (*common.Block
...
-			listedBlock := common.PublicBlock{BlockHeader: *header}
-			if rollup != nil {
-				listedBlock.RollupHash = rollup.Hash()
-				fmt.Println("added at block: ", header.Number.Int64(), " - ", listedBlock.RollupHash)
-			}
-			blocks = append(blocks, listedBlock)
+		listedBlock := common.PublicBlock{BlockHeader: *header}
+		if rollup != nil {
+			listedBlock.RollupHash = rollup.Hash()
+			fmt.Println("added at block: ", header.Number.Int64(), " - ", listedBlock.RollupHash)
```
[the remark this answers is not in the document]
BedrockSquirrel: Oh yeah good spot, thanks!

PR #1637, opened 2023-11-07
Gateway unsubscribe for multiple subscribed clients

zkokelj:
### Why this change is needed

Since we  (can) have multiple accounts/addresses registered per user and Obscuro Gateway subscribes for all of them. We return  new subscriptionID. And in case users wants to unsubscribe we want to make sure the gateway also unsubscribes from all subscriptions to the node that are/were related to that specific subscriptionID.

### PR checks pre-merging

### DIFF
--- a/tools/walletextension/accountmanager/account_manager.go
+++ b/tools/walletextension/accountmanager/account_manager.go
@@ -81,7 +81,7 @@ func (m *AccountManager) ProxyRequest(rpcReq *wecommon.RPCRequest, rpcResp *inte
 		}
 		subscriptionID, ok := rpcReq.Params[0].(string)
 		if !ok {
-			return fmt.Errorf("subscriptionID needs to be a string. Got: %d", rpcReq.Params[0])
+			return fmt.Errorf("subscriptionID needs to be a string. Got: %v", rpcReq.Params[0])
 		}
 		m.subscriptionsManager.HandleUnsubscribe(subscriptionID, rpcResp)
 		return nil
--- a/tools/walletextension/accountmanager/account_manager.go
+++ b/tools/walletextension/accountmanager/account_manager.go
@@ -61,6 +61,8 @@ func (m *AccountManager) AddClient(address gethcommon.Address, client *rpc.EncRP
 // ProxyRequest tries to identify the correct EncRPCClient to proxy the request to the Obscuro node, or it will attempt
 // the request with all clients until it succeeds
 func (m *AccountManager) ProxyRequest(rpcReq *wecommon.RPCRequest, rpcResp *interface{}, userConn userconn.UserConn) error {
+	// We need to handle a special case for subscribing and unsubscribing from events,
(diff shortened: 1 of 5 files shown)


tools/walletextension/accountmanager/account_manager.go
```diff
@@ -73,6 +75,17 @@ func (m *AccountManager) ProxyRequest(rpcReq *wecommon.RPCRequest, rpcResp *inte
 		}
 		return nil
 	}
+	if rpcReq.Method == rpc.Unsubscribe {
+		if len(rpcReq.Params) != 1 {
+			return fmt.Errorf("one parameter (subscriptionID) expected, %d parameters received", len(rpcReq.Params))
+		}
+		subscriptionID, ok := rpcReq.Params[0].(string)
+		if !ok {
+			return fmt.Errorf("subscriptionID needs to be a string. Got: %d", rpcReq.Params[0])
```
otherview, line 84:
```suggestion
			return fmt.Errorf("subscriptionID needs to be a string. Got: %v", rpcReq.Params[0])
```

tools/walletextension/subscriptions/subscriptions.go
```diff
@@ -159,3 +183,20 @@ func prepareLogResponse(idAndLog common.IDAndLog, userSubscriptionID gethrpc.ID)
 	}
 	return jsonResponse, nil
 }
+
+func (sm *SubscriptionManager) HandleUnsubscribe(userSubscriptionID string, rpcResp *interface{}) {
+	subscriptions, exists := sm.subscriptionMappings[userSubscriptionID]
+
+	if !exists {
```
otherview, line 190:
```suggestion
	subscriptions, exists := sm.subscriptionMappings[userSubscriptionID]
	if !exists {
```

tools/walletextension/subscriptions/subscriptions.go
```diff
@@ -106,39 +100,69 @@ func readFromChannelAndWriteToUserConn(channel chan common.IDAndLog, userConn us
...
+func (sm *SubscriptionManager) unsubscribeAndRemove(userSubscriptionID string, subscription *gethrpc.ClientSubscription) {
+	sm.mu.Lock()
+	defer sm.mu.Unlock()
+
+	subscription.Unsubscribe()
+
+	subscriptions, exists := sm.subscriptionMappings[userSubscriptionID]
+	if !exists {
+		return
+	}
```
otherview, line 112: Is this possible to happen ? that a subscription is cancelled but there isn't one in our mapping ?

otherview (approved):
Minor bits nothing blocking imo

tools/walletextension/subscriptions/subscriptions.go
zkokelj: Should not happen, added it here to be 100% sure not to crash. I added a log that will clearly tell us if & what happened.

PR #1642, opened 2023-11-13
fix test ports

otherview:
### Why this change is needed

Seems to be a collision on the ports assigned to some of the tests

### PR checks pre-merging

### DIFF
--- a/integration/contractdeployer/contract_deployer_test.go
+++ b/integration/contractdeployer/contract_deployer_test.go
@@ -33,7 +33,7 @@ const (
 	erc20ParamThree               = "1000000000000000000"
 	testLogs                      = "../.build/noderunner/"
 	receiptTimeout                = 30 * time.Second // The time to wait for a receipt for a transaction.
-	_portOffset                   = 1000
+	_portOffset                   = 100
 )
 
 func init() { //nolint:gochecknoinits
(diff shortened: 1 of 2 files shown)


BedrockSquirrel (commented):
I agree this looks like a problem because it's using the same port range as the wallet ext unit test constant but I'm not sure this reliably fixes it because both the tests in `contract_deployer_test` call `createObscuroNetwork` with a start port which then has offsets like `DefaultEnclaveOffset` (which are multiples of 100) added to it. So instead of creating two networks from 39000-39999 and 40000-40999, it now does them from 39000-39999 and 39100-40099.

otherview:
@coderabbitai pause

otherview:
> I agree this looks like a problem because it's using the same port range as the wallet ext unit test constant but I'm not sure this reliably fixes it because both the tests in `contract_deployer_test` call `createObscuroNetwork` with a start port which then has offsets like `DefaultEnclaveOffset` (which are multiples of 100) added to it. So instead of creating two networks from 39000-39999 and 40000-40999, it now does them from 39000-39999 and 39100-40099.

Good point - after some digging I removed the `_portOffset` as it's not doing anything because tests inside the same `_test.go` run sequentially by default.

BedrockSquirrel (approved):
lgtm

PR #1653, opened 2023-11-20
Contract deployment: store important addresses

BedrockSquirrel:
### Why this change is needed

When L1Bridge and L1CrossChainMessenger contracts are deployed we want to write their addresses to the mgmt contract so they are stored in an accessible source of truth (useful for automated testing with often short-lived testnets)

- pass mgmt contract address to L2 contract deployment
- call mgmt contract 'SetImportantContractAddress' method after the two deployments mentioned above

### PR checks pre-merging


StefanIliev545 (approved):
LGTM

PR #1663, opened 2023-11-23
Proxy contracts deployment

StefanIliev545:
### Why this change is needed

In order to deploy the contracts as proxies so they can be upgraded.

The deploy plugin is now instructed to deploy the contracts as open zeppelin proxies. The contracts are now using initialize functions instead of constructors.

### PR checks pre-merging


contracts/config/networks.json
```diff
@@ -1,6 +1,6 @@
 {
     "localGeth": {
-      "url": "http://127.0.0.1:37000",
```
BedrockSquirrel, line 3: Did you mean to change this port and the EncRpcUrl one?

BedrockSquirrel (approved):
LGTM if those port changes were meant to be committed, maybe worth double checking all the log messages you put in were meant to stay in too.

Be good to run all the e2e tests for this branch too imo just to make sure it's not broken anything.

contracts/config/networks.json
StefanIliev545: Yep, those are the local testnet ports, someone changed them to something else, but this should be the default state in order to run hardhat stuff locally.

PR #1669, opened 2023-11-28
Contract addresses: Have host fetch L2 Msg Bus address from enclave

BedrockSquirrel:
### Why this change is needed

Unlike other network contract deployments the L2 message bus is deployed by the enclave with a sythetic tx.

Until we have a better method available this change has the host fetching that address from the enclave so it can serve it with the other contract addresses.

### PR checks pre-merging


go/host/host.go
```diff
@@ -200,13 +200,18 @@ func (h *host) HealthCheck() (*hostcommon.HealthCheck, error) {
 
 // ObscuroConfig returns info on the Obscuro network
 func (h *host) ObscuroConfig() (*common.ObscuroNetworkInfo, error) {
+	l2MessageBusAddress, err := h.EnclaveClient().GetL2MessageBusAddress()
+	if err != nil {
+		return nil, responses.ToInternalError(fmt.Errorf("unable to get L2 message bus address - %w", err))
+	}
```
otherview, line 206: Should we have a generic enclave config  method instead of the specific GetL2MessageBusAddress ? Perhaps I'm future-proofing..

otherview (approved):
Minor name change if that makes sense

otherview (approved):

PR #1671, opened 2023-11-28
L1 contracts: increase timeout for extra transactions

BedrockSquirrel:
### Why this change is needed

Testnet deployments are failing because new transactions on the L1 contract deployer (to write important contract addresses to the management contract) have slowed the script down but the timeout wasn't increased.

Bump the allowed time from 1min to 3mins (was taking about 1.5mins on dev-testnet, plus some padding for sepolia being slower).

When things go wrong the container generally fails fast so shouldn't add much time when there are issues.

### PR checks pre-merging


moraygrieve (approved):
LGTM

PR #1674, opened 2023-11-29
Host RPC: use checksum format for contract addresses

BedrockSquirrel:
### Why this change is needed

Moray suggested we should provide contract addresses in the standard checksum format for callers.

Changed the json serialisation of the network config response to use the checksum formatting geth Address type.

I ran the test to verify this doesn't break the RPC client.

### PR checks pre-merging

### DIFF
--- a/go/host/rpc/clientapi/client_api_obscuro.go
+++ b/go/host/rpc/clientapi/client_api_obscuro.go
@@ -1,6 +1,7 @@
 package clientapi
 
 import (
+	gethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ten-protocol/go-ten/go/common"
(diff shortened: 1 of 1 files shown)


tudor-malene (approved):
lgtm

otherview (approved):
Not trying to be pedantic but what's the json-change here ? it seems to only be the EIP55 address change right ? Or did I miss something..

lgtm, perhaps changing `JSONFriendlyObscuroNetworkInfo` to something in the lines of `EIP55ObscuroNetworkInfo` ?

BedrockSquirrel:
Yeah I drew a blank on the name. The change is to make a struct for the response that has the desired json marshalling. I was just thinking if you want another field formatted a different way you don't want to have to rename it. But take your point.

PR #1676, opened 2023-11-30
Contract deployers: configure signer for address publishing

BedrockSquirrel:
### Why this change is needed

Addresses aren't getting written successfully to the L1 after deployments because the signer wasn't being connected to the mgmt contract ethers object.

### PR checks pre-merging


BedrockSquirrel:
Just a note on this PR, I've just had a look at it and tidied it to make it match style of other L1 tx we make.

We are currently blocked by errors like this:
```
Error: cannot estimate gas; transaction may fail or may require manual gas limit [ See: https://links.ethers.org/v5-errors-UNPREDICTABLE_GAS_LIMIT ] (reason="execution reverted: Ownable: caller is not the owner
```
Which are the same issues with the proxy contract blocking recovery of funds. They will be fixed when we fix up the Open Zepplin contracts that we are using.

StefanIliev545 (approved):

PR #1679, opened 2023-12-04
Enclave: no panic for bad rlp format

BedrockSquirrel:
### Why this change is needed

We don't want to fail for badly formatted requests, we want to log the error and allow chance to recover.

Currently we panic and this recently took down the sepolia testnet. (Investigation of the root cause of that is ongoing).

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #1683, opened 2023-12-07
Jennifer/scan updates

Jennievon:
### Why this change is needed

https://github.com/ten-protocol/ten-internal/issues/2559
https://github.com/ten-protocol/ten-internal/issues/2560
https://github.com/ten-protocol/ten-internal/issues/2561
https://github.com/ten-protocol/ten-internal/issues/2562

### PR checks pre-merging

### DIFF
--- a/tools/obscuroscan_v3/frontend/src/components/modules/dashboard/recent-blocks.tsx
+++ b/tools/obscuroscan_v3/frontend/src/components/modules/dashboard/recent-blocks.tsx
@@ -13,7 +13,7 @@ export function RecentBlocks({ blocks }: { blocks: any }) {
           </Avatar>
           <div className="ml-4 space-y-1">
             <p className="text-sm font-medium leading-none">
-              #{block?.blockHeader?.number}
+              #{+block?.blockHeader?.number}
             </p>
             <p className="text-sm text-muted-foreground word-break-all">
               {formatTimeAgo(block?.blockHeader?.timestamp)}
--- a/tools/obscuroscan_v3/frontend/src/lib/constants.ts
+++ b/tools/obscuroscan_v3/frontend/src/lib/constants.ts
@@ -2,6 +2,7 @@ export const socialLinks = {
   github: "https://github.com/ten-protocol",
   discord: "https://discord.gg/2JQ2Z3r",
   twitter: "https://twitter.com/tenprotocol",
+  twitterHandle: "@tenprotocol",
 };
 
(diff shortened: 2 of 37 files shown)


tools/obscuroscan_v3/frontend/src/components/modules/dashboard/recent-blocks.tsx
```diff
@@ -13,7 +13,7 @@ export function RecentBlocks({ blocks }: { blocks: any }) {
           </Avatar>
           <div className="ml-4 space-y-1">
             <p className="text-sm font-medium leading-none">
-              #{block?.blockHeader?.number}
+              #{+block?.blockHeader?.number}
```
BedrockSquirrel, line 16: Is that `+` a typo or some javascript magic?

tools/obscuroscan_v3/frontend/src/lib/constants.ts
```diff
@@ -2,12 +2,31 @@ export const socialLinks = {
...
 export const maxRetries = 3;
 export const pricePollingInterval = 60 * 1000;
 
+export const getOptions = (query: {
+  page?: string | string[];
+  size?: string | string[];
+}) => {
+  const options = {
+    offset: Number.isNaN(parseInt(query.page as string, 10))
+      ? 1
```
BedrockSquirrel, line 18: This should be zero by default, it's causing you to skip the first row.

tools/obscuroscan_v3/frontend/src/lib/constants.ts
```diff
@@ -2,12 +2,31 @@ export const socialLinks = {
...
+      : parseInt(query.page as string, 10),
```
BedrockSquirrel, line 19: And this offset needs to be multiplied by the page size. I know that's a super weird API we should get that fixed but that seems to be how it is atm. So if you're on page 3 and your page size is ten you would send:

offset: 20, size: 10 -> that will give you rows 20-30

BedrockSquirrel (changes requested):
LGTM except a bit of confusion on the pagination API

tools/obscuroscan_v3/frontend/src/components/modules/dashboard/recent-blocks.tsx
Jennievon: It's same as using `{Number(block?.blockHeader?.number)}`

PR #1698, opened 2023-12-14
Enclave: produce no batch if all tx failed

BedrockSquirrel:
### Why this change is needed

Avoid spamming empty batches when there is a rejected tx but no successful transactions.

### PR checks pre-merging


tudor-malene (approved):
lgtm.

this is probably it.
I'm working on getting rid of that excluded txs thing

PR #1701, opened 2023-12-14
Testnet Deployments: set enclave TESTMODE=true

BedrockSquirrel:
### Why this change is needed

Testnet deployments and upgrades are currently failing because the enclave docker image is set to TESTMODE=false (the default).

This means it expects to find required flags from enclave.json but they're not there so it panics.

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #1715, opened 2023-12-21
Ziga/fix gateway spammy logs

zkokelj:
### Why this change is needed

Currently logs in Datadog are spammy and we log some things as errors even if those things are expected to happen or users are calling it with wrong parameters.

- Deleted one spammy log with an explanation
- Changed one Error to Warning (it turns out we probably call one of the endpoints in a wrong way and it results in those logs)

### PR checks pre-merging

### DIFF
--- a/tools/walletextension/wallet_extension.go
+++ b/tools/walletextension/wallet_extension.go
@@ -311,6 +311,7 @@ func (w *WalletExtension) UserExists(hexUserID string) bool {
 
 	// Check if user exists and don't log error if user doesn't exist, because we expect this to happen in case of
 	// user revoking encryption token or using different testnet.
+	// todo add a counter here in the future
 	key, err := w.storage.GetUserPrivateKey(userIDBytes)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


tools/walletextension/wallet_extension.go
```diff
@@ -309,9 +309,10 @@ func (w *WalletExtension) UserExists(hexUserID string) bool {
 		return false
 	}
 
+	// Check if user exists and don't log error if user doesn't exist, because we expect this to happen in case of
+	// user revoking encryption token or using different testnet.
```
otherview, line 313:
```suggestion
	// Check if user exists and don't log error if user doesn't exist, because we expect this to happen in case of
	// user revoking encryption token or using different testnet.
	// todo add a counter here in the future
```

otherview (approved):

zkokelj:
E2E tests passed: https://github.com/ten-protocol/ten-test/actions/runs/7289436275/job/19864034568

PR #1721, opened 2023-12-22
Fix build

otherview:
### Why this change is needed

There are a bunch of latent issues that seem to creep up when the gas limits are added.
I have a few guesses on what might be happening, mainly relating with contract sizes being large, and blocks getting filled faster than the testing is used to. However I think this needs to be more carefully inspected.

This PR addresses the following:
- Re-adds all reverted PRs:
https://github.com/ten-protocol/go-ten/pull/1682
https://github.com/ten-protocol/go-ten/pull/1710
https://github.com/ten-protocol/go-ten/pull/1715

- Adds missing log folders ( faucet, obscuroscan, tengateway, smartcontracts )
- Fixes gas limits on calls on both go code and HardHat
- Bump gas values to 30M ( Re-opened the ticket to reduce max batch size to 2.5M )
- Bump ports - now have a 2k distance between them
- Adds 1 sec delay between Genesis and Batch #2 ( producing 2 batches so close together was leading the synthetic contract deployment to be all screwy leading to complaints on the nonce and then to an unreliable batch state)
- Added `ensureNoDuplicatedNetwork` for geth to ensure no lingering network is running when it shouldnt
- Added `prefundedBalancesActive` making sure the prefunded balances are ok at the end of the eth2network start cycle

### PR checks pre-merging

### DIFF
--- a/go/config/enclave_cli_flags.go
+++ b/go/config/enclave_cli_flags.go
@@ -62,7 +62,7 @@ var EnclaveFlags = map[string]*flag.TenFlag{
 	UseInMemoryDBFlag:             flag.NewBoolFlag(UseInMemoryDBFlag, true, "Whether the enclave will use an in-memory DB rather than persist data"),
 	ProfilerEnabledFlag:           flag.NewBoolFlag(ProfilerEnabledFlag, false, "Runs a profiler instance (Defaults to false)"),
 	DebugNamespaceEnabledFlag:     flag.NewBoolFlag(DebugNamespaceEnabledFlag, false, "Whether the debug namespace is enabled"),
-	GasLocalExecutionCapFlag:      flag.NewUint64Flag(GasLocalExecutionCapFlag, 5_000_000, "Max gas usage when executing local transactions"),
+	GasLocalExecutionCapFlag:      flag.NewUint64Flag(GasLocalExecutionCapFlag, 6_000_000, "Max gas usage when executing local transactions"),
 }
 
 // enclaveRestrictedFlags are the flags that the enclave can receive ONLY over the Ego signed enclave.json
--- a/go/config/enclave_cli_flags.go
+++ b/go/config/enclave_cli_flags.go
@@ -55,14 +55,14 @@ var EnclaveFlags = map[string]*flag.TenFlag{
 	MaxRollupSizeFlag:             flag.NewUint64Flag(MaxRollupSizeFlag, 1024*64, "The maximum size a rollup is allowed to reach"),
 	L2BaseFeeFlag:                 flag.NewUint64Flag(L2BaseFeeFlag, 1, ""),
 	L2CoinbaseFlag:                flag.NewStringFlag(L2CoinbaseFlag, "0xd6C9230053f45F873Cb66D8A02439380a37A4fbF", ""),
-	GasBatchExecutionLimit:        flag.NewUint64Flag(GasBatchExecutionLimit, 10_000_000, "Max gas that can be executed in a single batch"),
+	GasBatchExecutionLimit:        flag.NewUint64Flag(GasBatchExecutionLimit, 30_000_000, "Max gas that can be executed in a single batch"),
 	ObscuroGenesisFlag:            flag.NewStringFlag(ObscuroGenesisFlag, "", "The json string with the obscuro genesis"),
 	L1ChainIDFlag:                 flag.NewInt64Flag(L1ChainIDFlag, 1337, "An integer representing the unique chain id of the Ethereum chain used as an L1 (default 1337)"),
 	ObscuroChainIDFlag:            flag.NewInt64Flag(ObscuroChainIDFlag, 443, "An integer representing the unique chain id of the Obscuro chain (default 443)"),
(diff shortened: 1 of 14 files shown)


otherview:
@coderabbit pause

otherview:
Stabilized Unit tests
image

The red failure is a port re-use.

otherview:
e2e passed at https://github.com/ten-protocol/ten-test/actions/runs/7330683116 and https://github.com/ten-protocol/ten-test/actions/runs/7329849247

With the https://github.com/ten-protocol/ten-test/pull/210 PR

go/config/enclave_cli_flags.go
```diff
@@ -55,14 +55,14 @@ var EnclaveFlags = map[string]*flag.TenFlag{
...
 	ObscuroGenesisFlag:            flag.NewStringFlag(ObscuroGenesisFlag, "", "The json string with the obscuro genesis"),
 	L1ChainIDFlag:                 flag.NewInt64Flag(L1ChainIDFlag, 1337, "An integer representing the unique chain id of the Ethereum chain used as an L1 (default 1337)"),
 	ObscuroChainIDFlag:            flag.NewInt64Flag(ObscuroChainIDFlag, 443, "An integer representing the unique chain id of the Obscuro chain (default 443)"),
 	UseInMemoryDBFlag:             flag.NewBoolFlag(UseInMemoryDBFlag, true, "Whether the enclave will use an in-memory DB rather than persist data"),
 	ProfilerEnabledFlag:           flag.NewBoolFlag(ProfilerEnabledFlag, false, "Runs a profiler instance (Defaults to false)"),
 	DebugNamespaceEnabledFlag:     flag.NewBoolFlag(DebugNamespaceEnabledFlag, false, "Whether the debug namespace is enabled"),
-	GasLocalExecutionCapFlag:      flag.NewUint64Flag(GasLocalExecutionCapFlag, 3_000_000, "Max gas usage when executing local transactions"),
+	GasLocalExecutionCapFlag:      flag.NewUint64Flag(GasLocalExecutionCapFlag, 40_000_000, "Max gas usage when executing local transactions"),
```
zkokelj, line 65: Why is it now set to 40 M, while `GasBatchExecutionLimit` is 30M?

go/enclave/nodetype/sequencer.go
```diff
@@ -163,6 +163,7 @@ func (s *sequencer) createGenesisBatch(block *common.L1Block) error {
 		return err
 	}
 
+	time.Sleep(time.Second)
```
zkokelj, line 166: Maybe just add a comment?
`Adds 1 sec delay between Genesis and Batch`

integration/constants.go
```diff
@@ -2,18 +2,18 @@ package integration
...
-	StartPortSimulationInMem         = 35000
-	StartPortSimulationFullNetwork   = 37000
-	StartPortSmartContractTests      = 38000
-	StartPortContractDeployerTest    = 39000
-	StartPortWalletExtensionUnitTest = 40000
-	StartPortFaucetUnitTest          = 41000
-	StartPortFaucetHTTPUnitTest      = 42000
-	StartPortTenscanUnitTest         = 43000
-	StartPortTenGatewayUnitTest      = 44000
+	StartPortEth2NetworkTests        = 10000
```
zkokelj, line 5: What is the reason we need to change ports? Were there conflicts?

integration/simulation/simulation.go
```diff
@@ -227,7 +227,7 @@ func (s *Simulation) deployObscuroERC20s() {
 
 			deployContractTx := types.DynamicFeeTx{
 				Nonce:     NextNonce(s.ctx, s.RPCHandles, owner),
-				Gas:       2_000_000,
+				Gas:       5_000_000,
```
zkokelj, line 230: What is the reason for increasing gas here, while decreasing it in `func PrefundWallets`?
Were previous estimates way off and we are adjusting to make them more optimised?

zkokelj (approved):
LGTM 👍 

Just minor comments / questions about the logic behind the changes.

go/config/enclave_cli_flags.go
otherview: I think we might have a bug somewhere, but it seems like if the GasLocalExecutionCapFlag <= GasBatchExecutionLimit we get more flaky errors, so I ended up bumping the local execution max gas to more than the batch size.

go/enclave/nodetype/sequencer.go
otherview: Yes ! Thank you, I meant to add that !
```suggestion
        // errors in unit test seem to suggest that batch 2 was received before batch 1
        // this ensures that there is enough gap so that batch 1 is issued before batch 2
	time.Sleep(time.Second)
```

integration/constants.go
otherview: Yes, that's right. There are still conflicts in ports, although less now. I've also added a log to tell us which process is occupying the conflicting port.

integration/simulation/simulation.go
otherview: Just making the gas limit equal everywhere pretty much.
Also, I'm not sure why we're using the DynamicFee tx in that particular tx and didn't want to over tweak it.

StefanIliev545 (approved):

PR #1737, opened 2024-01-07
[Design] Catastrophic enclave failure - Encrypted data recovery

StefanIliev545:
### Why this change is needed

### PR checks pre-merging


BedrockSquirrel:
For me the biggest problem with this is that the value of the brute-forcing to an attacker is hard to predict. We re-introduce some serious MEV-style exploits.

If the value of knowing, for example, a random number outcome in some viral game becomes high then people can pay to discover it then it'll be discovered. And what's more I don't think the network would even know if someone had done this? Whereas with other recovery mechanisms we've talked about it has always been very publicly visible when shenanigans are afoot.

Plus, probably negligible given it would be rare and tiny scale compared to BTC etc., but be a shame to burn a ton of CPU CO2 to recover the network.

That said, I do appreciate any ideas that can help us decentralise some of the protocol stuff, recovery ideas all seem super centralised at the moment.

PR #1738, opened 2024-01-09
Host: Small cache of blocks by hash

BedrockSquirrel:
### Why this change is needed

We are currently quite inefficient with re-requesting blocks when:
- checking for forks on every block
- if the host is busy when a new block comes in, we re-request it when they're ready to process it (exacerbated by multiple enclaves)

This is especially unfortunate for remote L1 data service like infura, there is a cost in latency and it counts against our request quota.

Add a small cache (using simple, popular LRU cache lib) for recent blocks so that the client doesn't go to RPC for the same data unnecessarily.

I did a test with a hit/miss counter to make sure it's getting used, it's at ~50% hit rate during resyncing (less during live streaming).

### PR checks pre-merging

### DIFF
--- a/go/ethadapter/geth_rpc_client.go
+++ b/go/ethadapter/geth_rpc_client.go
@@ -15,7 +15,7 @@ import (
 	"github.com/ethereum/go-ethereum/ethclient"
 	gethlog "github.com/ethereum/go-ethereum/log"
 
-	"github.com/hashicorp/golang-lru/v2"
+	lru "github.com/hashicorp/golang-lru/v2"
 
 	"github.com/ten-protocol/go-ten/contracts/generated/ManagementContract"
 	"github.com/ten-protocol/go-ten/go/common"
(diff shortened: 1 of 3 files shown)


go/ethadapter/geth_rpc_client.go
```diff
@@ -8,37 +8,37 @@ import (
...
+	"github.com/ten-protocol/go-ten/go/common/log"
+	"github.com/ten-protocol/go-ten/go/common/retry"
 )
 
 const (
 	connRetryMaxWait        = 10 * time.Minute // after this duration, we will stop retrying to connect and return the failure
 	connRetryInterval       = 500 * time.Millisecond
 	_maxRetryPriceIncreases = 5
 	_retryPriceMultiplier   = 1.2
+	_defaultBlockCacheSize  = 51 // enough for 50 request batch size and one for previous block
```
BedrockSquirrel, line 31: This is just a finger in the air placeholder, I tried it with 200 and it was fine but doesn't give you many more cache hits so I went for a reasonable minimum.

We can make this configurable at some point if it ever becomes useful.

otherview (approved):

PR #1770, opened 2024-01-31
Logging: use terminal format with full hashes

BedrockSquirrel:
### Why this change is needed

In recent PR I tried to fix the shortened hashes in our logs, to make them searchable in etherscan etc. But actually what I did had no effect on testnet logs and put our sim logs into a different, machine readable format.

We wanted to use the terminal format everywhere but with full hashes.

- Create a Ten log Format that we have full control over
- duplicate the utility classes to make this match geth Terminal format

In this screenshot the new format is above and the old format below. You can see the `L1 transaction successful` message now has searchable hashes:

image

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #1776, opened 2024-02-06
Add estimate optmisation that geth added

BedrockSquirrel:
### Why this change is needed

Our gas estimation method is lifted (copy-paste) from geth code. They added an optimisation to their binary search that means in average case it goes way quicker (e.g. 4 iterations instead of 37 in my testing).

https://github.com/ethereum/go-ethereum/blob/16ce7bf50fa71c907d1dc6504ed32a9161e71351/eth/gasestimator/gasestimator.go#L162 <-- is the code on their codebase.

### PR checks pre-merging


tudor-malene (approved):
lgtm
(I think..)

PR #1779, opened 2024-02-07
Ten Gateway Caching

zkokelj:
### Why this change is needed

We want to cache some results on the Ten Gateway and not forward all the requests to the Ten Node

- New cache added (`ristretto.Cache`  is used)
- Simple interface that allows us to simply change the cache library / use different type of caching in the future
- Per method TTL and logic to determine if methods are cacheable and for how long
- Tests

### PR checks pre-merging

### DIFF
--- a/tools/walletextension/cache/RistrettoCache.go
+++ b/tools/walletextension/cache/RistrettoCache.go
@@ -1,0 +1,54 @@
+package cache
+
+import (
+	"time"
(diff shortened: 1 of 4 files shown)


BedrockSquirrel (approved):
LGTM

tudor-malene (approved):
lgtm

PR #1806, opened 2024-02-20
NetworkTests: add options for test users to use gateway

BedrockSquirrel:
### Why this change is needed

The network tests are useful for end-to-end testing from the IDE against a local network or against the testnet environments.

But they could only test the direct to node 'AuthClient' connections and so they couldn't be used to debug or test GW issues.

- Add `WithGateway()` option to the LocalDevNetwork so it starts a (debuggable) gateway process alongside the network nodes.
- Add `UseGateway` boolean option to the `CreateTestUser` action to allow users to be created that would register and communicate with the gateway instead of directly. (Create User abstraction so both types of user can be used in all tests).
- Fix issue in the gateway where it was expecting params key to always be set even for methods like eth_gasPrice that have no params. This is an issue because the geth client does not send that key for that method.

### PR checks pre-merging

### DIFF
--- a/integration/networktest/actions/context.go
+++ b/integration/networktest/actions/context.go
@@ -16,18 +16,18 @@ var KeyNumberOfTestUsers = ActionKey("numberOfTestUsers")
 // ActionKey is the type for all test data stored in the context. Go documentation recommends using a typed key rather than string to avoid conflicts.
 type ActionKey string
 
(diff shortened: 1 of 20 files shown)


zkokelj (approved):
LGTM, nice job.. very helpful test!👍

PR #1810, opened 2024-02-21
fix mutex error

zkokelj:
### Why this change is needed

We experienced crashes on testnet and this is the most likely cause of them.
We were using only read mutex, but we also need write mutex to avoid concurrent write

Change read mutex to read+write mutex

### PR checks pre-merging


BedrockSquirrel (approved):
lgtm

PR #1813, opened 2024-02-21
create a shallow copy of a map before modifying it

zkokelj:
### Why this change is needed

The gateway was crashing because of modifying of requestID values after returning from cache

creating shallow copy of return value before modifying and returning it

### PR checks pre-merging


BedrockSquirrel (approved):
LGTM when you fix the return value we discussed.

PR #1828, opened 2024-03-04
store incoming transactions in a database

zkokelj:
### Why this change is needed

For debugging purposes we would like to store unencrypted transactions in the gateway to be able to reproduce some errors if they occur.

Additional database table, storing transactions

### PR checks pre-merging

- [X] PR checks reviewed and performed 

https://github.com/ten-protocol/ten-test/actions/runs/8143119529

### DIFF
--- a/tools/walletextension/storage/database/mariadb/mariadb.go
+++ b/tools/walletextension/storage/database/mariadb/mariadb.go
@@ -4,9 +4,10 @@ import (
 	"database/sql"
 	"encoding/hex"
 	"fmt"
-	"github.com/ethereum/go-ethereum/crypto"
 	"path/filepath"
 	"runtime"
+
+	"github.com/ethereum/go-ethereum/crypto"
 
 	_ "github.com/go-sql-driver/mysql" // Importing MariaDB driver
 	"github.com/ten-protocol/go-ten/go/common/errutil"
@@ -153,9 +154,13 @@ func (m *MariaDB) StoreTransaction(rawTx string, userID []byte) error {
 	defer stmt.Close()
 
 	// Calculate tx hash
+	if len(rawTx) < 3 {
+		fmt.Println("Invalid rawTx: ", rawTx)
+		return nil
+	}
 	rawTxBytes, err := hex.DecodeString(rawTx[2:])
 	if err != nil {
-		panic(err)
+		fmt.Println("Error decoding rawTx: ", rawTx)
 	}
 	txHash := crypto.Keccak256Hash(rawTxBytes)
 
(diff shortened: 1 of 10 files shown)


BedrockSquirrel (approved):
LGTM

tools/walletextension/wallet_extension.go
```diff
@@ -142,6 +142,20 @@ func (w *WalletExtension) ProxyEthRequest(request *common.RPCRequest, conn userc
...
 	}
 
+	// check if user is sending a new transaction and if we should store it in the database for debugging purposes
+	if request.Method == rpc.SendRawTransaction && w.config.StoreIncomingTxs {
+		userIDBytes, err := common.GetUserIDbyte(hexUserID)
+		if err != nil {
+			w.Logger().Error(fmt.Errorf("error decoding string (%s), %w", hexUserID[2:], err).Error())
+			return nil, errors.New("error decoding userID. It should be in hex format")
+		}
+		err = w.storage.StoreTransaction(request.Params[0].(string), userIDBytes)
```
BedrockSquirrel, line 152: Are we 100% certain that the first param exists, can be cast to a string and will have "0x" as the first two characters so you can strip them off?

tools/walletextension/storage/database/mariadb/mariadb.go
```diff
@@ -142,3 +144,25 @@ func (m *MariaDB) GetAllUsers() ([]common.UserDB, error) {
...
+	stmt, err := m.db.Prepare("INSERT INTO transactions(user_id, tx_hash, tx) VALUES (?, ?, ?)")
+	if err != nil {
+		return err
+	}
+	defer stmt.Close()
+
+	// Calculate tx hash
+	rawTxBytes, err := hex.DecodeString(rawTx[2:])
+	if err != nil {
+		panic(err)
```
BedrockSquirrel, line 158: We shouldn't panic here imo, just log the error and store it without the tx hash. (Same on the sqlite impl)

BedrockSquirrel (changes requested):
Couple minor tweaks/things I wanted to check.

tools/walletextension/storage/database/mariadb/mariadb.go
zkokelj: fixed, good catch!

tools/walletextension/wallet_extension.go
zkokelj: Things are checked in a function that calls that function, but added additional check there to avoid an edge case that might happen.

BedrockSquirrel (approved):
lgtm

PR #1834, opened 2024-03-14
Headers: use 65-byte signatures instead of R,S values

BedrockSquirrel:
### Why this change is needed

We want to be able to easily recover Enclave ID from batch and rollup header signatures.

Put signature logic into a small library. Replace R,S values in the batch/rollup headers with the 65-byte signature format.

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #1837, opened 2024-03-19
Display errors if other wallets are also installed alongside MetaMask

zkokelj:
### Why this change is needed

Some users experienced problems with our gateway when they have installed and enabled some other wallets.
Nest and Exodus wallets were the two identified as problematic. 
The problem is that all those wallets inject `window.ethereum` object and can override the one from Metamask.
Giving users clear message about that is very helpful for them.

This is short term solution until we decide to use walletconnect or something similar.

image

### PR checks pre-merging


tools/walletextension/frontend/src/services/ethService.ts
```diff
@@ -14,6 +14,14 @@ const ethService = {
...
     try {
       if (ethereum) {
+        if (ethereum.isExodus) {
+          showToast(ToastType.DESTRUCTIVE, "Exodus Wallet is connected and is conflicting with MetaMask. Please disable Exodus Wallet and try again.");
+          throw new Error("Exodus Wallet is connected and is conflicting with MetaMask. Please disable Exodus Wallet and try again.");
+        } else if (ethereum.isNestWallet) {
+          showToast(ToastType.DESTRUCTIVE, "Nest Wallet is connected and is conflicting with MetaMask. Please disable Nest Wallet and try again.");
+          throw new Error("Nest Wallet is connected and is conflicting with MetaMask. Please disable Nest Wallet and try again.");
+        }
+
```
Jennievon, line 24: @zkokelj LGTM. However, is this issue isolated to these wallets only? We might want to consider creating a map that associates each wallet's name with its key. For example, `ethereum.isExodus` could correspond to 'Exodus Wallet'. Then, we can loop through the map and display the error message accordingly.

Jennievon (approved):
Looks good to me, @zkokelj!

PR #1839, opened 2024-03-19
Host L1 txs: one-at-a-time to remove nonce risk

BedrockSquirrel:
### Why this change is needed

Currently we try to optimistically manage L1 account nonce on the host. We only resync it to chain data on restart.

This has recently started to cause a conflict on ten-sepolia where rollup tx start failing continuously because the nonce is off.

Currently the only L1 transactions produced by the host are rollups and secret responses, there's no need to have multiple in flight simultaneously. Both the rollups and the secret exchange protocols can recover from delayed or dropped transactions.

Limit host to one L1 tx in flight at a time, blocking for the next one. This only adds contention at testnet network initializations really, when secret exchanges can be happening simultaneously as nodes start up. 

The sim tests currently start the most nodes and we see a 10second increase in that test length with this change which seems reasonable for the simplification.

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #1847, opened 2024-03-20
Add security headers

Jennievon:
### Why this change is needed

https://github.com/ten-protocol/ten-internal/issues/3140
Add security headers as proposed by Beagle Security mitigation techniques

- set `X-Frame-Options` to `deny` to prevent the page from being loaded in an iFrame
-  set `X-Content-Type-Options` to `nosniff` to indicate the browser shouldn't interpret the response as something other than the specified content type

### PR checks pre-merging


zkokelj (approved):
LGTM!
Maybe worth checking if we need to include also some other security related tags with this PR

PR #1849, opened 2024-03-21
Network tests: refactor config to allow for HA

BedrockSquirrel:
### Why this change is needed

First tests for multi-enclave nodes will be run from network tests.

The config was a bit confusing, and wanted options to be more easily creatable for things like number of nodes, number of enclaves etc. so simplified it a bit.

Also did some Obscuro -> Ten renames where I saw them.

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #1851, opened 2024-03-25
remove legacy signature type

zkokelj:
### Why this change is needed

I already removed old wallet extension endpoints from the gateway. Now we only use and produce new signatures (EIP712 and PersonalSign). There is no need to support old legacy format anymore and it just adds additional complexity to our codebase. 

Removed old legacy format for signed viewing keys.

### PR checks pre-merging

### DIFF
--- a/go/common/viewingkey/viewing_key.go
+++ b/go/common/viewingkey/viewing_key.go
@@ -5,7 +5,6 @@ import (
 	"encoding/hex"
 	"fmt"
 
-	"github.com/ethereum/go-ethereum/accounts"
 	gethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/crypto/ecies"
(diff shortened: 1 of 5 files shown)


go/common/viewingkey/viewing_key.go
```diff
@@ -33,41 +32,45 @@ type RPCSignedViewingKey struct {
 }
 
 // GenerateViewingKeyForWallet takes an account wallet, generates a viewing key and signs the key with the acc's private key
-// uses the same method of signature handling as Metamask/geth
-// TODO @Ziga - update this method to use the new EIP-712 signature format / personal sign after the removal of the legacy format
 func GenerateViewingKeyForWallet(wal wallet.Wallet) (*ViewingKey, error) {
-	// generate an ECDSA key pair to encrypt sensitive communications with the obscuro enclave
-	vk, err := crypto.GenerateKey()
+	chainID := int64(443)
```
BedrockSquirrel, line 36: I think wallet has a chainID field on it, might be better to use that rather than hardcoding if poss. In case we end up with different chainIDs on diff Ten networks eventually.

BedrockSquirrel (approved):
LGTM, but probably better to avoid hardcoded chain ID if poss.

go/common/viewingkey/viewing_key.go
zkokelj: nice, fixed 👍

PR #1852, opened 2024-03-26
HA: Support multiple enclaves on a single host

BedrockSquirrel:
### Why this change is needed

We want the host to be able to failover if an enclave fails. This will be especially important for a highly available sequencer.

As a preliminary step we need hosts to be able to interact with multiple enclaves and keep them up-to-date.

- Change host config to receive comma-seperated list of enclave RPC addresses instead of a single one
- Create a guardian per enclave, only the first one is a 'sequencer' guardian for now (no support for failing over sequencer functionality yet, and no support for making RPC requests to the secondary enclave).
- Add a basic test that brings up the network with this double-enclave sequencer and checks it is functional 
- This tests basic functionality like the sequencer was able to perform the protocol secret handshake for both of its enclaves and keep them both up to date.

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #1857, opened 2024-03-29
Ziga/fix gateway frontend issues after geth rpc

zkokelj:
### Why this change is needed

After merging PR with Geth RPC there were some frontend issues. But we decided to fix them in this PR.

- fixed the response of the `/network-health` endpoint (before this endpoint was not used and request was forwarded to ethereumRPC handler which forwarded the response to the node. After implementing geth RPC we don't have that endpoint and this endpoint was actually called. I decided to return the response in the same format as before).

- fixed some differences regarding encryptionToken and `0x` prefix that emerged in that PR.

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #1860, opened 2024-04-02
Host: notify batch subscribers of new batches from any src

BedrockSquirrel:
### Why this change is needed

The L2 batch repo only notifies subscribers of new batches from p2p source. This was convenient but confusing.

It is more re-usable if it notifies on every batch that is new to the host, so it will work for sequencers.

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #1869, opened 2024-04-03
Fix parsing both tokens with 0x prefix and without it

zkokelj:
### Why this change is needed

This is a temporary fix that enables local testnet again.
In a separate PR I will provide full solution with consistent usage of `0x` prefix and fixed gateway issues.

Local testnet was successfully started with this branch: https://github.com/ten-protocol/ten-test/actions/runs/8543430314/job/23407243846

 accept both tokens with `0x` prefix and without it.

### PR checks pre-merging


BedrockSquirrel (approved):
LGTM

PR #1876, opened 2024-04-12
Ziga/add gateway reverse proxy

zkokelj:
### Why this change is needed

After RPC changes website and static files are served from `/static`. We want to keep serving website from root `/` as this is much more user friendly and keep API endpoints untouched at `/v1`.

- nginx is used as a reverse proxy (there are some additional changes needed in Dockefile, githib actions, etc. to make it work in Docker while allowing us to have same deployment process with Github actions).
- small bug with migrations was discovered and is also fixed with this bug (`ADD COLUMN IF NOT EXISTS`)

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #1881, opened 2024-04-19
Fix failed rollup insertion on Host

badgersrus:
### Why this change is needed

https://github.com/ten-protocol/ten-internal/issues/3246

The connections to postgres keep running out and theres a lot of failed rollup insertions that result in idle connections.

```
| idle in transaction (aborted) | -8488598446982951149 | INSERT INTO rollup_host (hash, start_seq, end_seq, time_stamp, ext_rollup, compression_block) values ($1, $2, $3, $4, $5, $6) | client backend |
```

If this doesn't fix it we can try connection pooling via Azure but theres a cost associated. 

* Add rollback to storage interface

### PR checks pre-merging

### DIFF
--- a/go/host/storage/hostdb/hostdb.go
+++ b/go/host/storage/hostdb/hostdb.go
@@ -59,3 +59,10 @@ func (b *dbTransaction) Write() error {
 	}
 	return nil
 }
+
+func (b *dbTransaction) Rollback() error {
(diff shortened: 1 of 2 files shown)


badgersrus:
> lgtm
> 
> interesting. without the rollback, it doesn't return the connection to the pool.

I suspect this is postgres specific behaviour

PR #1884, opened 2024-04-26
Ziga/gateway verbose logging and bugfix

zkokelj:
### Why this change is needed

- Currently there is a bug in the Gateway causing following errors: 
```
2024/04/25 20:49:35 http: panic serving 127.0.0.1:59676: interface conversion: interface {} is map[string]interface {}, not string
```
- To improve further error checking we want to turn on verbose error checking for all deployed testnets

- additional type check
- verbose flag set to true in the deployment script

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #1904, opened 2024-05-09
Host: notify separately for new heads and new validated heads

BedrockSquirrel:
### Why this change is needed

We have bugs that are likely caused by the host newHeads subscriptions getting notified of new batches before the enclave has processed them.

Separate out the L2 repo's subscribers between those that want to know about a new head added to the DB and a new head batch that has been validated by that node's own enclave. Use the latter for the new heads RPC subscriptions.

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #1920, opened 2024-05-16
change /info/health/ from post to get

zkokelj:
### Why this change is needed

use GET request for `/info/health/` like with other endpoints 

### PR checks pre-merging

### DIFF
--- a/tools/tenscan/frontend/api/general.ts
+++ b/tools/tenscan/frontend/api/general.ts
@@ -9,6 +9,5 @@ export const fetchTestnetStatus = async (): Promise<
   return await httpRequest<ResponseDataInterface<boolean>>({
     method: "get",
     url: pathToUrl(apiRoutes.getHealthStatus),
-    data: { jsonrpc: "2.0", method: "obscuro_health", id: 1 },
   });
 };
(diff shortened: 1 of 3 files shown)


tools/tenscan/frontend/api/general.ts
```diff
@@ -7,8 +7,8 @@ export const fetchTestnetStatus = async (): Promise<
   ResponseDataInterface<boolean>
 > => {
   return await httpRequest<ResponseDataInterface<boolean>>({
-    method: "post",
+    method: "get",
     url: pathToUrl(apiRoutes.getHealthStatus),
-    data: { jsonrpc: "2.0", method: "obscuro_health", params: [], id: 1 },
```
badgersrus, line 12: I think we can get rid of this param from both functions since we're calling the HTTP endpoint at `/info/health` and not the RPC endpoint at `obscuro_health`

badgersrus (approved):
If we get rid of those two `data` params and tested those functions still work then this is good to go.

tools/tenscan/frontend/api/general.ts
zkokelj: removed 🗑️

PR #1941, opened 2024-05-31
Host: avoid spamming stuck L1 transactions

BedrockSquirrel:
### Why this change is needed

While retrying a stuck transaction we were incrementing the nonce. These pending transactions eventually all went through together costing the sequencer a fortune in unnecessary gas. Also, I think they were blocking each other which is why the stuck transactions carried on so long.

The nonce is acquired only once when the sending lock is acquired so price updates overwrite the transaction in the mempool.

### PR checks pre-merging

### DIFF
--- a/go/host/l1/publisher.go
+++ b/go/host/l1/publisher.go
@@ -391,12 +391,19 @@ func (p *Publisher) publishTransaction(tx types.TxData) error {
 
 	retries := -1
 
+	// we keep trying to send the transaction with this nonce until it is included in a block
+	// note: this is only safe because of the sendingLock guaranteeing only one transaction in-flight at a time
+	nonce, err := p.ethClient.EthClient().PendingNonceAt(p.sendingContext, p.hostWallet.Address())
+	if err != nil {
+		return fmt.Errorf("could not get nonce for L1 tx: %w", err)
+	}
+
 	// while the publisher service is still alive we keep trying to get the transaction into the L1
 	for !p.hostStopper.IsStopping() {
 		retries++ // count each attempt so we can increase gas price
 
 		// update the tx gas price before each attempt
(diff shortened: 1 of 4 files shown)


go/host/l1/publisher.go
```diff
@@ -391,12 +391,19 @@ func (p *Publisher) publishTransaction(tx types.TxData) error {
...
+		return fmt.Errorf("could not get nonce for L1 tx: %w", err)
+	}
+
 	// while the publisher service is still alive we keep trying to get the transaction into the L1
 	for !p.hostStopper.IsStopping() {
 		retries++ // count each attempt so we can increase gas price
 
 		// update the tx gas price before each attempt
-		tx, err := p.ethClient.PrepareTransactionToRetry(p.sendingContext, tx, p.hostWallet.Address(), retries)
+		tx, err := p.ethClient.PrepareTransactionToRetry(p.sendingContext, tx, p.hostWallet.Address(), nonce, retries)
```
badgersrus, line 406: Do we want to add some sort of timeout here? given the original impl choice was to avoid nonce reuse/ conflicts it looks like we might be putting that back in. Having a configurable timeout for how many times we want to try a nonce before we aquire a new one might be useful

BedrockSquirrel: We avoid the nonce-conflicts with the synchronization. We block on the sendingLock whenever a transaction is in-flight. So long as the nonce is fetched after locking I think that's safe.

I do think we might want to revisit this at some point to make it more intuitive but I think it's very safe. The unintuitive bit is that the calls to publish rollups etc. are blocking and so the rollup timer won't fire again until the rollup it was trying to publish goes through, it never gives up.

But this bug where I was fetching the pending nonce on each attempt was **causing** the backing up, the later txs with higher gas prices weren't going through because the first one was stuck.

tudor-malene (approved):
lgtm

PR #1942, opened 2024-06-03
Gateway rate limiting

zkokelj:
### Why this change is needed

Due to huge amount of calls from some IPs we want to guard some of our expensive endpoints with rate limiter

### PR checks pre-merging


tools/walletextension/ratelimiter/rate_limiter.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"sync"
+	"time"
+)
+
+type RateLimiter struct {
+	mu        sync.Mutex
+	users     map[string]time.Time
+	threshold time.Duration
```
tudor-malene, line 11: this should be a uint32

tudor-malene (commented):
I suggested a slight change in the approach, as it looks like we need to handle a few more cases.

Each request that has the `calculateRateLimitScore` lambda defined in the ExecConfig,, will append that value to the current score for the user.

We need to make the RateLimiter type a service with a Start and Stop method.
In the start method, there should be an infinite loop with a timer that fires every second (or something configurable), and reduces the score of all users by some amount.

This way, we have the requests adding as they make them, and in the background we reduce

tools/walletextension/rpcapi/utils.go
```diff
@@ -196,6 +196,14 @@ func extractUserID(ctx context.Context, _ *Services) ([]byte, error) {
 	return userID, nil
 }
 
+func extractUserIDHex(ctx context.Context, _ *Services) (string, error) {
```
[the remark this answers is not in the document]
zkokelj: removed this function

tools/walletextension/ratelimiter/rate_limiter.go
```diff
@@ -0,0 +1,33 @@
+package ratelimiter
+
+import (
+	"sync"
+	"time"
+)
+
+type RateLimiter struct {
+	mu        sync.Mutex
+	users     map[string]time.Time
```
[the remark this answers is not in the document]
zkokelj: []byte is not a comparable type in Go as far as I know and cannot be used as a key.
If there is a way around that I am happy to make the change.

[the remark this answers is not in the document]
tudor-malene: good point. What about using the geth Address type?

tudor-malene (commented):
lgtm
(after avoiding the conversion of the userid to string)

tools/walletextension/ratelimiter/rate_limiter.go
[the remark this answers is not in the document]
zkokelj: changed to Address type ✅

tools/walletextension/main/cli.go
```diff
@@ -59,6 +59,18 @@ const (
 	storeIncomingTxs        = "storeIncomingTxs"
 	storeIncomingTxsDefault = true
 	storeIncomingTxsUsage   = "Flag to enable storing incoming transactions in the database for debugging purposes. Default: true"
+
+	rateLimitThresholdName    = "rateLimitThreshold"
+	rateLimitThresholdDefault = 1000
+	rateLimitThresholdUsage   = "Rate limit threshold per user. Default: 1000."
+
+	rateLimitDecayName    = "rateLimitDecay"
+	rateLimitDecayDefault = 0.2
```
tudor-malene, line 68: If we make this "1", then we keep parity with the milliseconds, and we can reason in units of time

tools/walletextension/main/cli.go
```diff
@@ -59,6 +59,18 @@ const (
...
+	rateLimitDecayUsage   = "Rate limit decay per user. Default: 0.2"
+
+	rateLimitMaxScoreName    = "rateLimitMaxScore"
```
tudor-malene, line 71: this parameter is confusing.

tools/walletextension/main/cli.go
```diff
@@ -59,6 +59,18 @@ const (
 	storeIncomingTxs        = "storeIncomingTxs"
 	storeIncomingTxsDefault = true
 	storeIncomingTxsUsage   = "Flag to enable storing incoming transactions in the database for debugging purposes. Default: true"
+
+	rateLimitThresholdName    = "rateLimitThreshold"
+	rateLimitThresholdDefault = 1000
```
tudor-malene, line 64: If this is 200 and the decay is 1, we allocate 200ms/per second to each user.

tools/walletextension/ratelimiter/rate_limiter.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+// Package ratelimiter implements a simple rate limiting mechanism
+// using a score-based approach. Each user has a score that decays
+// over time. Requests are allowed if the user's score is below a
+// specified threshold. The score increases with each request and
+// decays based on the time since the last request.
+
+type Score struct {
+	lastRequest time.Time
```
tudor-malene, line 19: let's add a "concurrentRequests"  value here as well, which we increment during `Allow` and decrement during "Update".

In "Allow" we should also check that the concurrentRequests needs to be below a configured threshold (A small value like 3).

With this mechanism we protect someone from Dos-ing the gateway by triggering a very large number of expensive requests

tudor-malene (changes requested):
The logic is clean now. Needs an additional mechanism

tools/walletextension/rpcapi/utils.go
```diff
@@ -149,7 +153,13 @@ func ExecAuthRPC[R any](ctx context.Context, w *Services, cfg *ExecCfg, method s
 		return nil, rpcErr
 	})
 
-	audit(w, "RPC call. uid=%s, method=%s args=%v result=%s error=%s time=%d", hexutils.BytesToHex(userID), method, args, res, err, time.Since(requestStartTime).Milliseconds())
+	executionDuration := time.Since(requestStartTime).Milliseconds()
+	// The Execution cost is 100 times the execution duration in milliseconds;
+	// we can change how much user can run by changing decay rate
+	executionCost := uint32(executionDuration) * 100
+	w.RateLimiter.UpdateScore(gethcommon.Address(userID), executionCost)
```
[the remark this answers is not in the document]
zkokelj: fixed

tools/walletextension/ratelimiter/rate_limiter.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// Allow checks if the user is allowed to make a request based on the rate limit threshold
+// before comparing to the threshold also decays the score of the user based on the decay rate
+func (rl *RateLimiter) Allow(userID common.Address) bool {
+	rl.mu.Lock()
+	defer rl.mu.Unlock()
+
+	// If the threshold is 0, allow all requests (rate limiting is disabled)
+	if rl.threshold == 0 {
```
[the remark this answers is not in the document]
zkokelj: fixed

tools/walletextension/ratelimiter/rate_limiter.go
```diff
@@ -0,0 +1,73 @@
...
+		rl.users[userID] = Score{lastRequest: now, score: uint32(newScore)}
+	}
+
+	// Check if the user's score is below the threshold
+	return rl.users[userID].score <= rl.threshold
+}
+
+// UpdateScore updates the score of the user based on the additional score (time taken to process the request)
+func (rl *RateLimiter) UpdateScore(userID common.Address, additionalScore uint32) {
+	rl.mu.Lock()
```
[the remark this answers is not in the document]
zkokelj: added

tools/walletextension/ratelimiter/rate_limiter.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// If the threshold is 0, allow all requests (rate limiting is disabled)
+	if rl.threshold == 0 {
+		return true
+	}
+
+	now := time.Now()
+	userScore, exists := rl.users[userID]
+
+	if !exists {
```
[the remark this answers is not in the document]
zkokelj: fixed

tools/walletextension/main/cli.go
zkokelj: fixed

tools/walletextension/main/cli.go
zkokelj: removed this parameter

tools/walletextension/ratelimiter/rate_limiter.go
zkokelj: added

tools/walletextension/main/cli.go
zkokelj: If the decay is 1 then every ms one ms is subtracted from the score.
Setting `decay` to `0.2` means that every second it decays for 200ms. 
Setting `threshold` to 1000 means that users can use 1000ms in a second, but it will decay with the decay rate above. So we are not rate limiting the user if he created a request that takes `250ms` and wants to call another request immediately after that. But if he/she keeps doing it rate limiting will kick in.

tools/walletextension/ratelimiter/rate_limiter.go
```diff
@@ -52,56 +53,62 @@ func NewRateLimiter(threshold uint32, decay float64, maxScore uint32) *RateLimit
...
 	defer rl.mu.Unlock()
-	newScore := rl.users[userID].score + additionalScore
-	rl.users[userID] = Score{lastRequest: time.Now(), score: newScore}
+
+	// Decay the score of the user based on the time since the last request
+	now := time.Now()
+	userScore, exists := rl.users[userID]
+	scoreDecay := uint32(0)
+	// if user exists decay the score based on the time since the last request
+	if exists {
```
tudor-malene, line 101: if there is no score for that user, it's a programming error. This method should always be called after the previous one

tools/walletextension/ratelimiter/rate_limiter.go
```diff
@@ -52,56 +53,62 @@ func NewRateLimiter(threshold uint32, decay float64, maxScore uint32) *RateLimit
...
+		// Decay the score based on the time since the last request and the decay rate
+		timeSinceLastRequest := float64(now.Sub(userScore.lastRequest).Milliseconds())
```
tudor-malene, line 103: I don't quite understand why we need to decay the score here. It's hard to reason about what's going on, as that responsibility is now in 2 places.

tools/walletextension/ratelimiter/rate_limiter.go
```diff
@@ -52,56 +53,62 @@ func NewRateLimiter(threshold uint32, decay float64, maxScore uint32) *RateLimit
...
+		// limit the decay to the user's current score
+		scoreDecay = min(uint32(timeSinceLastRequest*rl.decay), userScore.score)
+	}
+
+	// Increase the score of the user based on the execution duration of the request
+	// and update the last request time
+	newScore := rl.users[userID].score + executionDuration - scoreDecay
+	rl.users[userID] = Score{lastRequest: now, score: newScore, concurrentRequests: userScore.concurrentRequests - 1}
```
tudor-malene, line 111: let's not change the lastRequest here.

tudor-malene (changes requested):
I think we can simplify a bit by keeping the responsibilities more compact.

Also, the rate limit: allow/update needs to be added to the: `filter_api.GetLogs` method, which doesn't use `ExecAuthRPC`

tools/walletextension/ratelimiter/rate_limiter.go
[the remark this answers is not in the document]
zkokelj: should never happen, but added additional check for it

tools/walletextension/ratelimiter/rate_limiter.go
zkokelj: logging that error now

tools/walletextension/ratelimiter/rate_limiter.go
zkokelj: We decay it only here in `UpdateScore`. 
But we need to calculate it also in `Allow` function to compare it to the threshold is this now what you meant in this comment: https://github.com/ten-protocol/go-ten/pull/1942#discussion_r1632877645 ?

tools/walletextension/ratelimiter/rate_limiter.go
zkokelj: fixed

tools/walletextension/main/cli.go
zkokelj: yeah, but it means that we allow by default for every user to use 1000ms in 1000ms on average (100%). Rate limits will be hit very rarely and only with concurrent requests.

tools/walletextension/ratelimiter/rate_limiter.go
```diff
@@ -0,0 +1,215 @@
...
+	return false, zeroUUID
+}
+
+// PruneRequests deletes all requests that have ended before the rate limiter's window.
+func (rl *RateLimiter) PruneRequests() {
+	rl.mu.Lock()
+	defer rl.mu.Unlock()
+
+	// delete all the requests that have
+	cutoff := time.Now().Add(-time.Duration(rl.window) * time.Millisecond)
```
BedrockSquirrel, line 178: It might be nice if rl.window was already a Duration (like I mentioned in the config stuff) for lines like this.

tools/walletextension/ratelimiter/rate_limiter.go
```diff
@@ -0,0 +1,215 @@
...
+	if userComputeTimeForUser <= rl.userComputeTime {
+		requestUUID := rl.AddRequest(userID, RequestInterval{Start: time.Now()})
+		return true, requestUUID
+	}
+	rl.IncrementRateLimitedRequests()
+	return false, zeroUUID
+}
+
+// PruneRequests deletes all requests that have ended before the rate limiter's window.
+func (rl *RateLimiter) PruneRequests() {
```
BedrockSquirrel, line 173: Be nice to know if this starts being slow, maybe a log line at the end if it's over some amount like:

```
timeTaken := time.Since(startTime)
if timeTaken > 1 * time.Second {
   log.Printf("PruneRequests completed in %s", timeTaken)
}
```

Bit worried the contention on the mutex lock could become troublesome. If it does I guess we can add a per-user rw-lock in the User object so pruning doesn't stop new requests on existing users being allowed.

BedrockSquirrel (commented):
LGTM to what Tudor requested. My comments are minor things, please ignore any you disagree with!

.github/workflows/manual-deploy-obscuro-gateway.yml
```diff
@@ -137,7 +137,8 @@ jobs:
...
-              --log-opt max-file=3 --log-opt max-size=10m \
-              ${{ vars.DOCKER_BUILD_TAG_GATEWAY }} \
-              -host=0.0.0.0 -port=8080 -portWS=81 -nodeHost=${{ vars.L2_RPC_URL_VALIDATOR }} -verbose=true \
-              -logPath=sys_out -dbType=mariaDB -dbConnectionURL="obscurouser:${{ secrets.OBSCURO_GATEWAY_MARIADB_USER_PWD }}@tcp(obscurogateway-mariadb-${{  github.event.inputs.testnet_type }}.uksouth.cloudapp.azure.com:3306)/ogdb"'
+               --log-opt max-file=3 --log-opt max-size=10m \
+               ${{ vars.DOCKER_BUILD_TAG_GATEWAY }} \
+               -host=0.0.0.0 -port=8080 -portWS=81 -nodeHost=${{ vars.L2_RPC_URL_VALIDATOR }} -verbose=true \
+               -logPath=sys_out -dbType=mariaDB -dbConnectionURL="obscurouser:${{ secrets.OBSCURO_GATEWAY_MARIADB_USER_PWD }}@tcp(obscurogateway-mariadb-${{  github.event.inputs.testnet_type }}.uksouth.cloudapp.azure.com:3306)/ogdb" \
+               -rateLimitThreshold=${{ vars.GATEWAY_RATE_LIMIT_THRESHOLD }} -rateLimitDecay=${{ vars.GATEWAY_RATE_LIMIT_DECAY }} -maxConcurrentRequestsPerUser=${{ vars.GATEWAY_MAX_CONCURRENT_REQUESTS_PER_USER }} '
```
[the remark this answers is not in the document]
zkokelj: Fixed, thanks 👍

tools/walletextension/main/cli.go
```diff
@@ -59,6 +59,18 @@ const (
 	storeIncomingTxs        = "storeIncomingTxs"
 	storeIncomingTxsDefault = true
 	storeIncomingTxsUsage   = "Flag to enable storing incoming transactions in the database for debugging purposes. Default: true"
+
+	rateLimitUserComputeTimeName    = "rateLimitUserComputeTime"
```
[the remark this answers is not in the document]
zkokelj: I made it more clear that it is in ms. If you think it is better to use `time.ParseDuration` I can easily change it to that too.

tools/walletextension/ratelimiter/rate_limiter.go
```diff
@@ -0,0 +1,215 @@
...
+		}
+		rl.users[userID] = user
+	}
+	id := uuid.New()
+	user.CurrentRequests[id] = interval
+	return id
+}
+
+// UpdateRequest updates the end time of a request interval given its UUID.
+func (rl *RateLimiter) UpdateRequest(userID common.Address, id uuid.UUID) {
```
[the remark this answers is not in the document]
zkokelj: Renamed 👍

tools/walletextension/ratelimiter/rate_limiter.go
```diff
@@ -0,0 +1,215 @@
...
+	if user, userExists := rl.users[userID]; userExists {
+		if request, requestExists := user.CurrentRequests[id]; requestExists {
+			rl.mu.Lock()
+			defer rl.mu.Unlock()
+			now := time.Now()
+			request.End = &now
+			user.CurrentRequests[id] = request
+		} else {
+			log.Printf("Request with ID %s not found for user %s.", id, userID.Hex())
```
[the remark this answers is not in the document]
zkokelj: Logger used

tools/walletextension/ratelimiter/rate_limiter.go
```diff
@@ -0,0 +1,215 @@
...
+// and returns it as uint32 milliseconds.
+func (rl *RateLimiter) SumComputeTime(userID common.Address) uint32 {
+	rl.mu.Lock()
+	defer rl.mu.Unlock()
+
+	var totalComputeTime time.Duration
+	if user, exists := rl.users[userID]; exists {
+		cutoff := time.Now().Add(-time.Duration(rl.window) * time.Millisecond)
+		for _, interval := range user.CurrentRequests {
+			if interval.End != nil && interval.End.After(cutoff) {
```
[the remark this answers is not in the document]
zkokelj: fixed

tools/walletextension/ratelimiter/rate_limiter.go
```diff
@@ -0,0 +1,215 @@
...
+
+	// Check if the user has reached the maximum number of concurrent requests
+	if uint32(rl.CountOpenRequests(userID)) >= rl.GetMaxConcurrentRequest() {
+		rl.IncrementRateLimitedRequests()
+		return false, zeroUUID
+	}
+
+	// Check if user is in limits of rate limiting
+	userComputeTimeForUser := rl.SumComputeTime(userID)
+	if userComputeTimeForUser <= rl.userComputeTime {
```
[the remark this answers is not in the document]
zkokelj: fixed & noted.. will stick with happy path at the end of methods when possible.

tools/walletextension/ratelimiter/rate_limiter.go
zkokelj: added logs if time taken > 1s. 

I am aware of mutex lock problems (created getters, increments and not locking it for too long).
Do you think we should try with this and later (if it becomes a problem) switch to per-user locking?

tools/walletextension/main/cli.go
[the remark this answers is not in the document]
zkokelj: Using duration now, thanks.

tools/walletextension/ratelimiter/rate_limiter.go
BedrockSquirrel: Yeah I think it's fine for now all your locks looked good to me and might not be an issue at all. But something to consider if we start seeing unexplained slowness at some point.

tools/walletextension/ratelimiter/rate_limiter.go
zkokelj: removed not needed conversions to Duration

BedrockSquirrel (approved):
lgtm

tools/walletextension/rpcapi/utils.go
```diff
@@ -149,6 +154,8 @@ func ExecAuthRPC[R any](ctx context.Context, w *Services, cfg *ExecCfg, method s
 		return nil, rpcErr
 	})
 
+	w.RateLimiter.SetRequestEnd(gethcommon.Address(userID), requestUUID)
```
tudor-malene, line 157: one comment. I think this should better be in a defer 
(Same for filter_api)

zkokelj: fixed.

tudor-malene (approved):
lgtm

PR #1947, opened 2024-06-03
Network tests: start gateway synchronously to fix race

BedrockSquirrel:
### Why this change is needed

Running test against testnets but with local gateway was failing because it started async and wasn't up in time for the test interactions.

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #1959, opened 2024-06-11
[gateway] assets fix and clear console warnings

Jennievon:
### Why this change is needed

https://github.com/ten-protocol/go-ten/issues/1958

- update logo relative path for dark mode theme
- change `http-equiv` to `httpEquiv` as React recognises the former as an `Invalid DOM property`
- fix the fetch priority error: `Warning: React does not recognize the `fetchPriority` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `fetchpriority` instead.`
this occured because there was a bug in the next package version and the bug has been fixed but we needed to update the package version too

### PR checks pre-merging


tools/walletextension/frontend/src/components/head-seo.tsx
```diff
@@ -25,9 +25,9 @@ const HeadSeo = ({
         signature="_vd3udx2g2hfn9zclob5cat43b94q7fyk"
       ></meta>
       {/* SECURITY: to prevent the page from being loaded in an iFrame */}
-      <meta http-equiv="X-Frame-Options" content="deny"></meta>
+      <meta httpEquiv="X-Frame-Options" content="deny"></meta>
```
zkokelj, line 28: image

I stil get those errors in console.

tools/walletextension/frontend/src/components/layouts/header.tsx
```diff
@@ -14,7 +14,7 @@ export default function Header() {
       <div className="flex h-16 justify-between items-center px-4">
         <Link href="/">
           <Image
-            src="/static/assets/images/black_logotype.png"
+            src="/assets/images/black_logotype.png"
```
zkokelj, line 17: logos are now broken in both white and black version.
I run: npm run build and then serve static files with golang (same way as then deployed to Azure).

zkokelj (changes requested):
It doesn't fix things for me. If you want we can check that together.

zkokelj (approved):
LGTM.

I would also remove:

<meta http-equiv="X-Frame-Options" content="deny"></meta>

since it is reporting an error in console and since it is incorrect way to do this it is not protecting us at all.

PR #1965, opened 2024-06-20
Initiate bridge deployment

Jennievon:
### Why this change is needed

update the Dockerfile paths for bridge deployment

### PR checks pre-merging

### DIFF
--- a/contracts/src/bridge/frontend/Dockerfile
+++ b/contracts/src/bridge/frontend/Dockerfile
@@ -1,27 +1,9 @@
-# Use an official Node.js 22 as a parent image
-FROM node:22-alpine
+FROM node:18-buster as runner
+# setup container data structure
+RUN mkdir -p /home/obscuro/go-ten/tools/contracts/src/bridge
+COPY ./contracts/src/bridge/frontend /home/obscuro/go-ten/contracts/src/bridge/frontend
 
-WORKDIR /usr/src/app
-
-# ARG for build-time variable
-ARG API_HOST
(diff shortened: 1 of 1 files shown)


contracts/src/bridge/frontend/Dockerfile
```diff
@@ -1,7 +1,9 @@
-# Use an official Node.js 22 as a parent image
-FROM node:22-alpine
+FROM node:18-buster as runner
```
BedrockSquirrel, line 1: Probably better to use newest version if it works (like changing `node:18-buster` -> `node:22-alpine`). Shouldn't make a difference but might avoid some issues.

contracts/src/bridge/frontend/Dockerfile
```diff
@@ -17,11 +19,6 @@ ENV NEXT_PUBLIC_BRIDGE_L2=${L2_BRIDGE}
...
-COPY package*.json ./
-
 RUN npm install
-COPY . .
-RUN npm run build
-EXPOSE 80
-CMD ["npm", "start"]
\ No newline at end of file
+
+EXPOSE 80
```
BedrockSquirrel, line 24: Not sure if you still need `CMD ["npm", "start"]` here? Depends if that's getting called by whatever kicks off the server I guess, maybe you don't need it if you're copying how it works for some other app

BedrockSquirrel (approved):
LGTM - couple minor comments but no blockers if this works

PR #2001, opened 2024-07-31
Deployment scripts rework to work locally.

StefanIliev545:
### Why this change is needed

Deployment scripts are setup in a way that requires them to run in a docker container, launched from the hardhat launcher  go file and wired with some specific env variables. This PR will change this.

Instead of only using the ENV variables, whenever they are NOT declared, the deployment scripts will poll the provider for the `net_config` endpoint, which in turn goes to `obscuro_config` under the gateway and returns the contract addresses that matter - Management, L1 Bus, L2 Bus. This way the deployment can figure out its dependencies from the network and continue forth. This allows the local deployments to work. 

### PR checks pre-merging


tudor-malene (commented):
lgtm ( if it works)

tudor-malene (approved):
lgtm

PR #2002, opened 2024-07-31
Revert message address indexing.

StefanIliev545:
### Why this change is needed

Adding indexing was done for the front end for both value transfer events and message published events. Unfortunately, this broke the event filtering. This PR fixes this. 

Reverted the indexed keyword from the message published event along with all the extra stuff surrounding it.

### PR checks pre-merging


tudor-malene (approved):
lgtm ( after adding a short description)

PR #2019, opened 2024-08-13
Correct print output

moraygrieve:


tudor-malene (approved):
lgtm. ( after formatting)

PR #2055, opened 2024-09-13
fixed host name for promtail config

pkrishnath:
### Why this change is needed

Fixed hostname env var for promtail config

### DIFF
--- a/.github/workflows/manual-deploy-testnet-l1.yml
+++ b/.github/workflows/manual-deploy-testnet-l1.yml
@@ -113,20 +113,50 @@ jobs:
             && sudo snap refresh \
             && curl -fsSL https://get.docker.com -o get-docker.sh && sh ./get-docker.sh \
             && docker network create --driver bridge l1_network || true \
-            && docker run -d --name datadog-agent \
-               --network l1_network \
-               -e DD_API_KEY=${{ secrets.DD_API_KEY }} \
-               -e DD_LOGS_ENABLED=true \
-               -e DD_LOGS_CONFIG_CONTAINER_COLLECT_ALL=true \
-               -e DD_LOGS_CONFIG_AUTO_MULTI_LINE_DETECTION=true \
-               -e DD_CONTAINER_EXCLUDE_LOGS="name:datadog-agent" \
-               -e DD_SITE="datadoghq.eu"  \
-               -v /var/run/docker.sock:/var/run/docker.sock:ro \
-               -v /proc/:/host/proc/:ro \
-               -v /opt/datadog-agent/run:/opt/datadog-agent/run:rw \
-               -v /sys/fs/cgroup/:/host/sys/fs/cgroup:ro \
-               --log-opt max-file=3 --log-opt max-size=10m \
-               datadog/agent:latest \
+            && mkdir -p /home/obscuro/promtail \
+            && echo "
+            server:
+              http_listen_port: 9080
+              grpc_listen_port: 0
+
+            positions:
+              filename: /tmp/positions.yaml
+
+            clients:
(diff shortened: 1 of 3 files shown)


.github/workflows/manual-deploy-testnet-l1.yml
```diff
@@ -113,20 +113,50 @@ jobs:
...
+                  refresh_interval: 5s
+              relabel_configs:
+                - source_labels: [\"__meta_docker_container_name\"]
+                  regex: \"/(.*)\"
+                  target_label: \"container\"
+                - source_labels: [\"__meta_docker_container_log_stream\"]
+                  target_label: \"logstream\"
+                - source_labels: [\"__meta_docker_container_label_logging_jobname\"]
+                  target_label: \"job\"
+                - replacement: ${{ matrix.host_id }}-${{ github.event.inputs.testnet_type }}-${{ GITHUB.RUN_NUMBER }}
```
BedrockSquirrel, line 148: There is no 'matrix' in this script like there is in the L2 deploy script so maybe this will behave unpredictably. If it's just for keys to show how it will be rendered in Loki then maybe can be hardcoded to like `L1-${{ github.event.inputs.testnet_type }}` or something?

Not a blocker though, very rare we deploy the L1 for testnets and looks like someone already put in a matrix usage and it doesn't seem to be breaking so maybe it'll just do an empty string or something.

.github/workflows/manual-deploy-testnet-validator.yml
```diff
@@ -148,20 +148,50 @@ jobs:
...
+                  refresh_interval: 5s
+              relabel_configs:
+                - source_labels: [\"__meta_docker_container_name\"]
+                  regex: \"/(.*)\"
+                  target_label: \"container\"
+                - source_labels: [\"__meta_docker_container_log_stream\"]
+                  target_label: \"logstream\"
+                - source_labels: [\"__meta_docker_container_label_logging_jobname\"]
+                  target_label: \"job\"
+                - replacement: ${{ matrix.host_id }}-${{ github.event.inputs.testnet_type }}-${{ GITHUB.RUN_NUMBER }}
```
BedrockSquirrel, line 183: Similarly because this validator deploys just a single node rather than the 'matrix' of 3 nodes deployed by the main deploy script so not sure what will happen if we try to run this.

But again not a blocker, rarely used.

BedrockSquirrel (approved):
LGTM. I think the L1 and validator deploy scripts might not work properly with those matrix references though so can either fix that now or get it fixed replaced/later I guess (they're rarely used).

BedrockSquirrel:
Worth mentioning as well since we discussed separately, this is a stop-gap solution and all this stuff goes away in the new world with the CD tool doesn't it. But looks like this gets the automation of the agents working with our existing stuff nicely for now.

.github/workflows/manual-deploy-testnet-validator.yml
pkrishnath: fixed: ${{ vars.AZURE_RESOURCE_PREFIX }}-${{ github.event.inputs.node_id }}-${{ GITHUB.RUN_NUMBER }}

.github/workflows/manual-deploy-testnet-l1.yml
pkrishnath: fixed : ${{ github.event.inputs.testnet_type }}-eth2network-${{ GITHUB.RUN_NUMBER }} as host ID

BedrockSquirrel (approved):
Nice, LGTM

pkrishnath:
Replaced datadog agent with promtail in l1,l2 and validator dev-testnet workflows

PR #2067, opened 2024-09-30
Cadvisor and prometheus

pkrishnath:
### Why this change is needed

This is the ticket to add cadvisor  and prometheus to deploy in ten nodes for enhanced metrics collection 


BedrockSquirrel (approved):
LGTM - haven't checked the config/ports in detail but suggest we spin up a dev-testnet with it before merging to check it's all working (if you haven't already)

PR #2074, opened 2024-10-02
fix: added promtail for gateway and gateway DB

pkrishnath:
### Why this change is needed

Added promtail agents for gateway and gateway DB


zkokelj (approved):
LGTM

PR #2083, opened 2024-10-07
remove github run number from validator dns name

zkokelj:
### Why this change is needed

At the moment we have github run number in DNS name for our Ten nodes. This makes DNS names different after every redeploy. 
Having constant DNS names for our validators is necessary for load balancing with Azure gateway (if we want to remove manual work after every redeploy) and it also allows us to get rid of load balancer that points to the second validator for dexynth node. 

Removed github run number from dns name.

### PR checks pre-merging

### DIFF
--- a/go/enclave/rpc/EstimateGas.go
+++ b/go/enclave/rpc/EstimateGas.go
@@ -198,6 +198,15 @@ func (rpc *EncryptionManager) doEstimateGas(ctx context.Context, args *gethapi.T
 		hi = gasCap
 	}
 	cap = hi //nolint: revive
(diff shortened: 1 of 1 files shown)


BedrockSquirrel (approved):
LGTM

BedrockSquirrel (changes requested):
Actually do we need to update some other usages, like this one:
```
-sequencer_addr=obscuronode-0-${{ github.event.inputs.testnet_type }}-${{ GITHUB.RUN_NUMBER }}.uksouth.cloudapp.azure.com:10000 \
```

There's probably a few like that (unless I'm misunderstanding what this impacts).

PR #2091, opened 2024-10-16
WIP gateway encrypt and decrypt stored data

zkokelj:
### Why this change is needed

We want to store encrypted data in the database with the key generated inside the gateway (which runs inside an enclave).
By encrypting data ourselves we can use "normal" database and avoid EdgelessDB.

- Random key generator (this key is then used for symmetric encryption)
- Encryptor which handles all the encryption/decryption (It uses AES-GCM encryption with key generated above and random nonce)
- EncryptedStorage type
- Usage of EncryptedStorage to store data in the database
- We store `userID` field not only encrypted but also as a `hash`, because we need to query it in the database and querying encrypted data is impossible (different cypher text for same input text, due to nonces). 

At the moment I only implemented changes to sqlite database and after the confirmation that this is the approach we want to go with I can adapt it to MariaDB/CosmosDB. 

### PR checks pre-merging

### DIFF
--- a/tools/walletextension/common/common.go
+++ b/tools/walletextension/common/common.go
@@ -1,6 +1,7 @@
 package common
 
 import (
+	"crypto/rand"
 	"encoding/json"
 	"fmt"
 
@@ -75,4 +76,13 @@ func (r *RPCRequest) Clone() *RPCRequest {
 		Method: r.Method,
 		Params: r.Params,
 	}
+}
+
+func GenerateRandomKey(keySize int) ([]byte, error) {
+	key := make([]byte, keySize)
+	_, err := rand.Read(key)
+	if err != nil {
+		return nil, err
+	}
+	return key, nil
 }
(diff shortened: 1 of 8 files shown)


tudor-malene (commented):
The best approach for encryption/kv implementation.

Instead of the current fine-grained `Storage` we should always work on the entire `User` object. 
The main reason is that we need to minimize the information we're sending to I/O. 

1. Create this `EncryptedUserStorage` interface and implementation, and wire it where we currently use `getUser` and the Storage methods.

```golang
// high level, to be used from the current business logic 
type EncryptedUserStorage interface {
        SaveUser(GWUser) error
        LoadUser(userID []byte) (GWUser, error)
}
```

2. There should also be a class that knows how to persist stuff - the CosmosDb implementation)

```golang
// low level - just stores and reads bytes
type RawStorage interface {
        Read (key []byte )        ([]byte, error)
        Write (key []byte, value []byte)        error
}
```
-------
The implementation for `EncryptedUserStorage` should do the following:

#### LoadUser:

1. check the Cache whether there is a GWUser (we have this logic in `getUser` now) -  If there is then return it.
2. Hash the userId and call Read on the `RawStorage`
3. Decrypt the return value, unmarshall using Json into a GWUser object, and return


#### SaveUser:

1. Marshall to json byte array
2. Encrypt the result
3. Hash the userid
4. Call RawStorage.Write(hashedUserId, encryptedGWUserJson)
5. Update the cache with the new value

PR #2118, opened 2024-10-30
Fix local testnet gateway logs

zkokelj:
### Why this change is needed

We want to store logs of gateway container to file to be able to inspect them later.

- Use new Docker image that is used also in deployed version of the gateway.
- Use log file instead of stdout for storing logs

### PR checks pre-merging

Tested on local network:

image


tudor-malene (approved):
lgtm - if it works

PR #2122, opened 2024-11-02
Fix Gateway frontend dockerfile 

zkokelj:
### Why this change is needed

Currently docker container doesn't build. It needs fixing to be able to be deployed from main.

- Changes in Dockerfile to allow it to build and start a contaner.
- minor fix in deployment script

### PR checks pre-merging

### DIFF
--- a/tools/walletextension/frontend/Dockerfile
+++ b/tools/walletextension/frontend/Dockerfile
@@ -1,8 +1,8 @@
 # Use an official Node.js LTS version as a base image
 FROM node:20-alpine AS base
+
 # Install pnpm
 RUN npm install -g pnpm
-
 
 ARG NEXT_PUBLIC_NETWORK_NAME
 ARG NEXT_PUBLIC_TENSCAN_URL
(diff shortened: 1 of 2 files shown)


tools/walletextension/frontend/Dockerfile
```diff
@@ -21,25 +21,28 @@ WORKDIR /usr/src/app
...
+RUN pnpm install --frozen-lockfile
 
 # Build the Next.js app
 RUN pnpm run build
 
 # Reduce the size of the final image by using a lighter base image
 FROM node:20-alpine AS runner
 
+# Install pnpm
+RUN npm install -g pnpm
```
badgersrus, line 33: I think this one is redundant since its installed at the start? Not sure if that prod flag works the same with pnpm but maybe @pbienek might know more

badgersrus (approved):
one very minor comment, lgtm

tools/walletextension/frontend/Dockerfile
zkokelj: It is a different image (`base` and `runner` image)  and it needs `pnpm` installed. 
As far as I know `--prod` flag doesn't install dev dependencies which makes runner image smaller.

PR #2129, opened 2024-11-07
Log Gateway logs to a file

zkokelj:
### Why this change is needed

Use the same Dockerfile as in deployed version (but in simulation mode) for local testnet and log to file and not to `std_out`

At the moment we are sticking to non SGX version of the gateway for local testnets, due to a problem with creating files inside SGX mode in simulation mode. Priority is to implement other tasks more crucial for mainnet release.

### PR checks pre-merging


tudor-malene (approved):
lgtm ( if it works)

PR #2136, opened 2024-11-11
use ETag with CosmosDB

zkokelj:
### Why this change is needed

We need to implement optimistic locking for CosmosDB.

Two threads could read an user from the database at the same time and then both can modify and store the user.
With current implementation both writes to database are successful which can cause us to loose some data without seeing an error anywhere. (Example: in both threads we add an account to user and the second thread overwrites the account being added by the first one).

Using ETag from CosmosDB to check if the version that we are overwriting is the same as the one that we read from.

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #2149, opened 2024-11-19
Expose EnclaveID to RPC 

badgersrus:
### Why this change is needed

https://github.com/ten-protocol/ten-internal/issues/4250

* Expose `EnclaveID` on the `Status` RPC call
* New `GetEnclaveClients` host API to return all enclaves as the existing function is hard coded to the zero index  
* Manually tested using `sequencer_multi_enclave_test.go`

### PR checks pre-merging


go/common/enclave.go
```diff
@@ -26,6 +26,7 @@ type Status struct {
 	StatusCode StatusCode
 	L1Head     gethcommon.Hash
 	L2Head     *big.Int
+	EnclaveID
```
BedrockSquirrel, line 29: I think this field should be included as a named field like `EnclaveID EnclaveID` - embedding it like this might be confusing or have unexpected side-effects

BedrockSquirrel (commented):
LGTM, 1 quick comment about having it as a named field

PR #2156, opened 2024-11-22
Gateway - handle Ten Gateway key exchange

zkokelj:
### Why this change is needed

We use encrypted cosmosDB database with encryption key generated inside the enclave. In order to be able to share the key with another instance of the gateway (either new version of the gateway or dedicated gateway that needs to share the same database) we need to create a secure way of sharing the key between two enclaves.

- Sealing and unsealing of encryption keys
- Key exchange between two enclaves

### PR checks pre-merging


tools/walletextension/keymanager/keymanager.go
```diff
@@ -0,0 +1,287 @@
...
+			logger.Crit("unable to decode and decrypt the encrypted key", "error", err)
+			return nil, err
+		}
+
+		// Now you have the encryption key
+		logger.Info("Successfully retrieved and decrypted the encryption key")
+
+		fmt.Println("decrypted key: ", decryptedKey)
+
+		return decryptedKey, nil
```
BedrockSquirrel, line 143: shouldn't we seal/persisting key in this case too before we return?

tools/walletextension/httpapi/routes.go
```diff
@@ -517,3 +523,66 @@ func getMessageRequestHandler(walletExt *services.Services, conn UserConn) {
...
+
+	// Verify the attestation report
+	verifiedData, err := keymanager.VerifyReport(&attestationReport)
+	if err != nil {
+		handleError(conn, walletExt.Logger(), fmt.Errorf("error verifying attestation report: %w", err))
+		return
+	}
+
+	// TODO: How do we need to handle the verified data? What should we check to be sure it is correct?
+	// Is this the hash of the public key?
```
BedrockSquirrel, line 550: you put the public key directly in the report I think? So you don't need to hash, just check the `verifiedData == report.PubKey`.

In the node enclave we have a json obj (pubkey, host address and stuff) and we put the hash of that into the report. So on the verify side we have to reproduce that and check if `hash(jsonData) == verifiedData`.

BedrockSquirrel (commented):
Couple things to consider but approach looks good to me high-level.

Probably better to just do the log.crit in one place (higher level) and return errors in the method so that easier to change how we handle it if needed. But guess you'll probably tidy up all that stuff when you're cleaning up logging/todos etc.

tools/walletextension/keymanager/keymanager.go
zkokelj: fixed

tools/walletextension/httpapi/routes.go
zkokelj: I think we need hash, otherwise we get the following error:
`REQ: Failed to get attestation report    component=wallet_extension error="reportData too large"`

tools/walletextension/keymanager/keymanager.go
```diff
@@ -0,0 +1,327 @@
...
+	if found {
+		logger.Info("found sealed encryption key")
+		return encryptionKey, nil
+	}
+
+	// 3.) We have to exchange the key with another enclave if we have a key exchange url set
+	if config.KeyExchangeURL != "" {
+		encryptionKey, err = HandleKeyExchange(config, logger)
+		if err != nil {
+			logger.Warn("unable to exchange key", log.ErrKey, err)
```
BedrockSquirrel, line 72: I think we have to return the error here. If Gateway has been configured with a key exchange url and something's gone wrong then I don't think it's safe to assume we can create a new key - does that make sense or am I confused?

BedrockSquirrel (commented):
LGTM except there's one spot I think we really need to error so I'll hold off on approval for that.

Also minor thing but I found the "OG" name quite confusing when it popped up, I guess it means the original gateway? Like the first/genesis/initial gateway that is going to generate the secret? But maybe could put that explicitly somewhere in a comment or something.

tools/walletextension/httpapi/routes.go
```diff
@@ -613,3 +621,84 @@ func boolToByte(res bool) byte {
...
+
+	// Only compare the first 32 bytes since verifiedData is padded to 64 bytes
+	verifiedDataTruncated := verifiedData[:32]
+	if bytes.Equal(verifiedDataTruncated, pubKeyHash[:]) {
+		walletExt.Logger().Info("OG: Public keys match")
+	} else {
+		walletExt.Logger().Error("OG: Public keys do not match")
+	}
+
+	// Step 5Encrypt the public key using the received public key
```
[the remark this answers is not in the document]
zkokelj: fixed

tools/walletextension/keymanager/keymanager.go
zkokelj: Agreed, changed to `logger.Crit`

zkokelj: Also renames OG, REQ to: `KeyProvider` and  `KeyRequester` which are more self/explanatory

PR #2162, opened 2024-11-26
Generate a certificate inside the enclave

zkokelj:
### Why this change is needed

We want to guarantee e2e encryption between the user and the gateway.  To have this we need to terminate our https connection inside the enclave and this means also generating a certificate within the enclave.

- https server instead of http server
- 2 new parameters for the gateway: enableTLS, tlsDomain
- updated deployment script
- generating and auto renewal of certificates inside the enclave

#### Description of what is happening behind the scenes in library

We use `autocert` library and call `GetCertificate` to obtain the certificate.
If check if the cert already exists and proceeds to obtain it only in case it doesn't.

Generating a certificate consists of the following steps:
- generating a new private key
- domain ownership verification (HTTP-01 challenge since certManager.HTTPHandler(nil) is set)
-  Certificate Signing Request (CRS) is generated
- CRS is sent to CA (Let's Encrypt) via ACME (automated certificate management environment) client 
- CA verifies CRS and issues a certificate
- we store store certificate and private key (in memory and also in on a mounted volume attached to docker container)

Certificate and private key are stored on a mounted volume (same as sealed key for database key encryption) and this private key is then used also for renewing the certificate and in case of restarts/upgrades of the gateway.

image

### PR checks pre-merging


tudor-malene (approved):
LGTM!

Let's store the cert in the DB in the next PR

PR #2177, opened 2024-12-03
Store certificate and private key in encrypted CosmosDB database

zkokelj:
### Why this change is needed

We want to have static https certificate (using the same private key) even after restarts & certificate renewals etc.
Certificate and private key were previously stored in memory and in file on attached volume. 
This is not ideal as it is stored unencrypted and in case of volume is deleted the private key and certificate are deleted.

Much better solution is to store both private key and certificate in encrypted CosmosDB.

- implementation of `autocert.Cache` in which we store certificate and private key in cosmosDB and encrypt it.
- using this approach only for CosmosDB (for other databases - sqlite, we still use file storage for certificate and private key)
- 

### PR checks pre-merging

### DIFF
--- a/tools/walletextension/storage/database/cosmosdb/cert_storage_cosmos.go
+++ b/tools/walletextension/storage/database/cosmosdb/cert_storage_cosmos.go
@@ -65,9 +65,6 @@ func NewCertStorageCosmosDB(connectionString string, encryptionKey []byte) (*Cer
 func (c *CertStorageCosmosDB) Get(ctx context.Context, key string) ([]byte, error) {
 	keyString, partitionKey := c.dbKey([]byte(key))
 
-	fmt.Println("Get certificate from CosmosDB")
-	fmt.Println("keyString", keyString)
-
 	itemResponse, err := c.certsContainer.ReadItem(ctx, partitionKey, keyString, nil)
 	if err != nil {
 		if strings.Contains(err.Error(), "NotFound") {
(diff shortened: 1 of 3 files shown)


tools/walletextension/storage/database/cosmosdb/cert_storage_cosmos.go
```diff
@@ -0,0 +1,130 @@
...
+	_, err := c.certsContainer.DeleteItem(ctx, partitionKey, keyString, nil)
+	if err != nil && !strings.Contains(err.Error(), "NotFound") {
+		return fmt.Errorf("failed to delete certificate: %w", err)
+	}
+
+	return nil
+}
+
+// dbKey generates a consistent key for CosmosDB storage
+func (c *CertStorageCosmosDB) dbKey(key []byte) (string, azcosmos.PartitionKey) {
```
tudor-malene, line 125: I don't think we need to obfuscate the key. There will be only one

tudor-malene (commented):
lgtm.
minor comment

tools/walletextension/storage/database/cosmosdb/cert_storage_cosmos.go
zkokelj: image

First 2 are old (obfuscated keys) and the last two are keys after the change.

tudor-malene (approved):
lgtm

PR #2187, opened 2024-12-05
Fix network test startup; Wait in a loop for rpc key to be initialised in the gateway

StefanIliev545:
### Why this change is needed

The network test fails whenever the rpc key is not initialised as this is a critical issue instead of waiting.

Made RPC key not initialised non crit issue; Code will now loop and wait for it.

### PR checks pre-merging


StefanIliev545:
Closing as @tudor-malene already has a fix for this

PR #2188, opened 2024-12-06
gethfork: remove broken memsize dependency

BedrockSquirrel:
### Why this change is needed

There's a library called memsize that geth was using which is broken by Go 1.23 and the author of memsize is advising geth forks to remove the dependency with this patch: https://github.com/ethereum/go-ethereum/commit/e4675771eda550e7eeb63a8884816982c1980644

Remove memsize dep.

### PR checks pre-merging


tudor-malene (approved):
lgtm.
Nice find

PR #2191, opened 2024-12-09
refresh when new account is added to improve UX

zkokelj:
### Why this change is needed

In out current frontend code we don't refresh the page after adding new account.  Instead it shows us the page like we never added a network, etc.

PR with video: https://github.com/ten-protocol/ten-internal/issues/3567

- Improved frontend code with page refresh when new account is added to update everything needed.

### PR checks pre-merging

### DIFF
--- a/tools/walletextension/frontend/src/components/providers/wallet-provider.tsx
+++ b/tools/walletextension/frontend/src/components/providers/wallet-provider.tsx
@@ -67,9 +67,9 @@ export const WalletConnectionProvider = ({
       setVersion(await fetchVersion());
     } catch (error) {
       showToast(
-          ToastType.DESTRUCTIVE,
-      error instanceof Error ? error.message : "Error initializing wallet connection. Please refresh the page."
-    );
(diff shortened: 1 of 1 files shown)


tools/walletextension/frontend/src/components/providers/wallet-provider.tsx
```diff
@@ -189,16 +189,26 @@ export const WalletConnectionProvider = ({
...
       initialize(providerInstance);
 
-      ethereum.on("accountsChanged", fetchUserAccounts);
+      const handleAccountsChanged = async (accounts: string[]) => {
+        if (accounts.length === 0) {
+          setAccounts(null);
+          setWalletConnected(false);
+          setToken("");
+        } else {
+          window.location.reload();
```
badgersrus, line 198: ive had this cause issues with wallet connections in the past so you might wanna consider using `setAccounts(accounts)` and `setWalletConnected(true)` and then using that as a hook?

badgersrus (approved):
lgtm as going to be replaced by a new frontend

PR #2195, opened 2024-12-11
fix websockets not connecting via secure wss

zkokelj:
### Why this change is needed

After using separate domains for backend and frontend we are connecting to the backend directly (to VM) and the traffic doesn't go through Azure Application Gateway anymore. This gateway was handling TLS termination for websockets before.

- Using wss in case we have TLSConfig and https connection
- Using the same port for wss and https as is the standard

### PR checks pre-merging


moraygrieve (approved):
LGTM

PR #2196, opened 2024-12-11
Add Input (optional) to Skip Deployment and Perform Build Only - Gateway Frontend

pkrishnath:
### Why this change is needed

The current GitHub Action is configured to perform both build and deployment steps, but for Kubernetes deployments, the deployment step is unnecessary and can be removed to simplify the workflow.

GH Action workflow file 

### DIFF
--- a/.github/workflows/manual-deploy-ten-gateway-frontend.yml
+++ b/.github/workflows/manual-deploy-ten-gateway-frontend.yml
@@ -1,5 +1,5 @@
 name: "[M] Deploy Ten Gateway Frontend"
-run-name: "[M] Deploy Ten Gateway Frontend ( ${{ github.event.inputs.testnet_type }} )"
+run-name: "[M] Deploy Ten Gateway Frontend ( ${{ inputs.testnet_type }} )"
 on:
   workflow_dispatch:
     inputs:
@@ -21,7 +21,7 @@ on:
           - "primary"
           - "DEXYNTH"
       deploy_to_azure:
-        description: "Deploy to Azure"
+        description: "Release to ACR"
         required: true
         default: "true"
         type: choice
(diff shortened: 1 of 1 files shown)


.github/workflows/manual-deploy-ten-gateway-frontend.yml
```diff
@@ -1,16 +1,5 @@
-# Deploys TEN Gateway Frontend on Azure for Testnet
-# Builds the TEN Gateway image, pushes the image to dockerhub and starts the TEN Gateway on Azure VM
```
zkokelj, line 2: Why are you removing this comment?

.github/workflows/manual-deploy-ten-gateway-frontend.yml
```diff
@@ -31,46 +20,84 @@ on:
...
+        default: "latest"
+
 jobs:
   validate-inputs:
     runs-on: ubuntu-latest
     steps:
       - name: "Check if deployment is allowed"
         run: |
-          if [[ "${{ github.event.inputs.instance_type }}" == "DEXYNTH" && "${{ github.event.inputs.testnet_type }}" != "sepolia-testnet" ]]; then
+          if [[ "${{ inputs.instance_type }}" == "DEXYNTH" && "${{ inputs.testnet_type }}" != "sepolia-testnet" ]]; then
```
zkokelj, line 66: are you sure this works? can we just omit `github.event.`?

zkokelj (commented):
General question: why do we need to have an option to build only? do you need those builds in k8s?

Why are you changing other things and not just adding if statements for steps that need to be skipped if we decide to only build an image?

Dave you tried running it on dev-testnet?

.github/workflows/manual-deploy-ten-gateway-frontend.yml
pkrishnath: adding back

pkrishnath:
> General question: why do we need to have an option to build only? do you need those builds in k8s?
> 
> Why are you changing other things and not just adding if statements for steps that need to be skipped if we decide to only build an image?
> 
> Dave you tried running it on dev-testnet?


The current action is configured to build and deploy only for three dedicated environments, and environment variables are restricted to those specific environments. However, we need an option to perform a build using dynamic  environment variables for Kubernetes deployments, as the deployment itself is handled separately in those cases. This ensures flexibility and supports the Kubernetes workflow without altering the existing deployment setup

zkokelj (approved):
LGTM

PR #2200, opened 2024-12-12
remove getStorageAt returning token and use local storage instead 

zkokelj:
### Why this change is needed

At the moment we are using local getStorageAt with argument 0x0 to get userID/token that is used in Metamask.
This is very risky and security vulnerability since any app that we connect to can make this call and it will be able to see token and with it our private data.

- sqlite fix that allows us to run it locally 
- removing getStorageAt part where the gateway returns usedID and test that tests that
- frontend change to use local storage instead of getStorageAt calls

### PR checks pre-merging

### DIFF
--- a/go/common/custom_query_types.go
+++ b/go/common/custom_query_types.go
@@ -19,7 +19,6 @@ import "github.com/ethereum/go-ethereum/common"
 
 // CustomQuery methods
 const (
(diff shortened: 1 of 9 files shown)


tudor-malene (approved):
lgtm

PR #2211, opened 2024-12-17
persistent volume and option to deploy to the same VM

zkokelj:
### Why this change is needed

At the moment when we redeploy TEN Gateway we delete and recreate whole VM which deletes also the volume where the encryption key is sealed. Since we store certificates in cosmosDB encrypted and we don't get the encryption key from previous version of the gateway we cannot read old certificates resulting in certificate error.

New option when deploying to redeploy to the same VM (volume and sealed encryption key stays there) so we can easily redeploy the gateway without issues with the certificate.
Since we don't need to delete and recreate VM on Azure it also reduces deployment time. In case we want to make fresh deployment when we have some larger changes we can still do it by selecting it in github actions.

### PR checks pre-merging

### DIFF
--- a/.github/workflows/manual-deploy-obscuro-gateway.yml
+++ b/.github/workflows/manual-deploy-obscuro-gateway.yml
@@ -35,6 +35,14 @@ on:
         options:
           - "primary"
           - "DEXYNTH"
+      recreate_vm:
+        description: "Delete and recreate VM"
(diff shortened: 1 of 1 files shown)


pkrishnath:
LGTM

pkrishnath (approved):
LGTM

PR #2218, opened 2024-12-20
Add an option to disable certificate encryption

zkokelj:
### Why this change is needed

At the moment we always encrypt our https certificate. In case we need to do hard redeploy (often during testing different things) we need to request the certificate again from LetsEncrypt. It also has some weekly limits that we started to hit.
By using unencrypted certificate it can be used over redeploys even in case when the encryption key is lost/changed due to a redeploy of a whole VM.
This flag should be set to `true` in mainnet, but for testnets we can use the default value (`false`).

- new flag that can disable certificate encryption & decryption on testnets.

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #2222, opened 2024-12-20
Host: only evict HA sequencer enclaves

BedrockSquirrel:
### Why this change is needed

Treating single-enclave nodes and backup enclaves as being critical made some things more fragile than they needed to be.

Also evicting enclaves was happening a bit too quickly, causing race conditions at startup etc.

Slow down eviction. Enclave service only evicts active sequencers (no need for single-enclave validators).

### PR checks pre-merging

### DIFF
--- a/go/host/enclave/guardian.go
+++ b/go/host/enclave/guardian.go
@@ -275,6 +275,7 @@ func (g *Guardian) mainLoop() {
 		g.logger.Trace("mainLoop - enclave status", "status", g.state.GetStatus())
 		switch g.state.GetStatus() {
 		case Disconnected, Unavailable:
+			// todo make this eviction trigger configurable once we've settled on how it should work
 			if unavailableCounter > 10 {
 				// enclave has been unavailable for a while, evict it from the HA pool
 				g.evictEnclaveFromHAPool()
--- a/go/host/enclave/guardian.go
+++ b/go/host/enclave/guardian.go
@@ -275,9 +275,8 @@ func (g *Guardian) mainLoop() {
 		g.logger.Trace("mainLoop - enclave status", "status", g.state.GetStatus())
 		switch g.state.GetStatus() {
 		case Disconnected, Unavailable:
-			if unavailableCounter > 3 {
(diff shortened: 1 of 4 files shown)


go/host/enclave/service.go
```diff
@@ -122,7 +122,15 @@ func (e *Service) GetEnclaveClients() []common.Enclave {
 	return clients
 }
 
-func (e *Service) EvictEnclave(enclaveID *common.EnclaveID) {
+func (e *Service) NotifyUnavailable(enclaveID *common.EnclaveID) {
+	if len(e.enclaveGuardians) <= 1 {
+		e.logger.Info("not running in HA mode, no need to evict enclave", log.EnclaveIDKey, enclaveID)
+		return
+	}
+	if *e.activeSequencerID != *enclaveID {
```
badgersrus, line 130: any reason we can't compare the actual values here instead?

badgersrus (approved):
couple of comments, feel free to ignore and lgtm!

go/host/enclave/guardian.go
```diff
@@ -275,9 +275,8 @@ func (g *Guardian) mainLoop() {
 		g.logger.Trace("mainLoop - enclave status", "status", g.state.GetStatus())
 		switch g.state.GetStatus() {
 		case Disconnected, Unavailable:
-			if unavailableCounter > 3 {
+			if unavailableCounter > 10 {
```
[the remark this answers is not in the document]
BedrockSquirrel: Yeah agree but still a bit finger in the air on all this stuff (like it might end up being that we want to evict based on a timeout or just on failed batch submission). I'll add a todo to be configurable once we're happy with the trigger.

go/host/enclave/service.go
BedrockSquirrel: I think this does compare the values (although I haven't fully internalised the quirks around this tbh lol). I had it as `e.activeSequencerID == enclaveID` below and Tudor fixed it, think it was comparing the pointers with that.

PR #2226, opened 2024-12-24
ETag based retry logic for user updates

zkokelj:
### Why this change is needed

If the same user wants to add 2 users at the same time / quick succession we have a race condition and we only write on the of new values (the first one) to the database.

- new `updateUserWithRetries` function performs a read–mutate–write with ETag checks in a loop..
If a “Precondition Failed” response arises due to an ETag mismatch, the code now re-reads the latest user data, merges changes, and retries the write, ensuring both updates are preserved.

### PR checks pre-merging


tudor-malene (approved):
LGTM

PR #2230, opened 2024-12-30
Refactor batch headers to remove deprecated stuff.

StefanIliev545:
### Why this change is needed

There are a few fields that are artifacts of the initial xchain implementation. 

Removed the old fields; Ensured they are not used anywhere anymore.

### PR checks pre-merging


tudor-malene (approved):
LGTM - as a first step. 
In a future PR, we should get rid of the tree because it doesn't really add any value in the header.

PR #2233, opened 2025-01-03
Sequencer: require L1-permissioned enlave ID

BedrockSquirrel:
### Why this change is needed

We want validators to require batches and rollups be produced by permissioned Enclave IDs, so we need sequencers to wait until they're permissioned.

This permissioning tells them they are a sequencer, having 'MakeActive' called by their host puts them into active sequencer mode.

Changes to host, enclave and tests to make sequencer role be driven by L1 protocol level permissioning instead of config.

### PR checks pre-merging


tudor-malene (commented):
LGTM

PR #2237, opened 2025-01-06
Fix the off by one bug in public callbacks

StefanIliev545:
### Why this change is needed

Public callbacks return the wrong callback id; This fixes it

Changed the return value of addCallback

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #2246, opened 2025-01-13
add an option to disable caching in the gateway

zkokelj:
### Why this change is needed

In tests we would sometimes like to run the gateway without caching so that every request goes to the validator/sequencer directly.

- add a flag to disable caching & add it to the config
- add a NonOpCache and use it when caching is disabled

### PR checks pre-merging


badgersrus (approved):
lgtm

PR #2251, opened 2025-01-17
Mock Signature Verificaiton

badgersrus:
### Why this change is needed

https://github.com/ten-protocol/ten-internal/issues/4437

* new mock signature verification and check for in memory DB to set it 

### PR checks pre-merging

### DIFF
--- a/go/enclave/enclave_admin_service.go
+++ b/go/enclave/enclave_admin_service.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/ten-protocol/go-ten/integration/ethereummock"
 	"math/big"
 	"sync"
 	"time"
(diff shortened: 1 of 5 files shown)


go/enclave/enclave_admin_service.go
```diff
@@ -600,3 +601,10 @@ func exportCrossChainData(ctx context.Context, storage storage.Storage, fromSeqN
 	} // todo: check fromSeqNo
 	return bundle, nil
 }
+
+func getSignatureValidator(useInMemDB bool, storage storage.Storage) (components.SequencerSignatureVerifier, error) {
+	if useInMemDB {
```
BedrockSquirrel, line 606: Hopefully we have some checks in place that would barf for this but if we're in proper, SGX, attesting mode we definitely shouldn't be able to set that flag to true. In prod we'll lock it down with the enclave.json

PR #2252, opened 2025-01-17
WIP - signing over blob hash/ messages root

badgersrus:
### Why this change is needed

https://github.com/ten-protocol/ten-internal/issues/4450

Currently, the rollup submission process has several security concerns:
1. The hash is user-provided when submitting a rollup
2. Blobs are created by the host and signed there
3. The Management contract's validation is limited to verifying a signature from a sequencer, which could be replayed
4. The signing process needs better structure and security (signature should be over combined hash of blob hash, cross chain messages and block binding)

- Management contract changes:
Modified `AddRollup` function to use `blobhash` opcode,  validate composite hash and verify sequencer signature over this composite hash.

- Modified `CreateRollup`
To create rollups inside the enclave, compute blob hash using KZG commitment, sign the composite hash.

### Security Improvements
1. Blob creation and hash computation now happens entirely in the enclave
2. Signature includes all critical elements (blob hash, block binding, sequence number)
3. Contract verification matches enclave signature computation
4. Eliminated possibility of replay attacks through block binding

### TODO/Future Work
1. Implement cross-chain message root (currently set to zero)
 

### PR checks pre-merging


go/ethadapter/blob.go
```diff
@@ -154,3 +154,17 @@ func ReconstructRollup(blobs []*kzg4844.Blob) (*common.ExtRollup, error) {
 
 	return &rollup, nil
 }
+
+// ComputeBlobHash computes the versioned hash of a blob that matches the blobhash opcode
+func ComputeBlobHash(blob *kzg4844.Blob) (gethcommon.Hash, error) {
```
badgersrus, line 159: this appears to be unused

go/ethadapter/mgmtcontractlib/mgmt_contract_lib.go
```diff
@@ -118,31 +119,55 @@ func (c *contractLibImpl) DecodeTx(tx *types.Transaction) common.L1TenTransactio
...
+		return nil, fmt.Errorf("failed to encode rollup: %w", err)
+	}
+
+	blobs, err := ethadapter.EncodeBlobs(rollupData)
+	if err != nil {
+		return nil, fmt.Errorf("failed to encode rollup to blobs: %w", err)
+	}
+
+	// Verify the blob hash matches what was signed
+	commitment, err := kzg4844.BlobToCommitment(blobs[0])
```
badgersrus, line 137: there might be a security vulnerability here since we're recreating the blobs that have already been create by the sequencer, might be worth getting @StefanIliev545 to have a look. checking the hash should work but feels wrong. the alternative is that we send the blobs back over GRPC in something like this:

```
struct ExtRollupAndBlobs {
    Rollup *ExtRollup
    Blobs []*kzg4844.Blob
}
```

and then `PublishRollup(producedRollup *common.ExtRollupAndBlobs)`

go/ethadapter/mgmtcontractlib/mgmt_contract_lib.go
```diff
@@ -118,31 +119,55 @@ func (c *contractLibImpl) DecodeTx(tx *types.Transaction) common.L1TenTransactio
...
-	if err != nil {
-		return nil, fmt.Errorf("failed to encode rollup to blobs: %w", err)
+		return nil, err
 	}
 
-	var blobHashes []gethcommon.Hash
 	var sidecar *types.BlobTxSidecar
+	var blobHashes []gethcommon.Hash
 
+	// Use the pre-created blobs from the enclave
```
badgersrus, line 170: these are not the ones created in the sequencer, these have been recreated and we are just checking the hashes are the same

zkokelj: fixed the comment

go/ethadapter/blob.go
zkokelj: removed unused code

PR #2254, opened 2025-01-17
Remove active sequencer nodetype

BedrockSquirrel:
### Why this change is needed

We have some annoying scenarios with the sequencer failover that are going to make thing fragile in the short term unless we address them.

It is problematic that the sequencer enclaves persist whether they are active or not. It means if we failover then we have two active enclaves and the first one needs to be destroyed. This is not ideal while we are running with the current setup of host + 2xenclaves in a VM, the old turn it off and on again trick would no longer be available to us.

So I propose that enclaves do not persist that they are active. By the time a healthy one has restarted it would be treated as dead by the HA pool anyway so the persistence is never going to be useful. **We could eventually check against mgmt contract or historical data that the sequencer ID has never published before** before allowing to become active if that was the concern.

Eventually enclaves will be discarded/permanently excluded once they go down. The host will be able to start managing new enclaves dynamically without requiring initial config for them or a restart. But even in that end-state it doesn't help for the enclaves to persist that they were active I don't think.

Remove active/backup persisted node type separation. Introduce boolean for when the enclave is active.

### PR checks pre-merging


go/enclave/enclave_admin_service.go
```diff
@@ -119,17 +120,15 @@ func NewEnclaveAdminAPI(config *enclaveconfig.EnclaveConfig, storage storage.Sto
 	if eas.isBackupSequencer(context.Background()) || eas.isActiveSequencer(context.Background()) {
 		mempool.SetValidateMode(false)
 	}
-	if eas.isActiveSequencer(context.Background()) {
```
BedrockSquirrel, line 122: Sequencer is never going to be active when it first starts up so we don't check this in constructor.

tudor-malene (approved):
lgtm

PR #2255, opened 2025-01-20
use double hash for userID logging

zkokelj:
### Why this change is needed

At the moment we log userIDs to logs in plain text. This is a security issue as even we shouldn't be able to see that userIDs are.

https://github.com/ten-protocol/ten-internal/issues/4523

- Double hash userIDs before logging them
Double hashing prevents us or anyone who can see the logs from getting userID and being able to see what is happening with this user. 
Using double hash still gives us some information about which user is performing which action so we are still able to debug things.

### PR checks pre-merging

### DIFF
--- a/tools/walletextension/common/common.go
+++ b/tools/walletextension/common/common.go
@@ -2,6 +2,8 @@ package common
 
 import (
 	"crypto/rand"
+	"crypto/sha256"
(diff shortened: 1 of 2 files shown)


tudor-malene (approved):
lgtm

PR #2256, opened 2025-01-20
Don't crash the gateway if TEN node is unavailable

zkokelj:
### Why this change is needed

At the moment TEN Gateway crashes if TEN node is not available for longer than 20 minutes. We then need to restart both TEN node and gateway. We would like for the gateway to keep running and try to connect periodically.

- `NewExponentialBackoffInfiniteStrategy` - new retry strategy where we back off the next retry exponentially with max limit set. This is useful not to overload node with too frequent requests. It also ensures that we never stop trying until the node is reached.
- Removed 30s limit at startup in which the node needs to be available otherwise the gateway crashes
- Use logging instead of direct prints + change Obscuro -> TEN

### PR checks pre-merging


BedrockSquirrel (approved):
lgtm

PR #2270, opened 2025-01-27
HA: fixes and testnet env configuration

BedrockSquirrel:
### Why this change is needed

We had issues with VMs not being big enough and various timing, synchronization issues. As well as a recent docker API versioning issue that needs fixing on main.

- all testnet envs to have 2 enclaves in the sequencer node by default
- sequencer needs to run on newer hardware VM (same price on azure) which has double the mem
- host should only promote an enclave that has consumed up to the latest batch
- workaround docker API incompatibility issue

### PR checks pre-merging


go/host/enclave/guardian.go
```diff
@@ -199,6 +199,9 @@ func (g *Guardian) PromoteToActiveSequencer() error {
 		g.logger.Error("Unable to promote to active sequencer, already active")
 		return nil
 	}
+	if g.state.enclaveL2Head.Cmp(big.NewInt(0)) > 0 && !g.state.IsLive() {
```
badgersrus, line 202: is there an failure mode here where the backup sequencer never catches up with the L2 due to whatever killed the first enclave (bad batch or something) and never gets promoted?

badgersrus (approved):
lgtm - one comment around failure modes but looks gtg

go/host/enclave/guardian.go
BedrockSquirrel: Yeah that's something to consider. But that's gonna require human intervention, the idea is that any batch in the host's database (which the enclave needs to be considered 'live') is considered to have been broadcast to the whole network.

If we decide we need to rewind that batch somehow and basically say its transactions never happened we'd have to delete it from the host DB and do some gymnastics to keep the validators happy. Might even require certifying a new version of the enclave image which had the bad batch surgery baked into it.

We should discuss this sort of stuff in our red teaming/security/bricking discussions.

PR #2277, opened 2025-01-29
releases/v1.0 Rollup block timing fix 

badgersrus:
### Why this change is needed

We are trying to calculate the blockhash using the same block number as the current block.

* Poll the block number in publisher to wait for the next block 
* Do not allow block binding number that are equal to current block in mgmt contract

### PR checks pre-merging


tudor-malene (approved):
lgtm.
nice

PR #2287, opened 2025-02-02
Fix for adding rollups in testnet deployments using postgres

StefanIliev545:
### Why this change is needed

The new database stuff for crosschain messages requires id to be returned; Previously it was using lastinsertedId which is not available in postgres.

Added a change; Query now returns and the return is scanned in both sqllite and postgres as this is the compatible approach

### PR checks pre-merging


StefanIliev545:
https://github.com/ten-protocol/ten-test/actions/runs/13101614539/job/36550557459
https://github.com/ten-protocol/ten-test/actions/runs/13101591531

tudor-malene (approved):
lgtm

PR #2299, opened 2025-02-06
Enclave: avoid log spam during initial L1 catchup

BedrockSquirrel:
### Why this change is needed

This info level log msg was getting spammed by enclave during initial L1 catchup of a new node (because it doesn't see its own attestation until it gets close to the latest L1 blocks, it was only just sent).

Downgrade spammy log msg

### PR checks pre-merging


badgersrus (approved):
lgtm

PR #2310, opened 2025-02-12
Debug logging for rollups to see if we're losing any

badgersrus:
### Why this change is needed

Rollups not being found in blocks on k8s validator and no way to determine why currently. 

* Debug logging on the host and enclave side

### PR checks pre-merging

### DIFF
--- a/go/enclave/components/rollup_consumer.go
+++ b/go/enclave/components/rollup_consumer.go
@@ -159,11 +159,13 @@ func (rc *rollupConsumerImpl) extractAndVerifyRollups(processed *common.Processe
 			rc.logger.Warn(fmt.Sprintf("could not decode tx at index %d. Cause: %s", i, err))
 		}
 		if t == nil {
+			rc.logger.Debug(fmt.Sprintf("rollup event was provided with a tx that is not a rollup."))
 			continue
 		}
 
 		rollupHashes, ok := t.(*common.L1RollupHashes)
 		if !ok {
+			rc.logger.Debug(fmt.Sprintf("could not cast decoded transaction to l1 rollup hashes. Cause: %s", err))
 			continue
 		}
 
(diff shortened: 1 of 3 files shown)


go/enclave/components/rollup_consumer.go
```diff
@@ -159,11 +159,13 @@ func (rc *rollupConsumerImpl) extractAndVerifyRollups(processed *common.Processe
 			rc.logger.Warn(fmt.Sprintf("could not decode tx at index %d. Cause: %s", i, err))
 		}
 		if t == nil {
+			rc.logger.Debug(fmt.Sprintf("rollup event was provided with a tx that is not a rollup."))
 			continue
 		}
 
 		rollupHashes, ok := t.(*common.L1RollupHashes)
 		if !ok {
+			rc.logger.Debug(fmt.Sprintf("could not cast decoded transaction to l1 rollup hashes. Cause: %s", err))
```
BedrockSquirrel, line 168: Slightly prefer `.Debug("could not cast decoded transaction to l1 rollup hashes", log.ErrKey, err)

but I'm guessing AI put in those Sprintf's because it saw some others in the file already so fair enough haha.

BedrockSquirrel (approved):
LGTM, can ignore the comment, just an observation

go/host/l1/dataservice.go
```diff
@@ -303,6 +303,8 @@ func (r *DataService) processRollupLogs(l types.Log, txData *common.L1TxData, pr
 			},
 		}
 		processed.AddEvent(common.RollupTx, txData)
+	} else {
+		r.logger.Debug(fmt.Sprintf("error while fetching blobs. Cause: %s", err))
```
BedrockSquirrel, line 307: Actually this is no good, the err is defined in the if statement so it's out of scope at this point, you'll have the nil value from earlier on. Need to take that FetchBlobs call out of the if statement and do it in the natural way I think

BedrockSquirrel (changes requested):
Need to rejig the blob err capture, the if scope  means we won't capture it for logging I think

BedrockSquirrel (approved):

PR #2329, opened 2025-02-18
Fix max retries blob log message

badgersrus:
### Why this change is needed

The log message was showing nil values

* return the pricedTx from `executeTransaction` so we can log the values

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #2339, opened 2025-02-20
Fix docker api version command 

moraygrieve:


BedrockSquirrel (approved):
Oops sorry, my bad! LGTM, could change the step id while you're renaming it too but no biggy

PR #2348, opened 2025-02-24
Fix for eth call depending on gas price.

StefanIliev545:
### Why this change is needed

eth_call with a gas price would trigger noBaseFee to be false, leading to errors about the account having insufficient balance. 

Removed the old logic used for synthetic transactions; 

### PR checks pre-merging


tudor-malene (changes requested):
LGTM, but needs to be against the release branch

PR #2353, opened 2025-02-26
Fix for race condition on underpriced tx

StefanIliev545:
### Why this change is needed

Both containers start at the same time with diff hardhat instances using hte same private keys and sending transactions.

Made the process sequential as it should be. 
### PR checks pre-merging


moraygrieve (approved):
LGTM

badgersrus (approved):
lgtm

PR #2359, opened 2025-03-03
Host: Report seq status from enclave directly

BedrockSquirrel:
### Why this change is needed

If host was restarted then one enclave would still think it is the active sequencer but the host wouldn't know and all hell breaks loose.

This avoids any syncing issues by using the enclave as the source of truth for whether it is the active sequencer.

### PR checks pre-merging


BedrockSquirrel:
replaced by https://github.com/ten-protocol/go-ten/pull/2376

PR #2366, opened 2025-03-05
Fix block/ rollup error processing

badgersrus:
…critical errors

### Why this change is needed

There's a bug where we return a non-critical error but don't return the `ingestion` result in the block processing code causing a NPE. 

Rework the processing order. Only propagate the errors up to the block rejection check if its critical

### PR checks pre-merging

### DIFF
--- a/go/enclave/enclave_admin_service.go
+++ b/go/enclave/enclave_admin_service.go
@@ -172,11 +172,8 @@ func (e *enclaveAdminService) SubmitL1Block(ctx context.Context, blockData *comm
 
 	result, rollupMetadata, err := e.ingestL1Block(ctx, blockData)
 	if err != nil {
-		// only reject the block if it's a critical error ie duplicate block or signed rollup error
-		if e.isCriticalError(err) {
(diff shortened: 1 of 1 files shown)


go/enclave/enclave_admin_service.go
```diff
@@ -516,7 +512,8 @@ func (e *enclaveAdminService) ingestL1Block(ctx context.Context, processed *comm
 	var rollupMetadataList []common.ExtRollupMetadata
 	if processed.HasEvents(common.RollupTx) {
 		rollupMetadataList, err = e.processRollups(ctx, processed, rollupMetadataList)
-		if err != nil {
+		if err != nil && e.isCriticalError(err) {
+			// only propagate the error if we encounter a critical error on a sequencer signed rollup
```
BedrockSquirrel, line 516: We should still log it if it's not critical I guess

BedrockSquirrel (changes requested):

BedrockSquirrel (approved):
lgtm

PR #2372, opened 2025-03-10
use shorter error for user not found

zkokelj:
### Why this change is needed

We saw some error logs that were very long because of the CosmosDB error message. It was confusing and appeared as there is something wrong with CosmosDB instance.
It turns out that just someone tried to make a call with non existing user (or old user - before migration to cosmosDB).. we can expect more errors like those in the future so I am creating smaller error message in this case.

Check error message and if it is the case that it is "Not found" error message, then we should print custom error, otherwise print full error as it might indicate other issues with CosmosDB.

### PR checks pre-merging


tudor-malene (approved):
lgtm.
We have to choose between this or https://github.com/ten-protocol/go-ten/pull/2371

PR #2390, opened 2025-03-17
Guardian: check batch seq number before submitting

BedrockSquirrel:
### Why this change is needed

Validators were often logging an error: `could not find previous batch with seq`

This error will happen randomly during normal running depending on processing times etc. and the guardian will catch-up again. But there's no need for that batch to have been sent to the enclave at all, the guardian already knew it would be rejected.

Make the guardian check stricter so batches that are too new won't be submitted to the enclave.

### PR checks pre-merging


tudor-malene (approved):
nice. lgtm

PR #2394, opened 2025-03-19
Host: handle promotion for single-enclave node recovery

BedrockSquirrel:
### Why this change is needed

This change addresses an issue with single enclave nodes where the host would not promote the enclave after it was restarted.

Host changed to behave the same with promoting dead enclaves, regardless of how many enclaves are configured.

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #2402, opened 2025-03-21
Add set state root manager role to deployment scripts

badgersrus:
### Why this change is needed

As part of the management contract refactor we now have an additional step where we need to grant the DARegistry contract the permission to add state roots to the MerkleMessageBus contract. 

### PR checks pre-merging


StefanIliev545 (approved):
LGTM

PR #2403, opened 2025-03-23
Add decommission workflows

moraygrieve:
### Why this change is needed

Adds workflows to allow us to decommission an L1 or L2 network on UAT and Dev to allow us to save costs when not requiring them to be running.


BedrockSquirrel (approved):
LGTM, can't test these until it's merged anyway can you iirc

PR #2419, opened 2025-03-31
Host: include estimated transactions size in batches table

BedrockSquirrel:
This PR adds a new `txs_size` column to the `batch_host` table, modifying database schema and related SQL statements.

This allows us to more accurately estimate roll-up size.

image

Now we see that rollups are being triggered by interval timeout rather than size caps but the size is non-zero and the queries are working. There will still be tuning required of course but that's config based.


StefanIliev545 (approved):

tudor-malene (approved):
lgtm

PR #2421, opened 2025-04-01
Short term fix for issue in Metamask v12.14+ which prevented accounts…

pbienek:
### Why this change is needed
Metamask v12.14.* is throwing an error (l is not a function) when adding attempting to add a TEN chain to the wallet via `wallet_addEthereumChain`. While the network is still added to the wallet, the error prevents further methods firing which allow the user to add accounts to the chain. 

When this specific error message is returned after calling wallet_addEthereumChain we ignore it proceed connecting accounts.

This is intended as a short term fix until a thorough investigation into the cause of the issue can be carried out. 

### PR checks pre-merging

### DIFF
--- a/tools/walletextension/frontend/src/services/useGatewayService.ts
+++ b/tools/walletextension/frontend/src/services/useGatewayService.ts
@@ -35,14 +35,18 @@ const useGatewayService = () => {
 
   const finaliseConnectAccounts = async () => {
     const metamaskConnected = await isMetamaskConnected()
+    try {
+      if (!metamaskConnected) {
+        showToast(ToastType.INFO, "No accounts found, connecting...");
+        await connectAccounts();
+        showToast(ToastType.SUCCESS, "Connected to TEN Testnet");
+      }
 
-    if (!metamaskConnected) {
-      showToast(ToastType.INFO, "No accounts found, connecting...");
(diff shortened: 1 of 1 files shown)


tools/walletextension/frontend/src/services/useGatewayService.ts
```diff
@@ -67,15 +79,16 @@ const useGatewayService = () => {
...
-        await connectAccounts();
-        showToast(ToastType.SUCCESS, "Connected to TEN Testnet");
-      }
-      await fetchUserAccounts();
+      await finaliseConnectAccounts()
+
     } catch (error: Error | any) {
-      showToast(ToastType.DESTRUCTIVE, `${error?.message}`);
-      throw error;
+      if (error.message === 'l is not a function') {
```
BedrockSquirrel, line 85: Maybe worth a comment on why this hack was necessary if there's any context we could give to it.

BedrockSquirrel: Ah also if finaliseConnectAccounts() threw an error  at this point would it fail silently because we don't catch it? Maybe worth a catch and log.

BedrockSquirrel (approved):
LGTM, optional comments

PR #2424, opened 2025-04-07
add verifying contract to gateway eip712 message

zkokelj:
### Why this change is needed

With recent Metamask update field `verifyingContract` became mandatory. 
In the past we used EIP-712 message without it, because no contract is verifying it and we only check it in the gateway.

We added `verifyingContract` to both FE and BE inside EIP-712 message.

Verifying contract is set to `0x0` address. Is there any good reason why we should use the address of `management contract`.

### PR checks pre-merging

### DIFF
--- a/tools/walletextension/frontend/src/services/useGatewayService.ts
+++ b/tools/walletextension/frontend/src/services/useGatewayService.ts
@@ -89,7 +89,7 @@ const useGatewayService = () => {
       // MetaMask v12.14.* is throwing an unexpected error while adding a network to.
       // Despite the error the network is successfully added.
       // This is a temporary workaround.
-      if (error.message === 'l is not a function' || error.message === 'h is not a function') {
+      if (error.message.includes(' is not a function')) {
         await finaliseConnectAccounts()
       } else {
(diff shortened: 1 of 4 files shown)


tools/walletextension/frontend/src/services/useGatewayService.ts
```diff
@@ -89,7 +89,7 @@ const useGatewayService = () => {
       // MetaMask v12.14.* is throwing an unexpected error while adding a network to.
       // Despite the error the network is successfully added.
       // This is a temporary workaround.
-      if (error.message === 'l is not a function') {
+      if (error.message === 'l is not a function' || error.message === 'h is not a function') {
```
pbienek, line 92: A more longer term solution here is something like `if (error.message.includes(' is not a function')) {...`. This will catch this issue in the event the JS code around whatever is causing this error in metamask changes again and in the magnification process is assigned another shortened variable.

zkokelj: Nice one, thanks. Pushed a new commit with a fix.

pbienek (approved):

PR #2426, opened 2025-04-10
add testnet.pem to Gateway to allow reading sealed encryption key 

zkokelj:
### Why this change is needed

Before that testnet.pem was missing and was generated for each docker build.
This made the signerID different for each build and did not allow new enclaves to read old sealed encryption key.

Another change (in the second commit) is that we need to explicitly set if we want new encryption key (set exchangeURL parameter to `new` or if we want to exchange with `key exchange`).
By default (if that value is left empty) the gateway tries to read the sealed key from mounted volume.

### PR checks pre-merging


BedrockSquirrel (approved):
LGTM but confused as to how this worked before.

And we should create a ticket to get this pem and the main node signer key secured soon (if there isn't one already). Need to figure out the signing process for mainnet.

BedrockSquirrel (approved):
LGTM

PR #2434, opened 2025-04-17
Contracts Upgrade Design

badgersrus:
### Why this change is needed

### PR checks pre-merging


contracts/src/l1_management/contracts/NetworkConfig.sol
```diff
@@ -234,4 +299,66 @@ contract NetworkConfig is Initializable, OwnableUpgradeable {
...
+        emit ContractUpgraded(name, oldVersion, version, implementation);
+    }
+
+    /**
+     * @dev Records multiple contract upgrades in a batch
+     * @param names Array of contract names
+     * @param versions Array of new versions
+     * @param implementations Array of new implementation addresses
+     */
+    function recordBatchUpgrade(
```
StefanIliev545, line 332: I was imagining something simpler - string hardfork name and just emitting when an upgrade has happened; callable only by owner. This is fine i guess, but it doesn't bring in any trust benefits as implementations are still passed as calldata so we can say we've upgraded without any verifiability. 

I guess if we were to go down this approach it would be good to call this function with addresses of proxies, which are queried for their implementation address and then thats emitted - if we wanted the l2 to have some checks around contract correctness.

design/scratchpad/Contract_upgrade.md
```diff
@@ -0,0 +1,170 @@
...
+
+func NewForkManager(logger *log.Logger) *ForkManager {
+    return &ForkManager{
+        forks: make(map[string]*ForkConfig),
+        logger: logger,
+    }
+}
+
+// RegisterFork registers a new fork configuration
+func (fm *ForkManager) RegisterFork(name string, blockNumber uint64, useNewBridge bool) {
```
StefanIliev545, line 88: I think those must be hardcoded as it is usually in ethland. L1 contract emits upgraded to "hardfork-name" and then this manager does activateFork on that name or panics

StefanIliev545 (approved):

PR #2452, opened 2025-05-05
Fix gateway encryption key handling

zkokelj:
### Why this change is needed

It was confusing which value was read and there were some issues in the recent version of the code preventing gateway to get the keys we wanted.

### 1. Renamed Configuration Parameter
- Changed `KeyExchangeURL` to `EncryptionKeySource` for better clarity of purpose
- The new parameter name better reflects its role in the key management process

### 2. Enhanced Key Source Logic
The `EncryptionKeySource` parameter now supports three distinct modes:

#### a) Empty String (`""`)
- Attempts to unseal an existing encryption key
- If no key is found, fails with a clear error message
- Requires operator intervention to either:
  - Provide a key source
  - Generate a new key

#### b) "new" Value
- First checks for an existing sealed key
- Only generates a new key if no existing key is found
- Prevents unnecessary key regeneration
- Preserves existing keys when possible

#### c) URL Value
- Performs secure key exchange with the specified gateway
- Fails explicitly if key exchange is unsuccessful

### PR checks pre-merging

### DIFF
--- a/tools/walletextension/keymanager/keymanager.go
+++ b/tools/walletextension/keymanager/keymanager.go
@@ -53,49 +53,9 @@ func GetEncryptionKey(config common.Config, logger gethlog.Logger) ([]byte, erro
 		return nil, nil
 	}
 
-	if config.EncryptionKeySource != "" {
-		var encryptionKey []byte
-		var err error
-
-		// If the "new" keyword is used for the encryptionKeySource, we first check if there is an existing encryption key
-		// that can be unsealed and used. If no such key is found, we proceed to generate a new random encryption key.
-		// This ensures that we do not overwrite an existing key unless necessary, and a new key is only generated when
-		// there is no existing key available.
-		if config.EncryptionKeySource == "new" {
-			logger.Info("encryptionKeySource set to 'new' -> checking if there is an existing encryption key that we can use")
-			var found bool
-			encryptionKey, found, _ = tryUnsealKey(encryptionKeyFile, config.InsideEnclave)
-			if !found {
-				logger.Info("No existing encryption key found, generating new random encryption key")
-				encryptionKey, err = common.GenerateRandomKey()
-				if err != nil {
-					logger.Crit("unable to generate random encryption key", log.ErrKey, err)
-					return nil, err
-				}
-			}
-		} else {
-			// Attempt to perform key exchange with the specified key provider.
-			// This step is crucial, and the process should fail if the key exchange is not successful.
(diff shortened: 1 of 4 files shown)


tools/walletextension/keymanager/keymanager.go
```diff
@@ -40,63 +40,63 @@ type KeyExchangeResponse struct {
...
-		return encryptionKey, nil
-	}
-
-	// 3.) If KeyExchangeURL is "new", generate a new key
-	if config.KeyExchangeURL == "new" {
-		encryptionKey, err = common.GenerateRandomKey()
-		if err != nil {
-			logger.Error("unable to generate random encryption key", log.ErrKey, err)
-			return nil, fmt.Errorf("failed to generate random encryption key: %w", err)
+	if config.EncryptionKeySource != "" {
```
BedrockSquirrel, line 56: I'd probably un-nest this big block of code one level by handling `if config.EncryptionKeySource == "" {` first and then the rest of the method knows that's dealt with - but maybe personal preference, feel free to ignore.

BedrockSquirrel (approved):
LGTM - couple optional comments.

~~I'm still a bit confused about why we wouldn't want to try unsealing for the case where config has a URL in it. I guess the idea is the process should only start with that URL once and it expects to be cleared for future restarts but feel like it only makes sense if we expect the key to change and need replacing sometimes.~~

Update after speaking with Ziga: we **do** expect it to change, secondary gateways won't necessarily get redeployed when the primary gets deployed so it's important they always pull the current key. But we seal it anyway in case you need it to become primary in future.

tools/walletextension/keymanager/keymanager.go
```diff
@@ -40,63 +40,63 @@ type KeyExchangeResponse struct {
...
 }
 
-// GetEncryptionKey returns encryption key for the database
-// If we use sqlite database we don't need to do anything since sqlite does not need encryption and is usually running in dev environments / for testing
-// By default, we try to read from the sealed key
-// If KeyExchangeURL equals "new" we generate a new key
-// If KeyExchangeURL is a URL we try to perform key exchange
+// GetEncryptionKey returns the encryption key for the database
+// - If we use an SQLite database, no encryption key is needed as SQLite typically runs in development or testing environments.
+// - If a key exchange URL is provided, attempt to obtain the encryption key from the specified URL.
```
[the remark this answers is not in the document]
zkokelj: fixed.

tools/walletextension/keymanager/keymanager.go
zkokelj: fixed.

BedrockSquirrel (approved):
nice, LGTM

PR #2460, opened 2025-05-12
Read and write L1 config from git

BedrockSquirrel:
### Why this change is needed

In order to communicate between different GitHub Action scripts and Argo CD, we are now using GitHub config repo for the environmental one config. This change commits the git config as part of the L1 contract deployment and makes it so the post node startup script is able to read that config out.

- allows the L1 contract deployer to write the addresses to an environmental yaml file in Git
- has the k8s pre and post node startup GH action scripts write to and read from the git file
- I've added the env secret with the github Personal Access Token to all 3 envs

### PR checks pre-merging


tudor-malene (approved):
lgtm - if it works

PR #2469, opened 2025-05-21
Host: no need to wait for enclave to be up to start host services

BedrockSquirrel:
### Why this change is needed

The enclave connection doesn't actually need to be established at startup, the guardian is responsible for waiting and recovering when the enclave stops and restarts.

I think this code was a relic from a long time back and it's now becoming a pain in recovery situations because if one of the enclaves is unavailable then the host cannot start.

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #2493, opened 2025-06-02
Host: pass health check with any enclave healthy

BedrockSquirrel:
### Why this change is needed

We were hitting an issue with the health check failing, as if the node was down when one of the enclaves is down. When really so long as either of the enclaves or any of the enclaves are healthy, then the health check should pass.

The error message from the failing enclave will still be included in the response, it's just the overallHealth boolean flag at the top level, which can be set to true so long as there's a happy enclave.

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #2498, opened 2025-06-03
Pete/rainbow kit integration

pbienek:
### Why this change is needed

We want the gateway to be a hub for the TEN ecosystem. A place users can start exploring partner dapps as well as onboarding their wallets.

- Updated the aesthetic of the UI. 
- Refactored the wallet onboarding to make use of Wagmi
- Added new features to highlight TEN Dapps and allow the users to view their TEN token

### PR checks pre-merging


tools/walletextension/frontend/src/lib/supportedWallets.ts
```diff
@@ -0,0 +1,25 @@
...
+    //     id: 'com.coinbase.wallet',
+    // },
+    {
+        name: 'Rabby Wallet',
+        logo: '/assets/walletLogos/rabbyLogo.png',
+        supportStatus: '',
+        id: 'io.rabby',
+    },
+    // {
+    //     name: 'Brave Wallet',
```
zkokelj, line 20: Were there any issues with Coinbase and Brave wallets? Or was there a decision not to support them?

pbienek: Yeh were some edge cases that proved a little tricky to resolve so I decided to park them for now.

pbienek:
> * I don't see a way to revoke the token. Was there a decision not to include it in the UI?
> * if I delete the network and then try to add it again it will add the network and read the userID from localStorage (We will change that in the future), but not deleting the token from localStorage will cause issues when testing multiple networks.
>   For example we have uat-network token stored and then we try with sepolia-network, but we still have the token from UAT (and that token is not even valid there)

You can revoke the token here...
Screenshot 2025-06-06 at 14 20 38

pbienek:
> * Just a question: is there a way to set variables to docker ar run time? Previously we had to provide gateway backend url at docker build?

Not in the current implementation because the env vars are only read by next js at build time. I'll need to do some research into how we can change them without needing to rebuild.

PR #2502, opened 2025-06-04
Enclave: only allow one attestation per enclave_id in edb

BedrockSquirrel:
### Why this change is needed

We've seen that sometimes multiple attestations can end up in the database.

This can cause unexpected behaviour and is not desirable. To keep things simple we will guarantee a maximum of one attestation entry per Enclave ID.

### PR checks pre-merging


badgersrus (approved):
nice. lgtm

tudor-malene (approved):
lgtm

PR #2526, opened 2025-06-12
[UI] Fixes bug where when a wallet is connected to Gateway UI but no …

pbienek:
…TEN network is present , it adds the network without the TEN token present in the RPC url when switching.

### Why this change is needed

Fixes https://github.com/ten-protocol/go-ten/issues/2519

TEN Token is now included in network switches if the network has to be added to the wallet. 

### PR checks pre-merging


zkokelj (approved):
LGTM.

But I noticed another issue/different behaviour than before. 
You can decide if this should be part of this PR or a new one.

- I authenticated with TEN Gateway and added accounts
- then I switch to another chain and delete existing TEN network
- then I try to add the network again it adds with the existing token.

The situation above is not ideal in case we want to test things in different testnets.

PR #2538, opened 2025-06-17
Host: more visibility when L1 unavailable

BedrockSquirrel:
### Why this change is needed

We had an issue where the L1 url was unavailable and it took a long time for us to understand what wasn't working. This can also happen if L1 url is misconfigured. (See https://github.com/ten-protocol/ten-internal/issues/4672)

I was planning to fail fast but we use the same logic for initial connect and reconnect and having the host CRIT and restart isn't especially helpful, realised what we actually wanted was some acknowledgement of what was happening in the logs.

Extra logging if L1 client is failing to connect.

### PR checks pre-merging


badgersrus (approved):
lgtm

PR #2547, opened 2025-06-26
Deploy: always upload log artifacts even on failure

BedrockSquirrel:
### Why this change is needed

We were returning successful status code when L1 contract deploy failed to make sure the log artifacts always get captured.

This change allows logs to be captured even on failure so we can fail fast and not make people debug the issue when the nodes eventually fail to start up.

### PR checks pre-merging


testnet/launcher/l1contractdeployer/cmd/main.go
```diff
@@ -33,7 +33,7 @@ func main() {
 	networkConfig, err := l1ContractDeployer.RetrieveL1ContractAddresses()
 	if err != nil {
 		fmt.Printf("unable to fetch l1 contract addresses - %s", err)
-		os.Exit(0)
+		os.Exit(1)
```
BedrockSquirrel, line 36: This was making the step look successful when the contracts deployment had failed.

PR #2579, opened 2025-07-24
DB: fix migration config update to upsert

BedrockSquirrel:
### Why this change is needed

We try to insert the migration version value but that fails after the first time because key already exists, so we should so an update if present.

### PR checks pre-merging


tudor-malene (approved):
lgtm (rename for clarification)

PR #2591, opened 2025-08-08
BatchListing query placeholder fix

badgersrus:
### Why this change is needed

* the postgres placeholder is calculated by position and because we have hash=? and then limit =? size=? it was evaluating to $1, $1, $2 and passing the hash as limit param
* will do a tidier fix as a follow up

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #2595, opened 2025-08-11
Test migration on host DB

badgersrus:
### Why this change is needed

Follow up addressing PR comments on these two:
https://github.com/ten-protocol/go-ten/pull/2584 
https://github.com/ten-protocol/go-ten/pull/2585 

* Homogenise how migrations are applied across enclave/ host on sqlite, postgres and edgeless 
* Add `config` tables to host DBs 
* Move `block_count` table to a new migration script to test it works 
* Move config APIs to `/common` package and make use of `sqlx.Rebind` which replaces query placeholders based on driver name 
* Handle no rows found err on `GetBatchBySequence` call

### PR checks pre-merging

### DIFF
--- a/go/host/storage/hostdb/batch.go
+++ b/go/host/storage/hostdb/batch.go
@@ -209,9 +209,6 @@ func GetBatchByTx(db HostDB, txHash gethcommon.Hash) (*common.PublicBatch, error
 	}
 	extBatch, err := GetBatchBySequenceNumber(db, seqNo)
 	if err != nil {
-		if errors.Is(err, errutil.ErrNotFound) {
-			return nil, err
-		}
 		return nil, fmt.Errorf("failed to fetch ext batch - %w", err)
 	}
 	return toPublicBatch(extBatch), nil


go/host/storage/hostdb/batch.go
```diff
@@ -207,6 +209,9 @@ func GetBatchByTx(db HostDB, txHash gethcommon.Hash) (*common.PublicBatch, error
 	}
 	extBatch, err := GetBatchBySequenceNumber(db, seqNo)
 	if err != nil {
+		if errors.Is(err, errutil.ErrNotFound) {
```
BedrockSquirrel, line 212: `fmt.Errorf("failed to fetch ext batch - %w", err)` this produces a wrapped error such that you can still call errors.Is(ErrNotFound) on the thing it produces and it will still say true because it walks the chain of wrapped errors to check. So I'm not sure why we need to handle the ErrNotFound case separately?

BedrockSquirrel (approved):
lgtm, optional minor comment on check that seems unnecessary but I might be missing something

go/host/storage/hostdb/batch.go
badgersrus: I think you're right so I've reverted that for now, was originally in response to tudors PR comment here https://github.com/ten-protocol/go-ten/pull/2584#discussion_r2262303654

PR #2600, opened 2025-08-14
Check for registered postgres driver

badgersrus:
### Why this change is needed

`panic: sql: Register called twice for driver postgres` error on dev

* Check for presence of postgres driver before attempting to register, do nothing if already registered

### PR checks pre-merging


go/host/storage/init/postgres/postgres.go
```diff
@@ -85,6 +85,15 @@ func CreatePostgresDBConnection(baseURL string, dbName string, logger gethlog.Lo
...
 	driverName := "postgres"
+	// check if driver is already registered
+	drivers := sql.Drivers()
+	for _, driver := range drivers {
+		if driver == driverName {
+			logger.Info("PostgreSQL driver already registered, skipping")
+			return driverName
+		}
+	}
+	// register if it not already present
```
badgersrus, line 96: if for whatever reason the postgres driver isnt registered by default we still need to do it here

PR #2601, opened 2025-08-14
Ziga/gw http cookie frontend

zkokelj:
### Why this change is needed

We want to remove localStorage from being used in gateway frontend for storing the token as it can be accessed with javascript and vulnerable in xss attacks.

Backend changes:

- `/join` endpoint doesn't set cookie anymore
- cookies are set and read by `/set-token` and  `/get-token` endpoints (which are resricted to the frontend url -> added new parameter for the backend)
- cookie names are different for each environment ( based on TLS value used in the backend. example: `gateway_uat_gw_testnet_ten_xyz` instead of `gateway_token`) to avoid overwriting tokens when switching between networks (testnets, mainnet, etc.)

Frontend changes:
- removed local storage
- setting cookie with set-token and reading them with get-token endpoints

### PR checks pre-merging

### DIFF
--- a/tools/walletextension/httpapi/routes.go
+++ b/tools/walletextension/httpapi/routes.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strings"
 
 	tencommon "github.com/ten-protocol/go-ten/go/common"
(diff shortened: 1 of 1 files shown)


pbienek (approved):

BedrockSquirrel (approved):
LGTM

PR #2626, opened 2025-09-01
Remove SQL Statements

badgersrus:
### Why this change is needed

Leftover from the migration changes - we can get rid of the clunky `GetSQLStatements()` because we can use the `sqlx.Rebind` function to replace the placeholders based on driver name

* Remove `sql_statements.go`
* Move queries back into respective files and call `Rebind` on each query

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #2659, opened 2025-09-16
Network tests: configurable chain IDs per env

BedrockSquirrel:
### Why this change is needed

Now we're using a different chain ID per env we need to make sure they are configurable in network tests.

### PR checks pre-merging


BedrockSquirrel:
I guess we'll also want to change `integration.TenChainID` to `5443` at some point and check its usages, but for the network tests that should be non-impacting.

zkokelj (approved):
lgtm

PR #2664, opened 2025-09-17
Deploy: change bridged deployment eth from 0.5 to 0.005

BedrockSquirrel:
### Why this change is needed

Auto-bridging $2k dollars in a mainnet launch dry run is an unnecessary risk.

The last run against sepolia used 0.0003 eth and the gas price in mainnet is normally around 2x sepolia's price so 0.005 allows plenty of cushion in all environments. We can adjust if any problems but less risk to start low.

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #2666, opened 2025-09-18
Enclave: log error instead of crit when enclave waiting for initial batches

BedrockSquirrel:
### Why this change is needed

When enclave doesn't receive its first batches within 2mins of starting up it kills itself with a CRIT message. This can happen for a variety of reasons, especially when node is not starting immediately after L1 contract deployments and has some catching up to do.

I think we originally put this in because when we had some major bugs then the sim tests would hang instead of failing but we don't see them any more (and if we did then better to have timeouts in the test structure rather than in the production code).

The CRITs add noise and timing issues to the node startup and catch-up process.

Instead of CRIT after 2mins, we log how long the enclave has been waiting every 2 mins at ERROR level.

### PR checks pre-merging

### DIFF
--- a/go/enclave/components/txpool.go
+++ b/go/enclave/components/txpool.go
@@ -145,6 +145,7 @@ func (t *TxPool) start() {
 	)
 	defer close(newHeadCh)
 	defer newHeadSub.Unsubscribe()
+	startWait := time.Now()
 	for {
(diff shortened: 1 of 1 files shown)


tudor-malene (approved):
lgtm

PR #2674, opened 2025-09-22
Deploy: check gas prices within tolerance before L1 deploy

BedrockSquirrel:
### Why this change is needed

In mainnet if we accidentally start a deployment when gas prices are high we could waste hundreds of pounds.

This is important while we're testing dry runs of contract deployments.

We fail [like this](https://github.com/ten-protocol/go-ten/actions/runs/17917812030) when network prices are too high:
```
⚠️  Gas price too high: 0.940388626 gwei (limit: 0.1 gwei)
Deployment aborted. Please wait for gas prices to drop or increase the max_gas_gwei limit.
```

Gas limit defaults to 1.5 gwei but can be overridden when running the deployment if we need to allow it higher. Currently (sept 25) it's fairly common to have prices well under 1.5gwei each day.

### PR checks pre-merging


badgersrus (approved):
lgtm - added one question

PR #2697, opened 2025-10-02
read chainID from env variable and use hardcoded as fallback

zkokelj:
### Why this change is needed

When deploying l2 on UAT and other network we want to use the ENV variable set for the chainID.
If env variable is not present we can default to local testnet.

successful run: https://github.com/ten-protocol/go-ten/actions/runs/18195378655

### PR checks pre-merging


badgersrus (approved):
lgtm

PR #2700, opened 2025-10-06
Will/host rollup estimation calc

badgersrus:
### Why this change is needed

The host side rollup estimation is still a little off because its not linear for empty batches vs batches with txs. When the host adds batches we store the `tx_size` as `len(batch.EncryptedTxBlob` which is ~41b for an empty batch, we then apply a compression factor of 0.1 on the host side calculation. The logs show the metadata per empty batch compresses more in reality. 

* Empty batch metadata compresses raw size 5.9b to 0.28b ~4.8% 
* The current calculation compresses 41b to 4b at ~9% 

This leaves lots of space in the rollup, producing them more frequently and wastes gas. 

There are two options: 
1. Apply a larger compression estimation on the host which assumes linear compression between empty vs non-empty batches
2. Reduce the tx_size on the host for empty batches to account for the slightly different compression ratio 

This PR shows option 2 but option 1 would just be to change `batchCompressionFactor` in `service.go` to `0.07`.

* Reduce host tx_size for empty batches. 

### PR checks pre-merging

### DIFF
--- a/go/host/storage/hostdb/batch.go
+++ b/go/host/storage/hostdb/batch.go
@@ -41,13 +41,28 @@ func AddBatch(dbtx *dbTransaction, db HostDB, batch *common.ExtBatch) error {
 		return fmt.Errorf("could not encode L2 transactions: %w", err)
 	}
 
+	// this value is used on host side rollup size estimation
+	txDataSize := 0
+	if len(batch.TxHashes) > 0 {
+		// this will be compressed in the rollup, so apply estimated compression factor
+		txDataSize = len(batch.EncryptedTxBlob)
+	} else {
+		// empty batches still contribute to rollup size through:
+		// - RLP encoded empty transaction array (~3 bytes)
+		// - Time delta (~1-2 bytes when compressed)
+		// - L1 height delta (~1-2 bytes when compressed)
+		// actual data from mainnet: 12,769 empty batches = 26KB rollup = ~2 bytes/batch final
+		// batchCompressionFactor=0.1: 26KB ÷ 0.1 ÷ 12,769 = ~20 bytes raw estimate
+		txDataSize = 20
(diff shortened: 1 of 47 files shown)


go/host/storage/hostdb/batch.go
```diff
@@ -41,13 +41,28 @@ func AddBatch(dbtx *dbTransaction, db HostDB, batch *common.ExtBatch) error {
 		return fmt.Errorf("could not encode L2 transactions: %w", err)
 	}
 
+	// this value is used on host side rollup size estimation
+	txDataSize := 0
```
tudor-malene, line 45: a better name is `compressedSize`

go/host/storage/hostdb/batch.go
```diff
@@ -41,13 +41,28 @@ func AddBatch(dbtx *dbTransaction, db HostDB, batch *common.ExtBatch) error {
...
+	if len(batch.TxHashes) > 0 {
+		// this will be compressed in the rollup, so apply estimated compression factor
+		txDataSize = len(batch.EncryptedTxBlob)
+	} else {
+		// empty batches still contribute to rollup size through:
```
tudor-malene, line 50: Wouldn't it be more intuitive to move this stuff out of th else, and then have:

```golang

...
if len(batch.TxHashes) > 0 {
		// this will be compressed in the rollup, so apply estimated compression factor
		txDataSize += len(batch.EncryptedTxBlob)
	} 
```

tudor-malene: Because the 20bytes is roughly how much the header compresses to.

tudor-malene (commented):
nice one.
Left a few comments

tudor-malene (approved):
lgtm

PR #2701, opened 2025-10-06
Statedb.val is nullable, remove  unused `executor.lastExecutedBatch`

tudor-malene:
### Why this change is needed

to fix bugs

- Statedb.val is nullable
- remove  unused `executor.lastExecutedBatch`

### PR checks pre-merging


StefanIliev545 (approved):
LGTM if the not null removal is correct, tho not sure why we would write nil to the db, rather than removing the key altogether.

tudor-malene:
> LGTM if the not null removal is correct, tho not sure why we would write nil to the db, rather than removing the key altogether.

It's the evm that writes the null, not us

PR #2704, opened 2025-10-06
get the token from path in JSON-RPC

zkokelj:
### Why this change is needed

Some wallets (for example Trust wallet) don't support query parameters in the RPC url. We need to make a change in the gateway to also support tokens as part of the path `v1/<TOKEN>` and not only as query params: `v1/?token=<TOKEN>`.

Added path-token support in the shared HTTP params middleware so JSON-RPC accepts both query and path tokens.

The solution is not ideal as we are modifying code inside the gethfork, but there are no good alternatives:

- inside wallet extension we (without touching gethfork) cannot enable path tokens for JSON-RPC. The JSON-RPC stack reads the token from the request context set by the shared middleware; wallet extension cannot intercept and pass control onward to JSON-RPC after rewriting because theres no “next” handler exposed.

- a catch‑all /v1/ route inside wallet extension would shadow both the JSON-RPC handler and your REST endpoints unless we also implemented full delegation logic, which is effectively a major refactor or a reverse proxy.

- usually a nginx/traefik can do that, but due to the requirement that we run everything inside an enclave this is not an option.

### PR checks pre-merging

### DIFF
--- a/lib/gethfork/node/extract_params_handler.go
+++ b/lib/gethfork/node/extract_params_handler.go
@@ -3,6 +3,7 @@ package node
 import (
 	"context"
 	"net/http"
+	"strings"
 
 	"github.com/ten-protocol/go-ten/lib/gethfork/rpc"
 )
@@ -24,7 +25,41 @@ func newHTTPParamsHandler(exposedParam string, next http.Handler) http.Handler {
 func (handler *httpParamsHandler) ServeHTTP(out http.ResponseWriter, r *http.Request) {
 	q := r.URL.Query()
 	val := q.Get(handler.exposedParam)
+	if val == "" {
+		// Attempt to extract token from path as /v1/<TOKEN>
+		// Do not alter routing; only read the segment.
+		// Expected segments: ["", "v1", "<TOKEN>", ...]
+		parts := strings.Split(r.URL.Path, "/")
+		if len(parts) >= 3 && parts[1] == "v1" {
+			candidate := parts[2]
+			if isLikelyHexToken(candidate) {
+				val = candidate
+			}
+		}
+	}
 	ctx := context.WithValue(r.Context(), rpc.GWTokenKey{}, val)
 	handler.next.ServeHTTP(out, r.WithContext(ctx))
 	handler.next.ServeHTTP(out, r)
+}
+
+// isLikelyHexToken performs a lightweight validation for a user token
+// Accepts both 0x-prefixed (42 chars) and non-prefixed (40 chars) lowercase/uppercase hex
+func isLikelyHexToken(s string) bool {
+	n := len(s)
+	if n != 40 && n != 42 {
+		return false
+	}
+	if n == 42 {
+		if !(strings.HasPrefix(s, "0x") || strings.HasPrefix(s, "0X")) {
+			return false
+		}
+		s = s[2:]
+	}
+	for i := 0; i < len(s); i++ {
+		c := s[i]
+		if !((c >= '0' && c <= '9') || (c >= 'a' && c <= 'f') || (c >= 'A' && c <= 'F')) {
+			return false
+		}
+	}
+	return true
 }
--- a/tools/walletextension/httpapi/utils.go
+++ b/tools/walletextension/httpapi/utils.go
@@ -7,6 +7,7 @@ import (
 	gethlog "github.com/ethereum/go-ethereum/log"
 	"github.com/status-im/keycard-go/hexutils"
 	"github.com/ten-protocol/go-ten/go/common/log"
+	"github.com/ten-protocol/go-ten/lib/gethfork/rpc"
 	"github.com/ten-protocol/go-ten/tools/walletextension/common"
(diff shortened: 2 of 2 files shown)


lib/gethfork/node/extract_params_handler.go
```diff
@@ -24,7 +25,41 @@ func newHTTPParamsHandler(exposedParam string, next http.Handler) http.Handler {
...
+	if n != 40 && n != 42 {
+		return false
+	}
+	if n == 42 {
+		if !(strings.HasPrefix(s, "0x") || strings.HasPrefix(s, "0X")) {
+			return false
+		}
+		s = s[2:]
+	}
+	for i := 0; i < len(s); i++ {
```
tudor-malene, line 58: you can use hex.DecodeString

lib/gethfork/node/extract_params_handler.go
```diff
@@ -24,7 +25,41 @@ func newHTTPParamsHandler(exposedParam string, next http.Handler) http.Handler {
 func (handler *httpParamsHandler) ServeHTTP(out http.ResponseWriter, r *http.Request) {
 	q := r.URL.Query()
 	val := q.Get(handler.exposedParam)
+	if val == "" {
```
tudor-malene, line 28: better use `len(val) == 0 `

lib/gethfork/node/extract_params_handler.go
```diff
@@ -24,7 +25,41 @@ func newHTTPParamsHandler(exposedParam string, next http.Handler) http.Handler {
...
 	ctx := context.WithValue(r.Context(), rpc.GWTokenKey{}, val)
 	handler.next.ServeHTTP(out, r.WithContext(ctx))
 	handler.next.ServeHTTP(out, r)
 }
+
+// isLikelyHexToken performs a lightweight validation for a user token
+// Accepts both 0x-prefixed (42 chars) and non-prefixed (40 chars) lowercase/uppercase hex
+func isLikelyHexToken(s string) bool {
+	n := len(s)
+	if n != 40 && n != 42 {
```
tudor-malene, line 49: we have some constants: MessageUserIDLen and MessageUserIDLenWithPrefix

tools/walletextension/httpapi/utils.go
```diff
@@ -35,6 +36,20 @@ func getUserID(conn UserConn) ([]byte, error) {
 		return nil, fmt.Errorf("wrong length of userID from URL. Got: %d, Expected: %d od %d", len(userID), common.MessageUserIDLenWithPrefix, common.MessageUserIDLen)
 	}
 
+	// fallback: token from request context (set by shared gethfork HTTP params middleware for JSON-RPC)
+	// Note: REST endpoints are served by the mux and typically do NOT pass through that middleware,
+	// so this is mainly relevant for JSON-RPC calls.
+	req := conn.GetHTTPRequest()
+	if req != nil {
+		if token, ok := req.Context().Value(rpc.GWTokenKey{}).(string); ok && token != "" {
```
tudor-malene, line 44: I don't quite understant this. This ``getUserId`` function I think is not using the gethrpc. So you need to parse the request.
If this is incorrect, then why are we reading the query parameter directly and not from the context?

tudor-malene (changes requested):
few comments

lib/gethfork/node/extract_params_handler.go
zkokelj: thanks, fixed

lib/gethfork/node/extract_params_handler.go
zkokelj: fixed

tools/walletextension/httpapi/utils.go
zkokelj: Removed that part. `getUserID` is called only in the 'REST' methods `/v1/authenticate, /v1/query, /v1/revoke` and not when we are calling gethrpc JSON-RPC API.

For `/v1/authenticate, /v1/query, /v1/revoke` we still require a query parameter. They are not called by any wallet, but by users/custom (javascript) code.

For JSON-RPC API we can use /?token=TOKEN or /TOKEN.


+ I also added a test and fixed a few other minor bugs & comments in the code

tools/walletextension/httpapi/routes.go
```diff
@@ -421,8 +421,8 @@ func revokeRequestHandler(walletExt *services.Services, conn UserConn) {
 
 	userID, err := getUserID(conn)
 	if err != nil {
-		handleError(conn, walletExt.Logger(), errors.New("user ('u') not found in query parameters"))
-		walletExt.Logger().Info("user not found in the query params", log.ErrKey, err)
+		handleError(conn, walletExt.Logger(), errors.New("token ('token') not found in query parameters / URL"))
+		walletExt.Logger().Info("token not found in the query params / URL", log.ErrKey, err)
```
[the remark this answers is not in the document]
zkokelj: fixed

PR #2712, opened 2025-10-15
Upgrade geth to support cell proofs EIP-7594

badgersrus:
### Why this change is needed

Osaka/ pectra fork means we need to switch to `BlobSidecarVersion1`

* Upgrade geth to `1.16.4`
* Upgrade prysm to `6.1.2` 
* Use cell proofs per blob which proves a selection of points on the polynomial, instead of proof of the full commitment in `BlobSidecarVersion0`

### PR checks pre-merging

### DIFF
--- a/go/ethadapter/blob.go
+++ b/go/ethadapter/blob.go
@@ -24,20 +24,25 @@ const (
 )
 
 // MakeSidecar builds & returns the BlobTxSidecar and corresponding blob hashes from the raw blob
-// data.
+// data with cell proofs for Fulu/Fusaka compatibility (Version1).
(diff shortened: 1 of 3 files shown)


integration/eth2network/start-pos-network.sh
```diff
@@ -128,8 +139,9 @@ echo "BEACON PID $beacon_pid"
 # Run Prysm validator client
 ${VALIDATOR_BINARY} --beacon-rpc-provider=127.0.0.1:"${BEACON_RPC_PORT}" \
             --datadir="${VALIDATORDATA_DIR}" \
+            --interop-num-validators 4 \
```
tudor-malene: what's up with these 4 validators again?
Something needed by the new version?

PR #2713, opened 2025-10-15
[Checklist] Whitelist usdt and usdc. Add weth

StefanIliev545:
### Why this change is needed

For mainnet we need to support weth that has a special bridge flow. Furthermore we need to whitelist usdt and usdc by default in order to allow import of stables into the ten network.

Wired in env variables to expose usdt and usdc addresses to whitelist. 
Added a flow for weth. 

### PR checks pre-merging


contracts/deployment_scripts/testnet/layer1/002_whitelist_tokens_and_setup_weth.ts
```diff
@@ -0,0 +1,69 @@
...
+        console.error("USDT_ADDRESS environment variable is not set.");
+        process.exit(1);
+    }
+
+    if (!wethAddress) {
+        console.error("WETH_ADDRESS environment variable is not set.");
+        process.exit(1);
+    }
+
+    console.log(`Whitelisting USDC: ${usdcAddress}`);
```
BedrockSquirrel: minor thing but shouldn't we put these log messages before each relevant execute() so the logs aren't confusing on failure?

tudor-malene (commented):
Very hard to get my head around this.

Pls add a comment somewhere describing the scenarios. ( Which @moraygrieve can then use for testing).

1. Alice wants to send native ETH from TEN to Ethereum.  What happens?
2. Etc

StefanIliev545:
> Very hard to get my head around this.
> 
> Pls add a comment somewhere describing the scenarios. ( Which @moraygrieve can then use for testing).
> 
> 1. Alice wants to send native ETH from TEN to Ethereum.  What happens?
> 2. Etc

Alice wants to send WETH from TEN to Ethereum, not native. 

1. sendERC20 is called, with the asset being transfered being WETH. 
2. Internally the contract unwraps it and sends it over as native with recipient being the remote bridge. (Basically bridge transfers to self, but on other chain)
3. A message is encoded for a cross chain call of receiveNativeWrapped with the end recipient
4. On the other chain, we call the CrossChainMessenger with the payload (and proof) which internally invokes the receiveNativeWrapped that will pull out from the native amount at the bridge, put it in the WETH contract and then transfer it to Alice on Ethereum. 

The EthereumBridge's sendNative is used to burn. Otherwise its useless and not meant to be forwarded as on ethereum there is no native balance being minted.

On TenBridge the sendNative(to_bridge) is used in conjunction with the receiveNativeWrapped(to_alice) to mint native value on TEN with the bridge as owner, so that the call to also send it out as wrapped has something to wrap.

PR #2714, opened 2025-10-16
use default user if existing user is not set for ExecAuth

zkokelj:
### Why this change is needed

In order to enable public user access for unauthenticated users we should have a default user and use it to do the calls.

- Generate default user on gateway startup and keep it in memory
- Use that user if no token was provided in the gateway URL
- Skip rate limiting for the default user (since all unauthenticated users will use the default user we should skip it to give them good user experience and not rate limit them all the time)
- 

### PR checks pre-merging

### DIFF
--- a/tools/walletextension/services/sk_manager.go
+++ b/tools/walletextension/services/sk_manager.go
@@ -2,7 +2,6 @@ package services
 
 import (
 	"context"
-	"encoding/hex"
 	"errors"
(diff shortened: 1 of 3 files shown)


tudor-malene (changes requested):
looks good in principle. One small refactor

tudor-malene (approved):
lgtm

PR #2715, opened 2025-10-16
Host: eagerly load L1 blocks from l1StartHash

BedrockSquirrel:
### Why this change is needed

Currently the host has to wait until it has loaded all L1 blocks from l1StartHash to current head into the DB before it will start feeding blocks to the enclave, but we can do this simultaneously with feeding the enclave if we eagerly load from the start based on height.

- at startup fetch height of l1StartHash and begin eagerly loading blocks from there
- keep track (persisted) of how far that has got to avoid gaps if we have downtime
- change the fetchAncestor logic (which helps us self heal if there are issues, e.g. a fork) to be iterative rather than recursive and not to be used if there is a large gap (let the eager loading handle it).

Also changed, only slightly related:
- changed where the data service is initialise so it fits the pattern of the other services and can receive the hostInterrupt control that gets passed everywhere for shutdown

### PR checks pre-merging

### DIFF
--- a/go/host/l1/dataservice.go
+++ b/go/host/l1/dataservice.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/ten-protocol/go-ten/contracts/generated/DataAvailabilityRegistry"
+	"github.com/ten-protocol/go-ten/go/common/errutil"
 	"github.com/ten-protocol/go-ten/go/common/stopcontrol"
 
 	"github.com/ten-protocol/go-ten/go/ethadapter/contractlib"
@@ -37,7 +38,7 @@ import (
 const (
 	// minimum gap in block height to trigger eager fetching of historical blocks - 5 blocks should not happen in normal operation
 	_minGapForEagerFetch = 5
-	eagerFetchHeightKey  = "eager_fetch_height"
+	localHeightKey       = "local_fetch_height"
 )
 
 var (
@@ -58,7 +59,7 @@ type DataService struct {
 	blockResolver    storage.BlockResolver
 	l1StartHash      gethcommon.Hash
 
-	eagerFetchHeight atomic.Uint64 // the next block height to eagerly fetch historical blocks from
+	localBlockHeight atomic.Uint64 // the latest block height to eagerly fetch historical blocks from
 	head             gethcommon.Hash
 
 	running       atomic.Bool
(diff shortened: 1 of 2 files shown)


go/host/storage/hostdb/hostdb.go
```diff
@@ -70,3 +70,25 @@ func (b *dbTransaction) Rollback() error {
 	}
 	return nil
 }
+
+func GetMetadata(db HostDB, key string) (uint64, error) {
+	var value uint64
+	query := "SELECT CAST(val AS INTEGER) FROM config WHERE ky = ?"
+	err := db.GetSQLDB().Get(&value, query, key)
+	if err != nil {
+		if err == sql.ErrNoRows {
```
tudor-malene: let's use the standard errnnotfound

go/host/l1/dataservice.go
```diff
@@ -84,6 +100,51 @@ func (r *DataService) Stop() error {
...
 }
 
+func (r *DataService) initEagerFetchHeight() {
+	if r.blockResolver == nil {
+		r.logger.Error("blockResolver not set, cannot load eager fetch height")
+		return
+	}
+
+	height, err := r.blockResolver.GetMetadata(eagerFetchHeightKey)
+	if err != nil {
```
tudor-malene: this assumes the err is ErrNotFound

go/host/l1/dataservice.go
```diff
@@ -496,49 +566,77 @@ func (r *DataService) streamLiveBlocks() {
...
+				r.logger.Info("Eager fetch caught up to latest L1 block", "height", height)
+				return
+			}
+			r.logger.Warn("Eager fetch block by number failed", "height", height, log.ErrKey, err)
+			// abandon for now, retry will be triggered by L1 block arrival
+			return
+		}
+
+		r.storeBlockHeader(block)
+		height += 1
```
tudor-malene: shoudlnt' height be `block.height`? Just to avoid any weird corner cases

tudor-malene (commented):
Few comments, nothing major

go/host/l1/dataservice.go
```diff
@@ -496,49 +566,77 @@ func (r *DataService) streamLiveBlocks() {
...
+		r.updateEagerFetchHeight(height)
+	}
+}
+
 func (r *DataService) storeBlockHeader(blockHeader *types.Header) {
```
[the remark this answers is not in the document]
BedrockSquirrel: yeah, that's mad - I wonder how many issues that's been causing just carrying on when there's a problem... Have changed it to return an error and handle it in each usage.

PR #2726, opened 2025-10-21
enclave: clarity on errors causing 'conflict'

BedrockSquirrel:
### Why this change is needed

We have seen errors like this on the validators when there are ongoing issues in sepolia. It seems to be a weird race condition or something because I can see it's just reprocessing the batch with the same hash so it should just reject it as a duplicate but the lookup by hash  must be failing and we can't see why.

`Batch conflict detected when adding batch to L2 repo - enclave corrupted node_id=0xF7E8fb5b78dA9e37F1Ed210196e5EddbB7f868Fd cmp=host enclave_id=0x4467bE1560be413ce6Fd0d7F6902Dc265cBb88C9 batch=0xebdde2e8fe253123b332eeae33e4cb4053d8f37c6c5fd85434e2b325453f968a err=conflict`

### What changed

Verify the error when lookups fails is a 'not found' error instead of assuming it is. If it's an unexpected error then fail with that.

### PR checks pre-merging

### DIFF
--- a/go/host/l2/batchrepository.go
+++ b/go/host/l2/batchrepository.go
@@ -218,23 +218,26 @@ func (r *Repository) AddBatch(batch *common.ExtBatch) error {
 	defer r.batchLock.Unlock()
 
 	_, err := r.storage.FetchBatch(batch.Hash())
-	// we found the exact same batch
+	// no error means we found the exact same batch (same hash)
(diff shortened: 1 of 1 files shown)


tudor-malene (approved):
lgtm - but it's not on the enclave

PR #2729, opened 2025-10-22
Upgrade beacon API with paid tier

badgersrus:
### Why this change is needed

Fusaka fork has changed the way we verify blobs. The API doesn't return the legacy KZG proofs that we previously used to verify the blob proofs so we can only verify the hashes and commitments. If the API is enhanced to include cell proofs then we can verify them. 

* Switch from VerifyBlobProofs to verifying the hashes and commitments
* Added better logging so we know if its been thrown from the archive client
* Unit test to check the new logic
* Fixed the setMetaData query for postgres

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #2730, opened 2025-10-23
Include base ref in the repository dispatch

moraygrieve:
### Why this change is needed

When a PR is merged the repository dispatch is used to trigger running the e2e tests. At the moment the trigger always runs the tests against main ... it would be good to have the option to extract the target branch, and if a release branch, run the tests against that release branch.


zkokelj (approved):
lgtm 👍

PR #2749, opened 2025-11-25
Enclave: avoid corrupting data when block processing fails

BedrockSquirrel:
### Why this change is needed

We hit an issue a while back where a context canceled (host was restarted) caused the rollup processing to fail halfway through and then the node was in a bad state because it had missed the batches that were in a rollup but proceeded with the next L1 block.

- use a separate context to the RPC req context for data writes
- attempt to rollback processed block immediately if we don't complete processing
- restart enclave if that fails to avoid corrupt states

### PR checks pre-merging

### DIFF
--- a/go/common/errutil/errors_util.go
+++ b/go/common/errutil/errors_util.go
@@ -20,13 +20,14 @@ var (
 
 	// Standard errors that can be returned from block submission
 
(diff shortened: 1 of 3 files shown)


tudor-malene (approved):
lgtm

PR #2756, opened 2025-12-03
never panic when printing audit message

zkokelj:
### Why this change is needed

When fmt.Sprintf() formats a struct containing a nil *hexutil.Big, it calls the String() method, which in turn calls EncodeBig() on a nil pointer, causing a segfault.

I wrapped fmt.Sprintf() in an anon function with recover() to catch panics during formatting. If formatting fails, a safe fallback message is logged instead of crashing the application.

### PR checks pre-merging


BedrockSquirrel (approved):
oof, LGTM!

PR #2767, opened 2025-12-09
gw expire funds on session key delete

zkokelj:
### Why this change is needed

We already implemented expiring funds on the session key (sending them back to the users account). 
But this was not done when user deletes the session key and this can result in unexpected loss of funds for the user.

With this change we try to send funds back to user owned account before deleting the session key.
If funds cannot be sent back we log the error and don't delete the account as this could lead to loss of funds and in my opinion it is more important than deleting the session key (users can have up to 100 session keys at the same time).

### PR checks pre-merging

### DIFF
--- a/tools/walletextension/services/sk_manager.go
+++ b/tools/walletextension/services/sk_manager.go
@@ -101,7 +101,6 @@ func (m *skManager) DeleteSessionKey(user *common.GWUser, sessionKeyAddr gethcom
 			m.logger.Info("Successfully transferred funds from session key before deletion",
 				"userID", common.HashForLogging(user.ID),
 				"sessionKeyAddress", sessionKeyAddr.Hex())
-
 		}
 	}
 


tools/walletextension/services/sk_manager.go
```diff
@@ -79,6 +81,30 @@ func (m *skManager) DeleteSessionKey(user *common.GWUser, sessionKeyAddr gethcom
...
+				"sessionKeyAddress", sessionKeyAddr.Hex())
+		} else {
+			_, err := m.txSender.SendAllMinusGasWithSK(context.Background(), user, sessionKeyAddr, *firstAccount.Address)
+			if err != nil {
+				m.logger.Error("Failed to recover funds from session key before deletion",
+					"error", err,
+					"userID", common.HashForLogging(user.ID),
+					"sessionKeyAddress", sessionKeyAddr.Hex())
+				return false, err // fail deletion if funds transfer fails - we won't be able to recover funds after deletion
+			} else {
```
tudor-malene: this else branch is not needed

PR #2791, opened 2025-12-17
Tenscan contract listing

badgersrus:
### Why this change is needed

https://github.com/ten-protocol/ten-internal/issues/6043

This is just the first part of dev, second part will add balances and event logs

* Add a new goroutine enclave service that periodically fetches the contract data from the enclave. We store the last fetched batch as metadata in the host and each fetch is done from the last fetched batch 
* New table in the host to store the data so we dont jam up the edglessDB
* Contract listing test is local only because indexing on CI is slow and we don't want to add ~1min to every PR build

### PR checks pre-merging

### DIFF
--- a/go/host/storage/hostdb/contracts.go
+++ b/go/host/storage/hostdb/contracts.go
@@ -14,7 +14,7 @@ import (
 const (
 	insertContract      = "INSERT INTO contract_host (address, creator, transparent, custom_config, batch_seq, height, time) VALUES (?, ?, ?, ?, ?, ?, ?)"
 	selectContract      = "SELECT id, address, creator, transparent, custom_config, batch_seq, height, time FROM contract_host"
-	selectContractCount = "SELECT COUNT(*) FROM contract_host"
+	selectContractCount = "SELECT total FROM contract_count WHERE id = 1"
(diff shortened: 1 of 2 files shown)


badgersrus:
@zkokelj made a few important changes:
* Added the limit and sync interval as config so we can easily tweak in a live environment
* Use the primary key of the contract table to check where we last fetched from as the old version would miss batches if we have an instance where we had more than fetch limit number of contracts in a batch 
* Added a local only sim test to check the new behaviour 
* All the config params are added in [this](https://github.com/ten-protocol/ten-apps/pull/72/files) PR

go/host/storage/hostdb/contracts.go
```diff
@@ -0,0 +1,205 @@
...
+			Time:            contract.Time,
+		})
+	}
+
+	if err = rows.Err(); err != nil {
+		return nil, err
+	}
+
+	var total uint64
+	reboundCountQuery := db.GetSQLDB().Rebind(selectContractCount)
```
zkokelj: any reason why are we not using "SELECT total FROM contract_count WHERE id = 1" like we do in GetTotalContractCount?

PR #2793, opened 2025-12-19
Configurable max L1 blocks and max blob retries

badgersrus:
### Why this change is needed

We saw some skipped sepolia blocks which meant the window of `2*L1blockTime` was too short so its been made configurable and will be set to 5 for UAT/ Sepolia. We also have seen blob publishing get stuck when blob gas price spikes because we hit the max retries so added config for this so we can temporarily unblock them by increasing the retry price and avoid `replacement transaction underpriced` 

* Add new `L1TimeoutBlocks` param
* Add new `MaxBlobRetries` param

### PR checks pre-merging

### DIFF
--- a/go/config/defaults/0-base-config.yaml
+++ b/go/config/defaults/0-base-config.yaml
@@ -27,6 +27,9 @@ network:
     blockTime: 15s
     rollupRetryDelay: 5s
     startHash: 0x0 # hash of the L1 block that Ten can safely start processing from to get entire L2 history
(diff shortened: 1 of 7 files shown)


BedrockSquirrel (approved):
Nice, LGTM.

Minor quibble that comments like `// number of blocks we wait before disconnecting if no blocks seen` are slightly confusing. Could mention blocktimes or expected blocks or something but I think it's clear enough.

PR #2813, opened 2026-01-22
Attestation: use intel PCS for verification collateral

BedrockSquirrel:
### Why this change is needed

Piecing together the issue with OVH-Azure attestation compatibility it seems (from info [like this](https://learn.microsoft.com/en-us/azure/attestation/faq?utm_source=chatgpt.com#is-software-guard-extensions--sgx--attestation-supported-by-azure-attestation-in-non-azure-environments-)) that Azure servers need to use the Azure THIM service as their PCCS_URL because non-azure endpoints don't have all the data to produce the attestation but that Azure service is unable to verify non-azure attestations.

The COLLATERAL_SERVICE config allows you to supply a separate collateral service endpoint that gets used for verification for this case. We need to test it a bit because I've seen mention in a couple of places that it's less reliable (I think that might be because it's **stricter** than the azure endpoint).

Allow COLLATERAL_SERVICE to be configured for enclaves, default it to the Intel PCS endpoint.

### PR checks pre-merging


tudor-malene (approved):
lgtm

PR #2833, opened 2026-02-26
Bridge design doc update

badgersrus:
### Why this change is needed

Old one was stale as hell

* Complete refresh

### PR checks pre-merging


design/bridge/bridge_design.md
```diff
@@ -1,364 +1,595 @@
...
-      "from": "0xLockContract", 
-      "chainId": 1 
-    }
-    ```
-  * As an L2 developer I want to be able to query specific message topics coming from specific addresses.
-  * The TEN gas calculation should include the L1 fees of submitting a message that originates on the L2. Users should fully cover the storage fees and the Rollup publisher should not incur any financial loss.
-  * The messaging and bridge are decentralized as much as possible.
-  * Block reorganizations should have no impact on the security of the messaging protocol and the downstream apps that use this protocol.
+- Provide a generic, permissionless messaging API between L1 and L2
+- Build the bridge as an ordinary application on top of that API (no special privileges)
```
tudor-malene: Build a reference bridge...

design/bridge/bridge_design.md
```diff
@@ -1,364 +1,595 @@
...
+    uint64  nonce;            // app-defined deduplication key
+    uint32  topic;            // message category (TRANSFER, MANAGEMENT, VALUE)
+    bytes   payload;          // arbitrary encoded data
+    uint8   consistencyLevel; // block confirmations before finality
+}
+```
 
-* `System contract` - A contract that is deployed by the TEN protocol when the Layer 2 network is created. System contracts serve the purpose of providing system API access on-chain. 
+- `sender` + `sequence` uniquely identify a message globally
+- `nonce` lets applications group related messages or deduplicate identical payloads
```
tudor-malene: would be good to add a bit more detail on  sequence vs nonce. A bit unclear to me what the nonce does

design/bridge/bridge_design.md
```diff
@@ -1,364 +1,595 @@
...
+- `topic` provides basic routing and versioning — the bridge uses topics `0` (transfer), `1` (management), and `2` (value)
+- `consistencyLevel` controls how many L1 block confirmations must pass before the message is considered final
```
tudor-malene: only applies to L2->L1 messages

design/bridge/bridge_design.md
```diff
@@ -1,364 +1,595 @@
...
+    address sender;
+    address receiver;
+    uint256 amount;
+    uint64  sequence;
 }
 ```
 
-The messaging API will be provided by the new `MessageBus` system contract. This contract's interface will be available for L1 and L2. This contract needs to be [`Ownable`](https://docs.openzeppelin.com/contracts/2.x/access-control) or [`RBAC`](https://docs.openzeppelin.com/contracts/2.x/access-control#role-based-access-control) based as some of the functions should only be callable from an administrative trusted address.
-This trusted address can either be the `Enclave` or the `ManagementContract`. As a system contract, the `MessageBus` should be created during the network bootstrap process.
+This exists because native value transfers don't go through the messenger relay — they're verified directly via Merkle proofs on L1.
```
tudor-malene: why?

design/bridge/bridge_design.md
```diff
@@ -1,364 +1,595 @@
...
+
+---
 
+## 4. MessageBus (Transport)
 
+The MessageBus is the foundation. It does two things:
```
tudor-malene: this needs a bit more context. What exactly is the `MessageBus` on the L1 and on the L2?

design/bridge/bridge_design.md
```diff
@@ -1,364 +1,595 @@
...
-        uint32       nonce,
-        bytes memory topic,
-        bytes memory payload
-) public returns (bool)
+```solidity
+function verifyMessageFinalized(CrossChainMessage message) external view returns (bool);
 ```
 
-Internally, the function will hash the message and compare it with the result of the key in `receivedMessages` map. If the map contains a `true` under this key, then the message has been received by this contract and `verifyMessageReceived` will return true. This is useful as a cheap-ish way to verify something has happened by utilizing the end user's client (or browser) as the transfer mechanism for the full message.  
+Returns `true` if the message has been stored **and** its challenge period has elapsed.
```
tudor-malene, line 130: this feels a bit confusing. 
Storing happens only on the L2, and the challenge period is on the L1.

I'd say this needs a bit more context

design/bridge/bridge_design.md
```diff
@@ -1,364 +1,595 @@
...
+State roots have an activation time — they can't be used for verification until `block.timestamp >= activationTime`. This is the challenge period. State roots can also be disabled if a rollup is successfully challenged.
+
+---
+
+## 5. Cross Chain Messenger 
 
-When called, the function should store the message indexed in storage - map of `senders` that contains a map of `topics` which points to array of `messages` should be sufficient. 
-Along with it, the hash of the whole message should be used as a key to store a message received flag inside the map `receivedMessages`. This map enables quick verification that a message is valid and received.
+The messenger sits between the MessageBus and application contracts. It takes a verified message and executes it — calling the target contract with the encoded calldata.
```
tudor-malene, line 158: needs a bit more context. 
Maybe something like "xchain messages are generally remote function calls on the other chain. The transport layer sends the encoded message, but we also need a layer to execute. "

design/bridge/bridge_design.md
```diff
@@ -1,364 +1,595 @@
...
-With the API we will build an initial bridge that wraps assets and might progress this further.
+TEN is an L2 that needs to communicate with Ethereum L1. This document describes the cross-chain messaging protocol and reference bridge built on top of it..
 
-The API design proposal is inspired by the [Wormhole protocol](https://wormhole.com/). The idea is we would allow for authenticated messaging between Layer 1 and Layer 2 on TEN through a smart contract API available on both layers.
-It will be provided by two smart contracts deployed on network creation on both layers. 
-These contracts will be owned by the ManagementContract and the enclave on their respective layers.
+The messaging protocol is a general-purpose API. Any developer can use it to build cross-chain applications — bridges, oracles, governance relays, or anything else that requires authenticated data to move between L1 and L2.
 
-![Intro Diagram not found](./resources/IntroductionDiagram.svg)
+The reference bridge is one such application. It locks ERC20 tokens (and native ETH) on L1 and mints wrapped equivalents on L2, and vice versa.
```
tudor-malene, line 9: I would also add in the intro that the L2 chain is in perfect sync with the L1 because it consumes and links the L1 blocks directly into the L2 data structures.
Which means that L1->L2 messaging is real-time.

The advantage of using the native support is exactly this l1->l2 real time communication, and the security of l2->l1, because the xchain root is included in the signed and challenge-able payloads

design/bridge/bridge_design.md
```diff
@@ -1,364 +1,595 @@
...
-* Messaging architecture between TEN and Mainnet Ethereum.
-* Basic Bridge that wraps assets
-* Primitive fees implementation
+Each layer exists on both the L1 and L2, but with different implementations suited to each environment:
 
-## Assumptions
+| Component | L1 | L2 |
+|---|---|---|
+| Transport | `MerkleTreeMessageBus` verifies messages via Merkle proofs against rollup state roots | `MessageBus` stores messages submitted by the enclave from L1 block processing |
+| Relay | `CrossChainMessenger` relays messages using Merkle proofs | `CrossChainMessenger` relays messages after finality verification |
```
tudor-malene, line 33: a bit confused here. There is no finality verification on the L2

design/bridge/bridge_design.md
```diff
@@ -1,364 +1,595 @@
...
+```
+leaf = keccak256(abi.encode("v", keccak256(abi.encode(message))))
 ```
 
-This function should not be callable by any users or contracts unless they have the `ADMINISTRATOR` role or are the `Owner` of the `MessageBus` contract. Along with those requirements, on the TEN layer 2, there is an additional security measure - Any incoming transaction to this function is blocked, even if it is coming from an address that has the correct role. The protocol will block transactions that do not originate from inside of the enclave due to consuming L1 blocks. This means that even if the enclave gets hacked and the keys leak the verifiers will block withdraws.
+State roots have an activation time — they can't be used for verification until `block.timestamp >= activationTime`. This is the challenge period. State roots can also be disabled if a rollup is successfully challenged.
+
+---
+
+## 5. Cross Chain Messenger
```
tudor-malene, line 156: (Relayer)

design/bridge/bridge_design.md
```diff
@@ -1,364 +1,595 @@
...
+### Replay Protection
 
-![Diagram not found](./resources/PublishFromTen.svg)
+A `mapping(bytes32 => bool) messageConsumed` ensures each message can only be relayed once. The key is the hash of the full `CrossChainMessage`.
 
-> **_NOTE:_** **The messages must not be accessible unless** the challenge period has passed! On top of that the block where the message is submitted to L1 must have confirmations equal to `consistencyLevel` before the message is released. Those are simply counted on-chain as "confirmations" is meaningless for TEN L2.  
+---
 
-### Alternative approaches
+## 6. Developer API (CrossChainEnabledTEN)
```
tudor-malene, line 206: You can say these are convenience libraries.
I think they need to be instantiated with the address of the relayer?

tudor-malene, line 206: Do we have a section in the docs where we explain how to build a cross chain app?

design/bridge/bridge_design.md
```diff
@@ -1,364 +1,595 @@
-# TEN Cross Chain Messaging And Initial Bridge - Design Document
+# TEN Cross-Chain Messaging & Bridge
 
-We want to expose a system API that allows any developer to come and build a bridge on Ten. We also want this API to support all types of well known bridges - wrapping assets, liquidity bridges, etc. And ideally we want the system API to further allow even more exotic type of apps - providing randomness to the L1, hidden game state and anything other that can be enabled as a cross chain application!
+## 1. Introduction
 
-With the API we will build an initial bridge that wraps assets and might progress this further.
+TEN is an L2 that needs to communicate with Ethereum L1. This document describes the cross-chain messaging protocol and reference bridge built on top of it..
```
tudor-malene: This document describes the `native` cross-chain messaging protocol and reference bridge built on top of it.

design/bridge/bridge_design.md
```diff
@@ -1,364 +1,595 @@
...
-    *  `bridgeAssets` makes the bridge transfer assets to itself.
-2. On layer 2
-    *  `receiveAssets` makes the bridge mint tokens to the receiving address. The bridge can do that as it owns the ERC20 wrapper contracts.
-    *  `bridgeAssets` makes the bridge burn wrapped assets from the address of `msg.sender` instead of transferring them.
+| Threat | Mitigation |
+|---|---|
+| **Fake messages from L2** | L1 verifies messages against rollup state roots via Merkle proofs. Fake messages would require forging a state root, which requires compromising the rollup. |
+| **Premature message consumption** | Challenge period enforced via `activationTime` on state roots and `finalAfterTimestamp` on stored messages. |
+| **L1 block reorganisation** | Rollups are bound to specific L1 block hashes. If the L1 reorgs, the rollup referencing the old block becomes invalid, and any messages it contained are rejected. |
+| **Fake messages to L2** | Only synthetic transactions from the enclave can call `storeCrossChainMessage`. User transactions to the MessageBus are blocked at the protocol level. |
```
tudor-malene, line 467: what about the events emitted on the l1? How are those authenticatd?

design/bridge/bridge_design.md
```diff
@@ -1,364 +1,595 @@
...
+| Contract | Purpose |
+|---|---|
+| `NetworkConfig` | Registry of all deployed contract addresses |
+| `CrossChain` | Manages cross-chain value transfer withdrawals and bundle verification |
+| `DataAvailabilityRegistry` | Publishes rollup state roots to `MerkleTreeMessageBus` |
+| `NetworkEnclaveRegistry` | Manages enclave registration |
+| `Fees` | Configurable fee parameters for message publishing |
+
+
+# Addition stuff(not sure if we want to include)
```
tudor-malene: yes, this is pretty good

tudor-malene (commented):
Nice one. Few comments

design/bridge/bridge_design.md
badgersrus: good spot!

badgersrus: No thats a good point - ill add it

PR #2836, opened 2026-03-03
 use LRU eviction to prevent session key data loss in the gateway

zkokelj:
Why this change is needed
The SessionKeyActivityTracker has a hard limit of 100k entries. When this limit is reached, new session key activities are silently dropped (those keys will never expire and funds).

What changes were made as part of this PR
I replaced the simple map with an LRU cache. Instead of dropping new entries at capacity, we now evict the oldest entry and persist it to CosmosDB via an async batch writer.

Changes:

`session_key_activity.go` — LRU cache using container/list with O(1) eviction; background goroutine batches evicted entries (100 items or 5s flush interval) and writes to CosmosDB
`session_key_expiration.go `— Expiration check now queries both in-memory cache AND CosmosDB, merging results to catch entries that were evicted from memory
`session_key_activity_storage.go` — Added SaveBatch(), ListOlderThan(), and Delete() to storage interface
`cosmosdb/session_key_activity_storage.go` — Implemented incremental upsert (read-merge-write) for batched persistence
Graceful shutdown flushes pending writes before exit
Write frequency impact: No additional CosmosDB writes unless cache exceeds 100k entries. When evictions occur, writes are batched efficiently (~1-2 shard writes per 100 evictions).

### DIFF
--- a/tools/walletextension/services/session_key_activity.go
+++ b/tools/walletextension/services/session_key_activity.go
@@ -1,12 +1,14 @@
 package services
 
 import (
+	"container/list"
 	"sync"
 	"time"
 
 	gethcommon "github.com/ethereum/go-ethereum/common"
 	gethlog "github.com/ethereum/go-ethereum/log"
 	"github.com/ten-protocol/go-ten/tools/walletextension/common"
+	"github.com/ten-protocol/go-ten/tools/walletextension/storage"
 )
 
 // SessionKeyActivityTracker exposes a minimal API for tracking activity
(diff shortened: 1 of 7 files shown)


tools/walletextension/services/session_key_activity.go
```diff
@@ -16,62 +18,224 @@ type SessionKeyActivityTracker interface {
...
+			}
+		}
+	}
+}
+
+// Stop gracefully shuts down the tracker, flushing pending writes
+func (t *sessionKeyActivityTracker) Stop() {
+	t.stopOnce.Do(func() {
+		close(t.stopChan)
+		if t.persistQueue != nil {
```
badgersrus, line 148: I don't think we need to close this as well - the stopChan calls `close` in `stop_control.go`?

tools/walletextension/services/session_key_expiration.go
```diff
@@ -148,11 +161,44 @@ func (s *SessionKeyExpirationService) sessionKeyExpiration() {
...
 
-		// After successful external operation, delete from tracker
-		_ = s.activityTracker.Delete(c.Addr)
+		// After successful external operation, delete from both tracker and DB
+		s.deleteActivity(c.Addr)
 	}
 
 	// store all activities in the database to make them persistent and recoverable in case of restart
 	allActivities := s.activityTracker.ListAll()
 	_ = s.activityStorage.Save(allActivities)
```
badgersrus, line 170: I might be misunderstanding the code but won't this overwrite potentially expired entries that occur between eviction the time expiration threshold?

1. session key A gets evicted (as its hit limit) from memory and sent to cosmos via the persist queue
2. key A is in cosmos and not in the in-memory tracker 
3. expiration loop firsts sessionKeyExpirations() but key A isn't past the expiration
4. `ListAll()` returns only the memory (key A missing)
5. `Save` writes that list and overwrites and deletes key A?

